PR #20, opened 2025-02-24
[19] Testing GH actions

tjhunter:
Adds all the github action workflow to check  the code.


.github/workflows/ci.yml
```diff
@@ -15,11 +23,11 @@ jobs:
...
 
-      - name: Set up Python
-        run: uv python install
-
-      - name: Run black
-        run: uv run black --check src/wegen_utils     
+      - name: Run ruff (black)
+        # Do not attempt to install the default dependencies, this is much faster.
+        # Run temporarily on a sub directory before the main restyling.
+        run: uv run --no-project --with "ruff==0.9.7" ruff format --check -n src/wegen_utils
```
clessig: What is src/wegen_utils ?

pyproject.toml
```diff
@@ -34,3 +35,37 @@ evaluate = "weathergen:evaluate"
...
+  "pytest",
+  "ruff==0.9.7"
+]
+
+
+# The linting configuration
+[tool.ruff]
+
+# Wide rows
+line-length = 120
```
clessig: Can we change this to 100 or is this uncommon these days?

clessig (approved):

.github/workflows/ci.yml
clessig: I think it should be src/weathergen_utils to be consistent with src/weathergen

Issue #28, opened 2025-02-27
Refactor data handling between data loader and model


clessig:
### What maintenance does this project need?

The data that needs to be passed from the dataloader (MultiStreamDataSampler) to the model is currently not well structured. Having a reasonable structure there is also non-trivial because we make very little assumptions about the structure of the data at the moment, so that one has lists of lists of lists ... to handle the potentially ragged shape of the data across streams and steps. Nonetheless, to make the code more readable and maintainable, the data handling needs to be improved. 

The current proposal is to wrap all data per stream in one class. This will hide some of the complexity. A very limited number of data items computed in the dataloader cut across streams (e.g. some precomputed indices for the scatter operation in the local assimilation engine). These should be kept separate

### Organisation

ECMWF

clessig:
Closed with #34 .

Issue #16, opened 2025-02-20
Refactor data loaders


clessig:
Currently, two data loaders are supported. One is for anemoi datasets and one for obs datasets. While these have a common interface, this is rather haphazard. A base class for the data loaders should be defined that defines a template for the required interface. This should simplify the development of additional data loaders. The base class should also implement the infrastructure to ensure that datasets are aligned in time using a reference index set.

CC @tjhunter

PR #34, opened 2025-03-02
Refactor of batch data handling and passing to model

clessig:
The way the data is handled in the dataloader has been completely revised, addressing #28 . Per stream data is now collected in a StreamData class instance, which allows for a much cleaner program flow in the data loader and with the complexity hidden in the StreamData class. This PR won't be the end of the refactoring needed for the data loading and handling (e.g. it does not directly address #16) but it should be a major step forward.


tjhunter (approved):
I will still need to give a thorough look at this part of the code

PR #40, opened 2025-03-06
Enable parsing arguments to train and evaluate

mlangguth89:
Implement argparse to parse arguments of `train` and `evaluate` via command line. 
Other configurations are largely left untouched, except for the `start_date` and `end_date`-configuration in `evaluate`, which is now performed via a pandas datetime-object instead of an integer. Thus, any parsed date-argument that can be converted into a datetime-object can now be handled and will be correctly used for configuration.


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/WeatherGenerator?pullRequest=40) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/WeatherGenerator?pullRequest=40) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/WeatherGenerator?pullRequest=40) it.</sub>

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
-    cf.loader_num_workers = min(cf.loader_num_workers, samples)
+    cf.loader_num_workers = min(cf.loader_num_workers, args.samples)
 
     trainer = Trainer()
-    trainer.evaluate(cf, run_id, epoch, True)
+    trainer.evaluate(cf, args.run_id, args.epoch, True)
 
 
 ####################################################################################################
-def train(run_id=None) -> None:
```
clessig, line 69: There should be a separate function train_continue that loads config. Here you likely unintentially overwrite some things.

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
-    forecacast_steps=None,
-    samples=10000000,
-    shuffle=False,
-    save_samples=True,
-    gridded_output_streams=[],
-):
+def evaluate():
+    """
+    Main evaluation function for WeatherGenerator model.
+    Parses command line arguments and runs evaluation by calling Trainer.evaluate().
```
clessig: Remove

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
-    masking_mode=None,
-    forecacast_steps=None,
-    samples=10000000,
-    shuffle=False,
-    save_samples=True,
-    gridded_output_streams=[],
-):
+def evaluate():
+    """
+    Main evaluation function for WeatherGenerator model.
```
clessig: We only have one evaluation function so it should be: "Evaluation function ..."

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
+    Parses command line arguments and runs evaluation by calling Trainer.evaluate().
+    Note: The parsed arguments are used to overwrite the model configuration for evaluation.
```
clessig: That's not correct since, e.g., run_id is not overwritten.

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
+
+    Args:
+        run_id (str): Run/model id of pretrained WeatherGenerator model.
+        start_date (str): Start date for evaluation. Format must be parsable with pd.to_datetime.
+        end_date (str): End date for evaluation. Format must be parsable with pd.to_datetime.
+        epoch (int, optional): Epoch of pretrained WeatherGenerator model used for evaluation (-1 corresponds to last epoch). Defaults to -1.
+        masking_mode (str, optional): Masking mode for evaluation. Defaults to None.
```
clessig, line 34: Remove, not supported at the moment

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
+    """
+    Main evaluation function for WeatherGenerator model.
+    Parses command line arguments and runs evaluation by calling Trainer.evaluate().
+    Note: The parsed arguments are used to overwrite the model configuration for evaluation.
+
+    Args:
+        run_id (str): Run/model id of pretrained WeatherGenerator model.
+        start_date (str): Start date for evaluation. Format must be parsable with pd.to_datetime.
+        end_date (str): End date for evaluation. Format must be parsable with pd.to_datetime.
+        epoch (int, optional): Epoch of pretrained WeatherGenerator model used for evaluation (-1 corresponds to last epoch). Defaults to -1.
```
clessig, line 33: -1 is the last checkpoint generated during training, not the last epoch

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
+        masking_mode (str, optional): Masking mode for evaluation. Defaults to None.
+        forecast_steps (int, optional): Number of forecast steps for evaluation. Defaults to None.
+        samples (int, optional): Number of samples for evaluation. Defaults to 10000000.
+        shuffle (bool, optional): Shuffle samples for evaluation. Defaults to False.
+        save_samples (bool, optional): Save samples for evaluation. Defaults to True.
```
clessig, line 38: sample from evaluation not for

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
+        analysis_streams_output (list, optional): Analysis output streams during evaluation. Defaults to ['ERA5'].
```
clessig, line 39: Should be: "streams that are saved"

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
+        gridded_output_streams(list, optional): Currently unused and threrefore omitted here
```
clessig, line 40: Please remove

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
+    parser.add_argument('--analysis_streams_output', type=list, default=['ERA5'], help='Analysis output streams during evaluation.')   
+
+    args = parser.parse_args()
+
     # TODO: move somewhere else
     init_loggers()
+
     # load config if specified
-    cf = Config.load(run_id, epoch if epoch is not None else -1)
+    cf = Config.load(args.run_id, args.epoch if args.epoch is not None else -1)
```
clessig: cf = Config.load(args.run_id, args.epoch)

(since the default argument is -1)

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
     cf.run_history += [(cf.run_id, cf.istep)]
 
-    cf.samples_per_validation = samples
-    cf.log_validation = samples if save_samples else 0
+    cf.samples_per_validation = args.samples
+    cf.log_validation = args.samples if args.save_samples else 0
 
-    if masking_mode is not None:
-        cf.masking_mode = masking_mode
+    if args.masking_mode is not None:
```
clessig: Remove

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
+        samples (int, optional): Number of samples for evaluation. Defaults to 10000000.
+        shuffle (bool, optional): Shuffle samples for evaluation. Defaults to False.
+        save_samples (bool, optional): Save samples for evaluation. Defaults to True.
+        analysis_streams_output (list, optional): Analysis output streams during evaluation. Defaults to ['ERA5'].
+        gridded_output_streams(list, optional): Currently unused and threrefore omitted here
+    """    
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('--run_id', type=str, required=True, help='Run/model id of pretrained WeatherGenerator model.')
+    parser.add_argument('--start_date', '-start', type=str, required=True, help='Start date for evaluation. Format must be parsable with pd.to_datetime.')
```
clessig: I think we should set sensible defaults (e.g. WeatherBench2 test period)

clessig:
@tjhunter : could you help with the auto-formatting and linting. Thanks!

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
+
     # TODO: move somewhere else
     init_loggers()
     private_cf = private_conf()
+
     cf = Config()
 
     # directory where input streams are specified
     # cf.streams_directory = './streams_large/'
+    # cf.streams_directory = './streams_anemoi/'
```
tjhunter: more generally, if some streams are missing in the config folder, we should add them

src/weathergen/__init__.py
```diff
@@ -11,71 +11,118 @@
...
-    trainer.evaluate(cf, run_id, epoch, True)
+    trainer.evaluate(cf, args.run_id, args.epoch, True)
 
 
 ####################################################################################################
-def train(run_id=None) -> None:
+def train() -> None:
+    """
+    Main training function for WeatherGenerator model.
+    Parses command line arguments and runs training by calling Trainer.train().
```
tjhunter: remove

tjhunter: This is where I would say that the model configuration is loaded from the configuration files.

src/weathergen/__init__.py
mlangguth89: I've just implemented argument parsing without overwriting things. `run_id` was an optional argument here, so I added it as parsing argument. 
However, since continunig training with the `train`-function won't work, it would be implement this function. Shall this happen in this pull request or in a separate feature-issue?

src/weathergen/__init__.py
mlangguth89: Why should these lines be removed, especially in context of the next comment?

src/weathergen/__init__.py
mlangguth89: So, one would pass a YAML-config file here? 
Shouldn't this happen in (another) feature-issue?

src/weathergen/__init__.py
tjhunter: Christian may have a different reason, but my take on this particular line: it describes somewhat what the code is doing  and it is not super informative. I would personally say: "Entry point for calling the evaluation code from the command line" . This implies that it is going to behave like a program.

It is minor in my opinion though

src/weathergen/__init__.py
tjhunter: to me, the defaults should be the ones in the configuration file.

src/weathergen/__init__.py
tjhunter: I would do it separately to not block this change.  We need to think more about the whole configuration system anyway.

src/weathergen/__init__.py
tjhunter: Sorry, I made my comment more clear above.

tjhunter (approved):
@mlangguth89  I approved the change. I still have a small comment about the comments, but you should feel free to edit it and merge. Thanks for your change!

clessig:
Can you please run ruff to fix the formatting so that we can merge.

Issue #57, opened 2025-03-10
Reproducible instructions to install flash attention on all working clusters


tjhunter:
### Is your feature request related to a problem? Please describe.

We do not have a standard way to install flash attention. This is a source of bugs and an issue for onboarding new people on the project.

Task:
- write a SOP for compiling/installing flash attention on HPC2020 (ideally it should also work on the other clusters)

tjhunter:
These instructions should also be relevant to pin the version of pytorch:
https://docs.astral.sh/uv/guides/integration/pytorch/#installing-pytorch

PR #68, opened 2025-03-12
[29] Logging in json format

### LINKED ISSUE #29: Refactor TrainLogger

The current logging system for training metrics has two critical shortcomings:

- Lack of granular logging: It is currently impossible to log loss function values for individual channels within each data stream. This limits visibility into model performance at a per-channel level, hindering detailed analysis.

- Fragile and unreadable format: Metrics are stored in unstructured plain-text (txt) files, which are not human-readable and lack a consistent schema. Minor changes to the order of logged metrics or their contents break backward compatibility, making post-processing, visualization, or comparison across runs error-prone and unmaintainable.

Replace the current plain-text logging format with CSV files using multi-level column headers to organize metrics hierarchically.

This structure would:

- Separate global training statistics (e.g., epoch, total loss) from per-stream and per-channel metrics.

- Ensure human readability while maintaining machine-readability.

- Prevent compatibility breaks due to explicit column naming and hierarchical organization.

Example: 
```
global ,global                     ,global  ,global   ,global   ,global                 ,global              ,FESOM               ,FESOM                ,FESOM               ,FESOM               ,FESOM               ,FESOM              ,FESOM              ,FESOM                 ,FESOM                 ,FESOM                 ,FESOM               ,FESOM               ,FESOM              ,FESOM
step   ,time                       ,samples ,perf_gpu ,perf_mem ,learning_rate          ,loss_mean           ,mse                 ,a_ice                ,evap                ,fh                  ,fw                  ,prec               ,snow               ,ssh                   ,sss                   ,sst                   ,swr                 ,tx_sur              ,ty_sur             ,std
1      ,2025-02-27 14:38:31.542443 ,320     ,98.5     ,48.75    ,2.9802549459833396e-06 ,0.9765826463699341  ,0.9765826463699341  ,0.9043928980827332   ,0.9301251173019409  ,0.972670316696167   ,0.936567485332489   ,0.9416562914848328 ,1.0318517684936523 ,1.0079926252365112    ,1.0319256782531738    ,0.941299319267273     ,0.9138585329055786  ,1.0519397258758545  ,1.054713249206543  ,
```

This solution was implemented in my fork: https://github.com/kacpnowak/WeatherGenerator2/pull/3

An alternative solution is to use an embedded database (e.g., DuckDB or SQLite) to store metrics in structured tables. Benefits include:

- Support for complex queries across multiple training runs.
- Native schema enforcement, eliminating fragility caused by format changes.
- Efficient storage and retrieval of large-scale experiments.
    
However, CSV files provide a simpler, more accessible intermediate solution that meets immediate needs without introducing database dependencies. Either approach would be preferable to the current system, which is unmaintainable and error-prone due to its reliance on unstructured text.

### Additional context

The current logging implementation’s lack of structure and flexibility actively impedes debugging, analysis, and iterative improvements. A structured format (CSV or database) is critical for scaling experimentation and ensuring reproducibility.

### Organisation

AWI

tjhunter:
Closes #29 

First part: prototyping the new format.


src/weathergen/utils/train_logger.py
```diff
@@ -8,8 +8,12 @@
 # nor does it submit to any jurisdiction.
 
 import datetime
+import json
+import os.path
+import time
 
 import numpy as np
+from mlflow.tracking.metric_value_conversion_utils import convert_metric_value_to_float_if_possible
```
tjhunter: this currently introduces a dependency on mlflow. I think this is fine but we can discuss. This is not trivial conversion code.

src/weathergen/utils/train_logger.py
```diff
@@ -21,31 +25,61 @@ def __init__(self, cf, path_run) -> None:
...
+        clean_metrics = {
+            "weathergen.timestamp": time.time_ns() // 1_000_000,
+            "weathergen.time": int(datetime.datetime.now().strftime("%Y%m%d%H%M%S")),
+        }
+        for key, value in metrics.items():
+            clean_metrics[key] = convert_metric_value_to_float_if_possible(value)
+
+        # TODO: performance: we repeatedly open the file for each call. Better for multiprocessing
+        # but we can probably do better and rely for example on the logging module.
+        with open(os.path.join(self.path_run, "metrics.json"), "ab") as f:
```
tjhunter, line 45: I suggest that we start with this simple version, we can always improve performance if it turns out to be a bottleneck

src/weathergen/utils/train_logger.py
kacpnowak: This function seems to quite an overkill. As far as I tested it can be replaced with this:
```python
def to_float(obj):
    try:
        return float(obj)
    except Exception:
        return obj
```
It seems to be that conversion logic is handled inside float()

pyproject.toml
```diff
@@ -91,4 +94,7 @@ ignore = [
   "F811",
   # To ignore, not relevant for us
   "E741",
-]
\ No newline at end of file
+]
+
+[tool.uv.sources]
+flash-attn = { url = "https://github.com/Dao-AILab/flash-attention/releases/download/v2.7.4.post1/flash_attn-2.7.4.post1+cu12torch2.6cxx11abiFALSE-cp312-cp312-linux_x86_64.whl" }
```
tjhunter: I had to make this change to use uv on the hpc2020 cluster. I am not sure if this is going to be a breaking change for people. @clessig , do we assume that different HPCs can use different versions of CUDA? That sounds like a nightmare.

clessig: We do not assume it, we know it ;) One can write a script that detects the available CUDA (and the python version if this is a variable) and then assembles the string that defines the wheel to be downloaded. @tjhunter : To what extent could one integrate this into pyproject toml?

clessig: And could we open an issues to track this? :)

tjhunter: I have the script in branch of the private repo but not committed yet: 
https://github.com/ecmwf/WeatherGenerator/issues/57

clessig (approved):

PR #75, opened 2025-03-14
[57] Pinning flash-attention and other dependencies

### LINKED ISSUE #57: Reproducible instructions to install flash attention on all working clusters

tjhunter:
Closes #57 

This PR ensures that the build environment is (mostly) reproducible:
- save version of flash-attn 
- pinned version of torch, cuda, ...

There should be no need for extra setup of flash-attn. I removed all caching and virtual environments, and uv downloaded the right dependencies.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -10,12 +10,6 @@
 %>cd WeatherGenerator
 %>uv sync
  
-# install flash_attn which is not uv compatible
-# (this step might take very long, it is faster when ninja is available but then
-# MAX_JOBS=4 should be set as environment variable; the step might need to be run
-# on a compute node to ensure the correct CUDA environment is available) 
-%>uv pip install torch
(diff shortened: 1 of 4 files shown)


clessig:
I think we should then also update here: https://github.com/ecmwf/WeatherGenerator/blob/b0c72a90f054c0de251930cab0ae073fa23ef435/INSTALL.md?plain=1#L18

clessig (approved):

tjhunter:
@clessig thank you for pointing out the install.md! Merging now.

PR #86, opened 2025-03-17
[57] Follow-up: fixing flash attention for local non-gpu development

tjhunter:
With this change, flash attention will not be loaded on macOS platforms. This is convenient for local development.

Fix for #57

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -104,6 +104,10 @@ ignore = [
 # exclude-newer = "2025-03-14T00:00:00Z"
 
 # Following the recommendations from https://docs.astral.sh/uv/guides/integration/pytorch
+# The current setup is:
+# linux == GPU + flashattention
+# windows == GPU
+# macos == CPU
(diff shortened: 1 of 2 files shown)


grassesi (approved):
Looks good to me, but just for clarifying: This does two things ?
1. exclude installation of flash-attention from non-linux systems
2. install pytorch-cpu on mac systems

Will this also allow for development on non-CUDA linux/windows machines ?

tjhunter:
@grassesi good point: this assumes right now that linux == GPU + flashattention, non-linux == CPU 

I think we can keep it this way for the time being, since flashattention is compiled only for linux.

grassesi:
I think its rather
- linux == GPU + flashattention
- windows == GPU
- macos == CPU

but I agree thats not essential right now

tjhunter:
Thank you @grassesi , I will update the comments in pyproject.toml and merge this

PR #89, opened 2025-03-17
added command line input for private and overwrite config

kctezcan:
With this implementation we have two command line inputs for only the train():
overwite_config and private_config

overwrite_config takes a yaml file path and overwrites the parameters given in the function body or adds to them.

private_config takes a yaml file path and sets some path configurations. If no path provided in the command line, the code looks for this in the environement variable WEATHERGEN_PRIVATE_CONF. if this is also empty, the code throws an error.

### DIFF
--- a/src/weathergen/__init__.py
+++ b/src/weathergen/__init__.py
@@ -16,7 +16,7 @@ import traceback
 import pandas as pd
 
 from weathergen.train.trainer import Trainer
-from weathergen.utils.config import Config, private_conf
+from weathergen.utils.config import Config, load_private_conf, load_overwrite_conf
 from weathergen.utils.logger import init_loggers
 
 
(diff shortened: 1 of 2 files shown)


src/weathergen/__init__.py
```diff
@@ -150,12 +150,29 @@ def train() -> None:
...
+    
+    parser.add_argument(
+        "--private_config",
+        type=str,
+        default=None,
+        help="Path to private configuration file for paths.",
+    )
+    
+    parser.add_argument(
+        "--overwrite_config",
```
tjhunter: I would call it just `config` instead of `overwrite-config` since the current config in the code is default values

src/weathergen/utils/config.py
```diff
@@ -69,11 +69,29 @@ def load(run_id, epoch=None):
...
+    " Return the overwrite configuration."
+    "If path is None, return an empty dictionary."
+    if not pth:
+        return {}
+    else:
+        overwrite_path = Path(pth)
+        overwrite_conf = yaml.safe_load(overwrite_path.read_text())
+        return overwrite_conf
+    
+def load_private_conf(pth: str = None) -> dict:
```
tjhunter: no default in private code please.

src/weathergen/utils/config.py
```diff
@@ -69,11 +69,29 @@ def load(run_id, epoch=None):
...
+    " Return the private configuration."
+    "If none, take it from the environment variable WEATHERGEN_PRIVATE_CONF. "
+    
+    if not pth:
+        if "WEATHERGEN_PRIVATE_CONF" in os.environ:
+            private_home = Path(os.environ["WEATHERGEN_PRIVATE_CONF"])
+            private_conf = yaml.safe_load(private_home.read_text())
```
tjhunter, line 89: great, I was not aware of `read_text`

tjhunter (commented):
Thanks @kctezcan ! I just have one question, and you will need to solve the linting issues.

kctezcan:
thanks for the suggestions @tjhunter , I just implemented these and ran the ruff command, so the check is OK now.

tjhunter (approved):

Issue #95, opened 2025-03-18
Defaulting epoch to -1 as the latest, but the code is expecting None as default


kctezcan:
### What happened?

After the recent merge of "Enable parsing arguments to train and evaluate #40", the epoch parameter default to -1 and as description it says that when epoch=-1, it should take the latest model:
```python
    parser.add_argument(
        "--epoch",
        type=int,
        default=-1,
        help="Epoch of pretrained WeatherGenerator model used for evaluation (-1 corresponds to the last checkpoint).",
    )
``` 
See line 67 here: https://github.com/ecmwf/WeatherGenerator/pull/40/files#diff-12c45a5f9512e05679defa0ea4c0ca1069edf5619bba5f180e443f8fa258b8efR67

However rest of the code has not been changed so rest of the code still expects None as the default value for loading the latest checkpoint. 

So 
```python
trainer = Trainer()
    trainer.evaluate(cf, args.run_id, args.epoch, True)
``` 

calls inside
```python
self.model.load(run_id_trained, epoch)
``` 

which then loads the model as 
```python
def load(self, run_id, epoch=None):
        path_run = self.cf.model_path + "/" + run_id + "/"
        fname = path_run + f"{run_id}"
        fname += f"_epoch{epoch:05d}.chkpt" if epoch is not None else "_latest.chkpt"
``` 

As can be seen, the code expects epoch=None for reverting to latest checkpoint.

An easy fix is to simply set the default to None as it was before the merge.

### What are the steps to reproduce the bug?

as above

### Version

0.0

### Platform (OS and architecture)

linux

### Relevant log output

```shell

```

PR #97, opened 2025-03-18
reverted default epoch to None from -1

kctezcan:
as mentioned in Issue https://github.com/ecmwf/WeatherGenerator/issues/95


tjhunter (approved):
@kctezcan  thanks for checking that

kctezcan:
Hi @tjhunter , I don't see the option to merge this, do you need to do it? Thanks

Issue #96, opened 2025-03-18
Backward compatibility for the start/end dates in evaluate function


kctezcan:
### Is your feature request related to a problem? Please describe.

Before the merge https://github.com/ecmwf/WeatherGenerator/pull/40 I could simply run evaluation without providing starting and ending dates. After the merge this is not possible anymore as these are required parameters (see line 50 in https://github.com/ecmwf/WeatherGenerator/pull/40/files#diff-12c45a5f9512e05679defa0ea4c0ca1069edf5619bba5f180e443f8fa258b8efR50). This is creating backward compatibility issues for me. 

### Describe the solution you'd like

I want to set these parameters as not required as before,

PR #98, opened 2025-03-18
made start/end dates not required in the evaluate function

kctezcan:
as described in Issue https://github.com/ecmwf/WeatherGenerator/issues/96

### DIFF
--- a/src/weathergen/__init__.py
+++ b/src/weathergen/__init__.py
@@ -51,14 +51,14 @@ def evaluate():
         "--start_date",
         "-start",
         type=str,
-        required=True,
+        required=False,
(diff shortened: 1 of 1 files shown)


tjhunter (approved):

mlangguth89 (commented):
@tjhunter and @kctezcan 
Doesn't this change break the code if `start_date` and `end_date` are not parsed?
The argument is converted to a pandas `datetime`-argument later and if it's not provided (is it `None` then?), the following line of code will raise an error:
https://github.com/MeteoSwiss/WeatherGenerator/blob/379becceb447561792e7b75a741bd7da44660d7f/src/weathergen/__init__.py#L105
This was the reason why I added `required=True`.
Alternatively, one could set 'sensible' defaults here.

Issue #103, opened 2025-03-20
Profiling: initial task


tjhunter:
We need the ability to profile our code:
- CPU profiling for I/O and CPU-side data manipulation
- GPU profiling for GPU I/O and kernels

Deliverables:
- operating procedures on how to use the current GPU profiler (nvidia nsight) with the code
- code changes if necessary

The focus is for now on single machines, but the tools and setup should be adaptable to clusters.

PR #104, opened 2025-03-20
[103] Initial profiling tools

tjhunter:
Solves #103 
Initial profiling support (which does not require many changes as it turns out).

SOPs and command lines are in a companion PR for the private repo.


src/weathergen/__init__.py
```diff
@@ -14,11 +14,16 @@
...
 import pandas as pd
+import torch.multiprocessing
 
 from weathergen.train.trainer import Trainer
 from weathergen.utils.config import Config, load_overwrite_conf, load_private_conf
 from weathergen.utils.logger import init_loggers
 
+# This strategy is required by the nvidia profiles to properly trace events in worker processes.
+# This may cause issues with logging. Alternative: "fork"
+torch.multiprocessing.set_start_method("spawn", force=True)
```
clessig: Do we know this doesn't have side effects, e.g. doesn't effect performance. I think the better place would be Trainer_Base::init_torch() and there we could also have an argument that enables it only when we run in profiling mode.

src/weathergen/datasets/anemoi_dataset.py
```diff
@@ -30,26 +35,26 @@ def __init__(
         assert len_hrs == step_hrs, "Currently only step_hrs=len_hrs is supported"
 
         # open  dataset to peak that it is compatible with requested parameters
-        self.ds = open_dataset(filename)
+        ds: Zarr = open_dataset(filename)
```
clessig: Is it really important here that it's a zarr store? I think that's an internal we shouldn't care about here.

src/weathergen/run_train.py
```diff
@@ -0,0 +1,8 @@
+"""
+The entry point to train the weathergen model.
+"""
+
+from weathergen import train
```
clessig, line 6: What is this needed for?

src/weathergen/datasets/anemoi_dataset.py
tjhunter: no it is not, this was a debugging statement. Will remove

src/weathergen/run_train.py
tjhunter: when launching the profiler, the entry point cannot be in a module init file. At least it complained loudly. Will comment.

src/weathergen/__init__.py
tjhunter: Good point, I moved it there

PR #137, opened 2025-03-27
Fixing bugs in forecasting

clessig:
- Data loader didn't load consecutive time steps
- Problem in internal roll-out with model parameter passing

### DIFF
--- a/src/weathergen/datasets/multi_stream_data_sampler.py
+++ b/src/weathergen/datasets/multi_stream_data_sampler.py
@@ -46,6 +46,7 @@ class MultiStreamDataSampler(torch.utils.data.IterableDataset):
         self.forecast_delta_hrs = (
             cf.forecast_delta_hrs if cf.forecast_delta_hrs > 0 else self.len_hrs
         )
+        assert self.forecast_delta_hrs == self.len_hrs, "Only supported option at the moment"
         self.forecast_steps = np.array(
             [cf.forecast_steps] if type(cf.forecast_steps) == int else cf.forecast_steps
         )
@@ -263,15 +264,10 @@ class MultiStreamDataSampler(torch.utils.data.IterableDataset):
                 idx = self.perms[idx_raw % self.perms.shape[0]]
                 idx_raw += 1
 
-                step_dt = self.len_hrs // self.step_hrs
-                step_forecast_dt = (
-                    step_dt + (self.forecast_delta_hrs * forecast_dt) // self.step_hrs
-                )
-
                 # TODO: this has to be independent of specific datasets
-                time_win1, time_win2 = (
+                time_win1, _ = (
                     self.streams_datasets[-1][0].time_window(idx),
-                    self.streams_datasets[-1][0].time_window(idx + step_forecast_dt),
+                    self.streams_datasets[-1][0].time_window(idx + self.len_hrs // self.step_hrs),
                 )
 
                 streams_data = []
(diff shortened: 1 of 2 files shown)


clessig:
@kacpnowak can you have a look. Thanks!

clessig:
> There's bug with FesomDataloader:
> 
> ```
> 3: Traceback (most recent call last):
> 3:   File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/__init__.py", line 388, in train_continue
> 3:     trainer.run(cf, None, args.run_id, args.epoch, True)
> 3:   File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/train/trainer.py", line 311, in run
> 3:     sources_size = self.dataset.get_sources_size()
> 3:                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 3:   File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/datasets/multi_stream_data_sampler.py", line 175, in get_sources_size
> 3:     ds[0].get_source_num_channels() + ds[0].get_geoinfo_size() + ds[0].get_coords_size()
> 3:                                                                  ^^^^^^^^^^^^^^^^^^^^^
> 3: AttributeError: 'FesomDataset' object has no attribute 'get_coords_size'. Did you mean: 'get_source_size'?
> ```

@kacpnowak : Thanks. But can we open a separate PR for this. Can you propose a fix based on the anemoi-dataset?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -263,15 +264,10 @@ def __iter__(self):
...
                 idx_raw += 1
 
-                step_dt = self.len_hrs // self.step_hrs
-                step_forecast_dt = (
-                    step_dt + (self.forecast_delta_hrs * forecast_dt) // self.step_hrs
-                )
-
                 # TODO: this has to be independent of specific datasets
-                time_win1, time_win2 = (
+                time_win1, _ = (
```
kctezcan, line 268: if we are leaving out the second variable, is there a reason for still creating a tuple? Would it not be enough to write:

 `time_win1 = self.streams_datasets[-1][0].time_window(idx)`

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -325,9 +321,15 @@ def __iter__(self):
                     for fstep in range(forecast_dt + 1):
                         # collect all sources
                         for _, ds in enumerate(stream_ds):
-                            (coords, geoinfos, target, times) = ds.get_target(
-                                idx + step_forecast_dt
+                            step_forecast_dt = (
+                                idx + (self.forecast_delta_hrs * fstep) // self.step_hrs
                             )
+                            _, time_win2 = (
```
kctezcan, line 327: similar to line 268, can we not just write?:

`time_win2  = self.streams_datasets[-1][0].time_window(step_forecast_dt) `

src/weathergen/model/model.py
```diff
@@ -585,6 +585,7 @@ def forward(self, model_params, batch, forecast_steps):
             # prediction
             preds_all += [
                 self.predict(
+                    model_params,
```
kctezcan, line 588: I assume this is a fix, since if the parameter was not being passed before, the predict function was not running?..

kctezcan (commented):
I cannot say much about the functional side, did not run/test the code, just some syntax comments

src/weathergen/datasets/multi_stream_data_sampler.py
clessig: That's correct. But let's address it in a separate PR. It's unrelated and is better handled when creating the base class for the data readers

Issue #149, opened 2025-04-01
Possibly a "silent" bug in the compute_loss


SavvasMel:
### What happened?

By reading the "compute_loss" I see that there is a scaling of the total loss taking place after the calculation of the loss for each channel (see first two rows).

```
val = val / ctr if (ctr > 0) else val
val_uw = val_uw / ctr if (ctr > 0) else val_uw

losses_all[j, i_obs] = val_uw
if self.cf.loss_fcts[j][0] == "stats" or self.cf.loss_fcts[j][0] == "kcrps":
    stddev_all[i_obs] = pred[:, mask_nan].std(0).mean().item()
# ignore NaNs so that training can continue even if one pred-net diverges
loss = loss + (
    (w * val * obs_loss_weight)
    if not torch.isnan(val)
    else torch.tensor(0.0, requires_grad=True)
)
```
The "val" is registered scaled in the loss here after also considering NaN values:

```
loss = loss + (
    (w * val * obs_loss_weight)
    if not torch.isnan(val)
    else torch.tensor(0.0, requires_grad=True)
)

```
However, after the losses for all streams and channels and metrics the "loss" is scaled a second time before it exits. See "loss/ctr" below:

```
return (
    loss / ctr,
    None
    if not log_data
    else [
        preds_all,
        targets_all,
        targets_coords_raw_rt,
        targets_times_raw_rt,
        targets_lens,
    ],
)
```

Is there a specific reason for that?

### What are the steps to reproduce the bug?

Just check the code.

### Version

v0.0

### Platform (OS and architecture)

JUWELS Booster

### Relevant log output

```shell

```

clessig:
Yes, this is a bug: there are two ctrs with the inner one aliasing the outer one (which is over output streams).

PR #150, opened 2025-04-01
Fix to bug in counting datasets contributing to loss

clessig:


clessig:
@SavvasMel could you please have a look if this fixes #149

SavvasMel:
@clessig Now ctr is over channels, ctr_target is over streams but **also over the forecasting steps**. If this is what you want then yes it is fixed.

Issue #127, opened 2025-03-25
implement simple sanity check run on new public ERA5 dataset


iluise:
### Is your feature request related to a problem? Please describe.

define a first simple sanity check run to be run by the users before each PR

PR #156, opened 2025-04-03
preliminary test script for the WeatherGenerator

iluise:
addressing issue: https://github.com/ecmwf/WeatherGenerator/issues/127

For the moment I have implemented it on the anemoi dataset as the data loader for the ERA5 dataset is not ready yet. 

- test `train`
- test `evaluation`
- test train and evaluation losses through the `metrics.py` json file.
- add checklist for PRs (to be tested).

Running all the tests takes a while now, even if:

- it **uses only wind** for `source` and `target` 
- it runs 100 steps for train.

currently getting some issues with `train_continue` so it will be added in another PR.


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/WeatherGenerator?pullRequest=156) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/WeatherGenerator?pullRequest=156) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/WeatherGenerator?pullRequest=156) it.</sub>

.github/pull_request_template.md
```diff
@@ -0,0 +1,47 @@
+## Description
```
clessig: Can we open a separate PR for this.

tests/test_metrics.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+    assert loss_metric is not None, (
+        "'stream.era5.loss_mse.loss_avg' metric is missing in metrics file"
+    )
+    assert loss_metric < 0.25, (
+        f"'stream.era5.loss_mse.loss_avg' is {loss_metric}, expected to be below 0.25"
+    )
+
+
+def test_gpu_performance_above_threshold(run_id):
```
clessig: Let's remove this. The GPU metrics are quite useless.

clessig:
CC @tjhunter

tjhunter:
Note by the way this is an integration test, I will have a few naming changes suggestions for when we also have other tests (ex: unit tests)

config/streams/streams_test/era5.yml
```diff
@@ -0,0 +1,36 @@
...
+  loss_weight : 1.
+  diagnostic : False
+  masking_rate : 0.6
+  masking_rate_none : 0.05
+  token_size : 32
+  embed : 
+    net : transformer
+    num_tokens : 1
+    num_heads : 8
+    dim_embed : 256
```
clessig: We could adjust some of the parameters here to make it train faster.

integration_tests/small1.py
```diff
@@ -0,0 +1,123 @@
...
+    yield
+    print("end fixture")
+
+
+@pytest.mark.parametrize("test_run_id", ["test_small1_" + commit_hash])
+def test_train(setup, test_run_id):
+    logger.info(f"test_train with run_id {test_run_id}")
+
+    train_with_args(
+        "--config=integration_tests/small1.yaml".split() #"--run_id=testsmall1 ".split()
```
clessig: Remove the tailing comment here?

integration_tests/small1.yaml
```diff
@@ -0,0 +1,9 @@
+streams_directory: "./config/streams/streams_test/"
+loss_fcts: [["mse", 1.0]]
```
clessig, line 4: Why can't we use the defaults here? loss_fcts and loss_fcts_val need to be set in the default config.

integration_tests/small1.py
```diff
@@ -0,0 +1,123 @@
...
+    )
+    assert_missing_metrics_file(test_run_id)
+    assert_train_loss_below_threshold(test_run_id)
+    assert_val_loss_below_threshold(test_run_id)
+    logger.info("end test_train")
+
+
+def load_metrics(run_id):
+    """Helper function to load metrics"""
+    file_path = f"./results/{run_id}/metrics.json"
```
clessig: This assumes we use ./results ./models. Should we use the global paths when specified?

integration_tests/small1.yaml
iluise: this was to overwrite whatever it is in the config and always ensure that the loss is only mse

config/streams/streams_test/era5.yml
iluise: Changed to 
```
num_heads: 4
dim_embed: 128
```

integration_tests/small1.py
iluise: uhm, how do you do this?

clessig: https://github.com/ecmwf/WeatherGenerator/blob/86715053551c8351d1cdb3ad36a2f111d1e1d8de/src/weathergen/train/trainer.py#L76

cf.run_path specifies the results dir.

tjhunter: The run_path is now set explicitly to `results` and `models` : https://github.com/ecmwf/WeatherGenerator/pull/156/files#diff-defc4a43fd25f1aeea2064a191d034711fdb43607e2328a4661921a180e14930R2

clessig: Better solution. Thanks!

src/weathergen/train/trainer.py
```diff
@@ -524,10 +527,7 @@ def train(self, epoch):
             self.perf_mem = self.ddp_average(torch.tensor([perf_mem])).item()
 
             self.log_terminal(bidx, epoch)
-            self.log(bidx, epoch)
-            # model checkpoint
-            if bidx % self.checkpoint_freq == 0:
-                self.save_model()
```
tjhunter, line 530: checkpoint is already done by epoch. the current logic confuses `final` model with `latest` model (not the same if a run aborts midflight`

src/weathergen/utils/config.py
```diff
@@ -118,14 +118,26 @@ def _load_private_conf(private_home: Path | None) -> OmegaConf:
...
+    if private_home is not None and private_home.is_file():
+        _logger.info(f"Loading private config from {private_home}.")
 
+    elif "WEATHERGEN_PRIVATE_CONF" in os.environ:
+        private_home = Path(os.environ["WEATHERGEN_PRIVATE_CONF"])
+        _logger.info(f"Loading private config fromWEATHERGEN_PRIVATE_CONF:{private_home}.")
+
+    elif env_script_path.is_file():
+        result = subprocess.run(
+            [str(env_script_path), "hpc-config"], capture_output=True, text=True
```
clessig, line 132: One runs the code not on the login nodes but the compute nodes. E.g. "as01r1b16" is a compute node but we only look for alogin. I think "as01" should be identical always but I am not entirely sure. For juwels it seems to be fine but we have the same issue on Leonardo.

clessig: So, on mn5 it's "as0?" where ? is a number.

tjhunter (approved):

src/weathergen/utils/config.py
tjhunter: I will check on leonardo. I think it is better to have this functionality merged in and then update teh list of nodes

Issue #147, opened 2025-03-31
MSE-trained roll out of model for 1 or 2 days on a small ERA5 dataset


SavvasMel:
### Is your feature request related to a problem? Please describe.

This is related to enhancement https://github.com/ecmwf/WeatherGenerator/issues/77.

Steps:

- [x] Pretrain the model on a small ERA5 dataset in 8 nodes in JUWELS Booster for 24 hours. The pretrained model is doing only 6h forecast. 
- [ ] Evaluate the model, plot some results to check the pretrained model.
- [ ] Finetune the model by using forecast steps of 6h to reach a forecast of 24h or 48h ahead.
- [ ] Do the analysis of those results.

### Additional context

Evaluation of the model is a bit tricky. The script "plot.py" will be used https://gitlab.jsc.fz-juelich.de/esde/WeatherGenerator-private/-/tree/main/teams/ecmwf?ref_type=heads.

SavvasMel:
I would like to ask a question regarding this part in the `train_continue()` here 

https://github.com/ecmwf/WeatherGenerator/blob/develop/src/weathergen/__init__.py.

 ```
    if cf.forecast_freeze_model:
        cf.with_fsdp = False
        import torch

        torch._dynamo.config.optimize_ddp = False
```

While I understand why we would need to freeze the model except from the forecast engine, what is the role of `cf.with_fsdp` and `torch._dynamo.config.optimize_ddp`? I would like to use `train_continue()` to train the model on many forecast steps but I get error when the model backpropagates. I am getting an error output very similar to this https://github.com/pytorch/pytorch/issues/146260.

**Comment by SavvasMel** [comment]
@clessig I am getting the error below under those setting in `train_continue()`: 

The error disappears if I set `cf.fe_num_blocks = 0` which mainly means that the forecast engine is turned off. Keep the rest of the settings the same but `cf.fe_num_blocks = 0` , the training part runs smoothly.

```
    cf.forecast_delta_hrs = 0  
    cf.forecast_steps = 1  
    cf.forecast_policy = "fixed" 
    cf.forecast_freeze_model = False
    cf.forecast_att_dense_rate = 1.0  # 0.25

    if cf.forecast_freeze_model:
        cf.with_fsdp = False
        import torch

        torch._dynamo.config.optimize_ddp = False

    cf.fe_num_blocks = 1
    cf.fe_num_heads = 16
    cf.fe_dropout_rate = 0.1
[...]

PR #158, opened 2025-04-04
Fixed some workarounds to torch FSDP

clessig:
These are no longer need with torch 2.6. This potentially also fixes some memory issues that were observed, cf #147. 

CC @kacpnowak @SavvasMel

### DIFF
--- a/src/weathergen/__init__.py
+++ b/src/weathergen/__init__.py
@@ -185,12 +185,6 @@ def train_continue() -> None:
         cf.forecast_freeze_model = True
         cf.forecast_att_dense_rate = 1.0  # 0.25
 
-        if cf.forecast_freeze_model:
-            cf.with_fsdp = False
-            import torch
-
-            torch._dynamo.config.optimize_ddp = False
-
         cf.fe_num_blocks = 8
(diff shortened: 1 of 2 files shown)


src/weathergen/__init__.py
kacpnowak, line 1: I have torch version: 2.6.0+cu124 and this issue is till present. I tried workaround showed [here](https://github.com/pytorch/pytorch/issues/100945) but it still crashes. It seems that we would need to implement FSDP2 to make it work with frozen tensors

clessig: Can you try out this branch. For me, train_continue work there with FSDP.

kacpnowak: It doesn't work. I got the expected error: ```ValueError: Must flatten tensors with uniform `requires_grad` when `use_orig_params=False```

kacpnowak: When setting `use_orig_params=True` I got `RuntimeError: mat1 and mat2 must have the same dtype, but got Float and Half`. After disabling MP I got `AssertionError: No inf checks were recorded for this optimizer.`

Issue #106, opened 2025-03-20
Introduce OmegaConf to streamline configuration


grassesi:
### Is your feature request related to a problem? Please describe.

Configuration needs to be merged and combined from different sources (default config, potential overwrites, private platform config, streams, ...). Currently the state of the Config object is often not well defined, since the attributes will be dynamically set. OmegaConf mitigates this issue: It provides YAML and primitive container (de)serialization routines and capabilities for merging.

### Describe the solution you'd like

In a first step OmegaConf should be used in the following steps to manage config from different sources:

1.  load base config:
     - previously trained model config (run_id is provided => train continue/evaluation):
     - default config (run_id is not provided => fresh training)
2. load/merge private config (see additional context)
3. load/merge streams config (currently Trainer_Base.init_streams)
4. overwrite with custom config:
     - from cli args (evalation)
     - from location specified with --config (training)
     - calculated at run time (fresh run_id, random seed, ...)
5. serialize config (see additional context)

In a following, support for static linting should be successively added through the means of dataclasses/structured configs

### Describe alternatives you've considered

serializing config to json (see context)

### Additional context

- There is currently no option to specify a private config when running the evaluation, is this by design? If not a new Issue should be raised.
- Currently most configuration code is in yaml, however run configs loaded for evaluation or when continuing training are expected to be json files. For the sake of consistency I would argue to move this to yaml (especially since the overwrite config is in yaml). All new run configs should be serialized to yaml, an option to deserialize json should be implemented as a backward compatibility measure. This can be later deprecated and removed if this need ceases to be relevant.

### Organisation

JSC

Issue #37, opened 2025-03-03
Request to Migrate Timestamp from Integer to Standard datetime Objects


kacpnowak:
### What maintenance does this project need?

The current implementation stores timestamps as integers in the `YYYYMMDDHHSS` format. This approach poses significant risks and limitations:  
1. Error-Prone: Manual parsing/formatting increases the risk of bugs (e.g., invalid dates, incorrect padding).  
2. Non-Standard: The format is not ISO-compliant, creating interoperability issues with external systems or libraries.  
3. Inefficient Operations: Date arithmetic (e.g., computing differences between timestamps) requires redundant conversions, adding unnecessary overhead.  
4. Poor Readability: Debugging or auditing timestamps in raw integer form is unintuitive and time-consuming.  

Example of current workaround:  
```python  
format_str = "%Y%m%d%H%M%S"
base_dt = datetime.datetime.strptime(str(base_yyyymmddhhmm), format_str)
self.start_dt = datetime.datetime.strptime(str(start), format_str)
self.end_dt = datetime.datetime.strptime(str(end), format_str)
```  

---  

### **Proposed Solution**  
Migrate all timestamp storage and operations to use Python’s native `datetime` objects (or `pandas.Timestamp`/`numpy.datetime64` where appropriate). Benefits include:  
- **Type Safety**: Leverage built-in validation and error handling.  
- **Standardization**: Align with widely accepted conventions.  
- **Readability**: Human-friendly representation and debugging.  
- **Ecosystem Compatibility**: Direct integration with pandas/numpy for analytics and serialization (e.g., JSON, databases).    

### Organisation

AWI

tjhunter:
It is certainly very error-prone. I suggest that we convert it to a well-known string format for the time being and that we do not store datetime objects for now in the configuration objects. This causes other issues because yaml -> timestamps is not a 1-1 conversion.

kacpnowak:
Actually it is 1-1 conversion. According to [PyYaml docs](https://pyyaml.org/wiki/PyYAMLDocumentation) !!timestamp tag is mapped directly to datetime.datetime

clessig:
@grassesi : We still specify dates in the config as "201301010000". I think it should be easy now to switch to ISO dates in the yaml. Could you have a look?

PR #159, opened 2025-04-04
Sgrasse/develop/issue 106

grassesi:
close issue #106


config/default_config.yml
```diff
@@ -0,0 +1,113 @@
...
+
+grad_clip: 5.0
+weight_decay: 0.1
+norm_type: "LayerNorm"
+nn_module: "te"
+
+start_date: 201301010000
+end_date: 202012310000
+start_date_val: 202101010000
+end_date_val: 202201010000
```
kacpnowak, line 98: It would be swap integers for strings when storing dates

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
+    # use OmegaConf.unsafe_merge if too slow
+    return OmegaConf.merge(base_config, private_config, overwrite_config)
+
+
+def load_overwrite_conf(overwrite_path: Path | None = None) -> Config:
     "Return the overwrite configuration."
+
     "If path is None, return an empty dictionary."
-    if not pth:
+    if overwrite_path is None or not overwrite_path.is_file():
```
kacpnowak, line 109: I think a warning should be produced here, so it won't be a silent error

kacpnowak (approved):
I think it's a great contribution. Looking forward to getting it merged

config/default_config.yml
tjhunter: @kacpnowak  what do you mean? 

In any case, for complicated changes such as this one, I suggest that we introduce the library change first and we fix the date format later. Otherwise we may break something along the way.

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
  'psutil',
  "flash-attn; sys_platform == 'linux'",
  "polars~=1.25.2",
+ "omegaconf>=2.3.0",
```
tjhunter: can you use `~=`? This pins the dependency

src/weathergen/__init__.py
```diff
@@ -165,16 +156,12 @@ def train_continue() -> None:
     )
 
     args = parser.parse_args()
-    # get the paths from the private config
-    private_cf = load_private_conf(args.private_config)
 
-    # load config if specified
-    model_path = private_cf["model_path"] if "model_path" in private_cf.keys() else "./models"
-    cf = Config.load(args.run_id, args.epoch, model_path)
+    if args.epoch == -2:
```
tjhunter: why -2 and not -1? that sounds magical. Also, I would change it to `<=-2`

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
 
-    def save(self, epoch=None):
-        path_models = Path(self.model_path)
-        # save in directory with model files
-        dirname = path_models / self.run_id
-        dirname.mkdir(exist_ok=True, parents=True)
+Config = OmegaConf
 
+
+def print_cf(config: Config):
```
tjhunter, line 30: docstring

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
+    for key, value in config.items():
+        if key != "streams":
+            print(f"{key} : {value}")
+        else:
+            for rt in value:
+                for k, v in rt.items():
+                    print("{}{} : {}".format("" if k == "reportypes" else "  ", k, v))
+
+
+def save(config: Config, epoch: int | None = None):
```
tjhunter: no default arg. docstring.

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
+    path_models = Path(config.model_path)
+    # save in directory with model files
+    dirname = path_models / config.run_id
+    dirname.mkdir(exist_ok=True, parents=True)
+
+    epoch_str = ""
```
tjhunter: I prefer this style which is more compatible with static analysis:

```py
if epoch is None:
  epoch_str = ""
elif epoch <= -1:
  epoch_str = "_latest"
else:
  ...
```

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
+        epoch_str = "_latest" if epoch == -1 else f"_epoch{epoch:05d}"
+    fname = dirname / f"model_{config.run_id}{epoch_str}.json"
+
+    json_str = json.dumps(OmegaConf.to_container(config))
+    with fname.open("w") as f:
+        f.write(json_str)
+
+
+def load_model_config(
+    run_id: str, epoch: int | None = None, model_path: str = "./models"
```
tjhunter: no default argument. I would suggest here to have `model_path: str | None` and then in the body:

`path_models = Path(model_path or "./models")`

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
-        with fname.open() as f:
-            json_str = f.readlines()
+    return OmegaConf.create(json.loads(json_str))
 
-        cf = Config()
-        cf.__dict__ = json.loads(json_str[0])
 
-        return cf
+def load_config(
+    private_home: Path | None = None,
```
tjhunter: no default args

tjhunter: also, this function seems to be the only one that users sohuld use, is that correct? I would prefix the other private functions with `_`. Ex: `load_overwrite_conf` -> `_load_overwrite_conf`

src/weathergen/utils/config.py
tjhunter: I would put a logging statement for both cases:
- loading a file / directory
- being passed none and indicating that no overwriting conf is loaded

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
-        return private_conf
+
+    private_cf = OmegaConf.load(private_home)
+    private_cf["model_path"] = (
+        private_cf["model_path"] if "model_path" in private_cf.keys() else "./models"
+    )
+    return private_cf
+
+
+def load_default_conf() -> Config:
```
tjhunter: private?

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
+    """Deserialize default configuration."""
+    return OmegaConf.load(DEFAULT_CONFIG_PTH)
+
+
+def load_streams(streams_directory: Path) -> list[Config]:
+    if not streams_directory.is_dir():
+        _logger.warning(f"Streams directory {streams_directory} does not exist.")
```
tjhunter: I think it should even be an error. This function should not be called on something that is not a directory

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
+
+    # read all reportypes from directory, append to existing ones
+    streams_directory = streams_directory.absolute()
+    _logger.info(f"Reading streams from {streams_directory}")
+
+    # append streams to existing (only relevant for evaluation)
+    streams = []
+    for config_file in sorted(streams_directory.rglob("*.yml")):
```
tjhunter, line 152: thanks for sorting

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
+        try:
+            stream_name, stream_config = [*OmegaConf.load(config_file).items()][0]
+        except yaml.scanner.ScannerError:
+            _logger.warning(f"Invalid yaml file: {config_file}")
+            continue
```
tjhunter: why not raise the exception and stop here?

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
+        _logger.info(f"loaded stream config: {stream_name}")
+
+    # sanity checking (at some point, the dict should be parsed into a class)
+    # check if all filenames accross all streams are unique
+    rts = [rt["filenames"] for rt in streams]
+    rts = list(itertools.chain.from_iterable(rts))
+    if len(rts) != len(set(rts)):
+        _logger.warning("Duplicate reportypes specified.")
+
+    # return OmegaConf.create({"streams": streams})
```
tjhunter: remove

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
+            _logger.warning(f"Invalid yaml file: {config_file}")
+            continue
+
+        stream_config.name = stream_name
+        streams.append(stream_config)
+        _logger.info(f"loaded stream config: {stream_name}")
+
+    # sanity checking (at some point, the dict should be parsed into a class)
+    # check if all filenames accross all streams are unique
+    rts = [rt["filenames"] for rt in streams]
```
tjhunter: any reason for not doing: `rts = [filename for rt in streams for filename in rt["filenames"]]` ?

src/weathergen/utils/train_logger.py
```diff
@@ -19,7 +19,7 @@
 import numpy as np
 import polars as pl
 
-from weathergen.utils.config import Config
+import weathergen.utils.config as config
```
tjhunter, line 22: why not do `from weathergen.utils.config import Config, load_model_config`?

tjhunter (changes requested):
@grassesi thanks a lot for this change, this is going to make config much easier. 

I have a couple of comments on the style, feel free to push back. The main high level suggestions:

- mark the functions that should be private with `_` prefix
- in general, I think we should refrain from default arguments: they are great for user-facing libraries but they make internal refactoring and error checking harder

Looking forward to this PR !

src/weathergen/__init__.py
clessig: Negative values shouldn't be used. We could use -1 as convention for the latest checkpoint but None seems more natural (although it remains a convention).

src/weathergen/__init__.py
```diff
@@ -235,13 +222,13 @@ def train() -> None:
     )
     parser.add_argument(
         "--private_config",
-        type=str,
+        type=Path,
         default=None,
         help="Path to private configuration file for paths",
     )
     parser.add_argument(
         "--config",
```
clessig, line 228: This should be called experiment_config to be more explicit and delineate it from the default config.

src/weathergen/__init__.py
```diff
@@ -235,13 +222,13 @@ def train() -> None:
     )
     parser.add_argument(
         "--private_config",
```
clessig, line 222: Does this provide only the paths? Then it should be called path_config? It's in the private repo but this shouldn't matter.

src/weathergen/__init__.py
```diff
@@ -251,155 +238,17 @@ def train() -> None:
...
-    cf.fe_num_heads = 16
-    cf.fe_dropout_rate = 0.1
-    cf.fe_with_qk_lnorm = True
-
-    cf.healpix_level = 5
-
-    # working precision
-    cf.with_mixed_precision = True
-    cf.with_flash_attention = True
+    cf = config.load_config(args.private_config, overwrite_path=args.config)
```
clessig: Where is the default_config.yml loaded?

src/weathergen/__init__.py
```diff
@@ -251,155 +238,17 @@ def train() -> None:
...
+
     if cf.with_flash_attention:
```
clessig, line 241: We can leave it for this PR but I think this checking should go elsewhere ultimately.

src/weathergen/__init__.py
```diff
@@ -251,155 +238,17 @@ def train() -> None:
...
-
-    # overwrite parameters from private config
-    for k, v in private_cf.items():
-        setattr(cf, k, v)
-    cf.data_path = private_cf["data_path_anemoi"]  # for backward compatibility
-
-    # overwrite parameters from overwrite config
-    for k, v in overwrite_cf.items():
-        setattr(cf, k, v)
+    cf.data_path = cf["data_path_anemoi"]  # for backward compatibility
```
clessig: Please remove. This should no longer be needed.

src/weathergen/train/trainer.py
```diff
@@ -75,20 +70,21 @@ def init(
         self.init_ddp(cf)
 
         # read configuration of data streams
-        cf = self.init_streams(cf, run_id_contd)
+        # FIXME should concat streams together really
```
clessig: Could you elaborate what is meant by the FIXME? This is too big to do with this PR?

src/weathergen/train/trainer.py
```diff
@@ -75,20 +70,21 @@ def init(
...
         path_run = Path(cf.run_path) / cf.run_id
         path_model = Path(cf.model_path) / cf.run_id
         if self.cf.rank == 0:
-            path_run.mkdir(exist_ok=True)
-            path_model.mkdir(exist_ok=True)
+            path_run.mkdir(exist_ok=True, parents=True)
+            path_model.mkdir(exist_ok=True, parents=True)
             # save config
-            cf.save()
+            config.save(cf)
```
clessig: This has changed. We need to rebase to catch this.

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
-                for rt in value:
-                    for k, v in rt.items():
-                        print("{}{} : {}".format("" if k == "reportypes" else "  ", k, v))
 
-    def save(self, epoch=None):
-        path_models = Path(self.model_path)
-        # save in directory with model files
-        dirname = path_models / self.run_id
-        dirname.mkdir(exist_ok=True, parents=True)
+Config = OmegaConf
```
clessig, line 27: Why do we want to have this typedef? Is this pythonic?

src/weathergen/train/trainer_base.py
```diff
@@ -111,47 +107,6 @@ def init_ddp(cf):
 
         return
 
-    ###########################################
-    @staticmethod
-    def init_streams(cf: Config, run_id_contd):
-        if not hasattr(cf, "streams_directory"):
-            return cf
-
-        # use previously specified streams when continuing a run
```
clessig, line 120: Where is this check in the new version?

src/weathergen/train/trainer_base.py
```diff
@@ -185,10 +140,10 @@ def ddp_average(self, val):
 
 ####################################################################################################
 if __name__ == "__main__":
```
clessig: The whole __main__ should be removed.

src/weathergen/utils/config.py
clessig, line 30: Why is this called print_cf() and below just save()?

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
+        base_config.run_id = get_run_id()
+    else:
+        base_config = load_model_config(run_id, epoch, private_config["model_path"])
 
-def load_overwrite_conf(pth: str) -> dict:
+    # use OmegaConf.unsafe_merge if too slow
+    return OmegaConf.merge(base_config, private_config, overwrite_config)
+
+
+def load_overwrite_conf(overwrite_path: Path | None = None) -> Config:
```
clessig: Why do we need this function still?

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
+        except yaml.scanner.ScannerError:
+            _logger.warning(f"Invalid yaml file: {config_file}")
+            continue
+
+        stream_config.name = stream_name
+        streams.append(stream_config)
+        _logger.info(f"loaded stream config: {stream_name}")
+
+    # sanity checking (at some point, the dict should be parsed into a class)
+    # check if all filenames accross all streams are unique
```
clessig: Let's remove this. Difficult to say at the moment what is sensible but I wouldn't raise an error.

src/weathergen/utils/config.py
```diff
@@ -7,98 +7,162 @@
...
+
+    # read all reportypes from directory, append to existing ones
+    streams_directory = streams_directory.absolute()
+    _logger.info(f"Reading streams from {streams_directory}")
+
+    # append streams to existing (only relevant for evaluation)
+    streams = []
+    for config_file in sorted(streams_directory.rglob("*.yml")):
+        try:
+            stream_name, stream_config = [*OmegaConf.load(config_file).items()][0]
```
clessig, line 157: Where does stream_name come here?

clessig (commented):
How have these changes been tested? The minimum would be:
- run train
- run train_continue
- run evaluate
and see if it's fully functional and if the workflow is right. Since the config is a critical part we should also try to catch corner cases in the workflow and handling.

config/default_config.yml
kacpnowak: Just have this as string, eg “2020-01-01” I don’t think it’s complicated. One just has to do conversion to desired type when consuming this. Currently we already do this in evaluation https://github.com/ecmwf/WeatherGenerator/blob/b03498669dcf70aa2fefffed3f232ef442778f23/src/weathergen/__init__.py#L37

clessig: We need to get it consistent everywhere, and make sure it's working as expected and intended. It is cleaner to handle it in a separate PR. But we can do it right after we get this merged.

src/weathergen/utils/config.py
grassesi: Stream config schema is {stream_name: stream_config} where stream_config
itself is a dict containing the actual options. stream_name needs to be
added to this dict since only stream_config will be further processed.

Example:
```
ERA5 :
  type : anemoi
  filenames : ['aifs-ea-an-oper-0001-mars-o96-1979-2022-6h-v6.zarr']
  ...
```
`stream_name` is `"ERA5"`.

I will add the above explanation as a comment to the code

clessig: And the name is taken from the yaml file?

src/weathergen/__init__.py
grassesi: its loaded within `config.load_config`

src/weathergen/utils/config.py
grassesi: From multiple different stream configs, only one could be invalid. The question here is if we want to abort if one config is invalid but the otheres possibly are?

src/weathergen/train/trainer_base.py
grassesi: I didnt want to remove functionallity so I just made sure it stays compatible, but I am happy to remove it if its dead code. Maybe this should be documented as a separate PR?

config/default_config.yml
grassesi: I agree we should make a separate PR taking also the discussion in #37 into account.

src/weathergen/utils/config.py
grassesi: I put it there to not break current type annotations in the rest of the code. It also signals that the underlying type of the configuration object should be (for now) considered an implementation detail.

src/weathergen/utils/config.py
grassesi: calling the method `print` would shadow the inbuild `print` method for the rest of the module, leading to faulty behaviour whereever the inbuild `print` should be used (eg. in this method itself). Alternatively it could be called `print_`.

src/weathergen/utils/config.py
grassesi: Do you mean the sanity checking? I am happy to remove that.

src/weathergen/train/trainer.py
grassesi: Already did that

src/weathergen/train/trainer.py
grassesi: The FIXME is missleading, I will remove that. I added it for me to mark that the behavior has slightly changed:
- before: If the `streams` attribute is already initialized with a list, all newly deserialized stream config get appended to this list, if its not initialized it will be.
- after: all stream configs are deserialized and added to a list, this list is then assigned to the `streams` attribute. Previously present entries in this list will be overwritten.

However there is no execution path that would show the behavior change:
- if `run_id_contd` the configuration was always returned as is: This happens for every invocation of `evaluate` and `train_continue`.
- if `run_id_contd` is `None` this implies fresh training, therefor the config wouldnt have the `streams` attribute yet.

src/weathergen/utils/config.py
grassesi: yes since #167 reintroduced the `--config` option allowing to specify a overwrite config

src/weathergen/__init__.py
grassesi: Was a quick hack to run `train_continue` with args.epoch == None. This is nonsensical epoch should never be None since you will need a model checkpoint to continue training which is only generated after one epoch. Will remove.

src/weathergen/__init__.py
grassesi: It also contains secrets.

src/weathergen/__init__.py
grassesi: I would avoid renaming it since the focus of this PR is mostly changing implementation details of the configuration rather then introducing breaking changes that will require broader adjustment to a new CLI interface.

src/weathergen/train/trainer_base.py
grassesi: I removed it since I did not see any valid reason it should be missing:
- if training is run from scratch, it is required for the input data.
- if training is continued or evaluation is run, it will be present since the loaded config will have it.

Please correct me if I am wrong with my reasoning here.

src/weathergen/utils/train_logger.py
grassesi: I tried to be consistent with my imports, if you want to I can change it.

src/weathergen/utils/config.py
clessig: Yes, let's remove the sanity checking.

tjhunter (commented):
I don't have further comments but I want to make sure the code has run at least an epoch on a cluster

tjhunter (approved):

clessig:
> > I don't have further comments but I want to make sure the code has run at least an epoch on a cluster
> 
> Can we please at least run train (for 1 epoch), train_continue, evaluate to see that the whole workflow works.

Can we implement the following:

`diff --git a/src/weathergen/__init__.py b/src/weathergen/__init__.py
index 0a52c21..d20a198 100644
--- a/src/weathergen/__init__.py
+++ b/src/weathergen/__init__.py
@@ -132,7 +132,7 @@ def train_continue() -> None:
         "--epoch",
         type=int,
         required=False,
-        default=-1,
+        default=None,
         help="epoch where to continue run",
     )
     parser.add_argument(
diff --git a/src/weathergen/model/model.py b/src/weathergen/model/model.py
index 1acdf28..f54cb50 100644
--- a/src/weathergen/model/model.py
+++ b/src/weathergen/model/model.py
@@ -536,9 +536,9 @@ class Model(torch.nn.Module):
         print("-----------------")
 
     #########################################
-    def load(self, run_id, epoch=-1):
+    def load(self, run_id, epoch=None):
         path_run = Path(self.cf.model_path) / run_id
-        epoch_id = f"epoch{epoch:05d}" if epoch != -1 else "latest"
+        epoch_id = f"epoch{epoch:05d}" if epoch is not None else "latest"
         filename = f"{run_id}_{epoch_id}.chkpt"
 
         params = torch.load(
`

Then train -> train_continue -> evaluate runs and it's ready to be merged.

src/weathergen/__init__.py
```diff
@@ -165,16 +156,10 @@ def train_continue() -> None:
...
-    model_path = private_cf["model_path"] if "model_path" in private_cf.keys() else "./models"
-    cf = Config.load(args.run_id, args.epoch, model_path)
+    cf = config.load_config(args.private_config, args.run_id, args.epoch, None)
 
     # track history of run to ensure traceability of results
-    if "run_history" not in cf.__dict__:
-        cf.run_history = []
     cf.run_history += [(cf.run_id, cf.istep)]
 
     #########################
```
clessig, line 165: These overwrites should in the future also come from a yaml file. For evaluate there should in the future also be an overwrite config although all standard parameters should be exposed as command line arguments. 

I would propose to merge as-is and immediately do a follow-up PR to address this.

src/weathergen/__init__.py
```diff
@@ -235,13 +220,13 @@ def train() -> None:
...
+        type=Path,
         default=None,
         help="Path to private configuration file for paths",
     )
     parser.add_argument(
         "--config",
-        type=str,
+        type=Path,
         default=None,
         help="Path to private configuration file for overwriting the defaults in the function body. Defaults to None.",
```
clessig: Should read: 

"Optional experiment specific configuration file"

src/weathergen/utils/config.py
```diff
@@ -8,97 +8,159 @@
...
+            # itself is a dict containing the actual options. stream_name needs to be
+            # added to this dict since only stream_config will be further processed.
+            stream_name, stream_config = [*OmegaConf.load(config_file).items()][0]
+        except yaml.scanner.ScannerError as e:
+            msg = f"Invalid yaml file while parsing stream configs: {config_file}"
+            raise RuntimeError(msg) from e
+
+        stream_config.name = stream_name
+        streams.append(stream_config)
+        _logger.info(f"loaded stream config: {stream_name}")
```
clessig: For consistency this should be "Loaded" instead of "loaded"

clessig (approved):

src/weathergen/__init__.py
grassesi: My Idea would have been to move management of run_history also into the `config.load_config` method, so that it is always assured that records are kept properly.

clessig: I didn't mean run_history but what comes after line 165: the forecast parameters. Sorry, could only tag this line.

The whole topic of experiment tracking, including maintaining the lineage of runs, will be one of the next things we tackle. But let's address it in a separate PR.

tjhunter:
This PR is 99% ready and other important PR's are depending on it. Let's merge it and address the final comment as a follow up.

PR #171, opened 2025-04-09
Misc fixes

clessig:
## Type of Change

-   [ ] Documentation update

## Issue Number

- Fix bug in handling of forecast step in model 
- Fixed bug in logging where not the final config was written (which led to problems especially with evaluate)
- Fixed problem with forecast_policy 'random'

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/datasets/multi_stream_data_sampler.py
+++ b/src/weathergen/datasets/multi_stream_data_sampler.py
@@ -46,9 +46,9 @@ class MultiStreamDataSampler(torch.utils.data.IterableDataset):
         self.forecast_steps = np.array(
             [cf.forecast_steps] if type(cf.forecast_steps) == int else cf.forecast_steps
         )
-        if cf.forecast_policy is not None :
-            if self.forecast_steps.max() == 0 :
-                logger.warning( "forecast policy is not None but number of forecast steps is 0.")
+        if cf.forecast_policy is not None:
+            if self.forecast_steps.max() == 0:
+                logger.warning("forecast policy is not None but number of forecast steps is 0.")
         self.forecast_policy = cf.forecast_policy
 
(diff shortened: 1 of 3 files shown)


src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -39,17 +39,16 @@ def __init__(self, cf, start_date, end_date, batch_size, samples_per_epoch, shuf
...
         self.forecast_delta_hrs = (
             cf.forecast_delta_hrs if cf.forecast_delta_hrs > 0 else self.len_hrs
         )
         assert self.forecast_delta_hrs == self.len_hrs, "Only supported option at the moment"
         self.forecast_steps = np.array(
             [cf.forecast_steps] if type(cf.forecast_steps) == int else cf.forecast_steps
         )
+        if cf.forecast_policy is not None:
+            if self.forecast_steps.max() == 0:
+                logger.warning("forecast policy is not None but number of forecast steps is 0.")
```
tjhunter, line 51: the message is not accurate: it is max number of forecast steps.

tjhunter (approved):
Looks good. Noting this part of the configuration is missing documentation.

src/weathergen/datasets/multi_stream_data_sampler.py
kacpnowak: Also if forecast steps is `int`, which is required when `fixed` policy is used, an `AttributeError` will be raised

clessig: No, line 50/46 converts it to a np.array

kacpnowak: Just noticed it, my bad

PR #178, opened 2025-04-14
[127] Merging with Ilaria's own work

tjhunter:
Merging my branch with @iluise  main PR


iluise:
Hi Tim, the changes look great. I have just two comments:

- the local 'results' and 'models' folders get full very quickly and to easy the shaking of information in AtmoRep we were using common results and models folders on $SCRATCH symlinked to the local repo. In this case having the same name for the tests might create conflicts if two people will test the code at the same time. Can we add e.g. the username (ideal, but difficult on the different hpcs I think) or some additional tag?

- why can't we merge `evaluate` and `evaluate_from_args` and same for train? 

p.s. did you test the code or it needs testing?

Issue #198, opened 2025-04-26
Use of path instead of string


clessig:
### Is your feature request related to a problem? Please describe.

Here we still use str for path handling: https://github.com/ecmwf/WeatherGenerator/blob/7b8c8d8f88c13321b2e155c8cb1540c50f92c914/src/weathergen/datasets/multi_stream_data_sampler.py#L84

### Describe the solution you'd like

This should be pathlib.Path. It would be good to have the option to specify a absolute path in the config, i.e. if fname is an absolute path an not just a file name then it should be used unchanged.

CC: @grassesi

grassesi:
I will convert it to pathlib, however `fname` is a value within `filenames` attribute of a stream config. I would assume stream configs should not contain platform dependent information. This information should be contained instead in the private config (which it is => the `data_path`). The ability to specify an absolute path in the stream config would break all assumptions associated with this design. This is why I don't see the use case for checking if `fname` is an absolute path.

clessig:
We have experiment specific config overwrites. These are not commited. Multiple people have at the moment absolute paths in there and it's a very convenient feature for development; we just don't have the well organized dataset with everything in one place yet--and not for the forseeable future.

grassesi:
I think that can be closed now that #199 is merged

grassesi:
This issue is not present anymore in develop

Issue #192, opened 2025-04-21
Empty stream file / or fully commented stream file break code


clessig:
### What happened?

Empty stream file / or fully commented stream file breaks with:

```
Traceback (most recent call last):
  File "/gpfs/home/ecm/ecm327663/WeatherGenerator/src/weathergen/__init__.py", line 248, in train
    trainer.run(cf)
  File "/gpfs/home/ecm/ecm327663/WeatherGenerator/src/weathergen/train/trainer.py", line 140, in run
    self.init(cf, run_id_contd, epoch_contd, run_id_new)
  File "/gpfs/home/ecm/ecm327663/WeatherGenerator/src/weathergen/train/trainer.py", line 73, in init
    cf.streams = config.load_streams(Path(cf.streams_directory))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfs/home/ecm/ecm327663/WeatherGenerator/src/weathergen/utils/config.py", line 157, in load_streams
    stream_name, stream_config = [*OmegaConf.load(config_file).items()][0]
                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
> /gpfs/home/ecm/ecm327663/WeatherGenerator/src/weathergen/utils/config.py(157)load_streams()
-> stream_name, stream_config = [*OmegaConf.load(config_file).items()][0]
```

CC @grassesi  @tjhunter 

### What are the steps to reproduce the bug?

Add empty / fully commented out stream file to stream directory.

### Version

develop

### Platform (OS and architecture)

Linux

### Relevant log output

```shell

```

### Organisation

ECMWF

grassesi:
What is the expected behavior here? From my point of view this is still an erroneous  state, therefore I would just catch the exception but raise a more understandable/actionable exception instead.

grassesi:
This should be also closed

clessig:
> This should be also closed

But the PR isn't merged yet.

clessig:
> I just realized sorry :|

There are some minor points from Tim; once these are addressed I can merge and then the issues hopefully close automatically.

PR #199, opened 2025-04-30
Sgrasse/develop/issue 189 198 192

### LINKED ISSUE #189: Remove print statement when loading config

https://github.com/ecmwf/WeatherGenerator/blob/86715053551c8351d1cdb3ad36a2f111d1e1d8de/src/weathergen/utils/config.py#L81

This prints the unformatted config dict; this seems a debug message and it's not useful for a user. It should be removed or a formatted version should be printed.

@tjhunter @grassesi 

### Organisation

ECMWF

grassesi:
## Description

Multiple small fixes

## Type of Change

-   [ ] Documentation update

## Issue Number

closes #189 #198 #192 

## Code Compatibility

### Dependencies

### Documentation


src/weathergen/utils/config.py
```diff
@@ -170,6 +169,9 @@ def load_streams(streams_directory: Path) -> list[Config]:
         except yaml.scanner.ScannerError as e:
             msg = f"Invalid yaml file while parsing stream configs: {config_file}"
             raise RuntimeError(msg) from e
+        except IndexError as e:
+            msg = f"Parsed stream configuration file is empty: {config_file}"
```
clessig: An empty stream config file should generate a warning but not an error. It's too useful for development to disallow it.

clessig (commented):
Looks good for me; thanks for the work. Just a minor point about empty stream config files.

src/weathergen/utils/config.py
grassesi: This would be inconsistent with the handling of a malformed (invalid yaml) error. I can change it but then I would also change the above exception to a warning to have consistent behavior

clessig: There is nothing inconsistent with one generating a warning and one an error. An empty yaml file is in my opinion as valid as an empty string or an empty set; degenerate but valid and useful.

grassesi: what do you think @tjhunter ?

tjhunter: Personally, I struggle to see a good case for an empty stream config. I would rather have an error now and then someone later complaining that this should be allowed, rather than be permissive and people not realizing they are running silently garbage configuration (no one checks their logs unless they have an error)

clessig: I am currently copying files in and out of the stream directory in my experiments--this is really error prone and annoying. Being able to comment the content of a file in and out is much more convenient and safer.

clessig: > Personally, I struggle to see a good case for an empty stream config. I would rather have an error now and then someone later complaining that this should be allowed, rather than be permissive and people not realizing they are running silently garbage configuration (no one checks their logs unless they have an error)

On the error: yes, there should be an error if the stream config is garbage (and non-empty). An empty one is not garbage, worth a warning but it's just something that is useful and needed.

grassesi: > I am currently copying files in and out of the stream directory in my experiments--this is really error prone and annoying. Being able to comment the content of a file in and out is much more convenient and safer.

Maybe we should have a better mechanism to select experiment with different stream configs

grassesi:
PR is now tested successfully tested on juwels

src/weathergen/utils/config.py
clessig: I have multiple directories with different stream configs. That's great for scientific experiments but not for model development.

tjhunter: Ok I get it about empty vs commented out. No strong opinion then.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -66,49 +67,38 @@ def __init__(self, cf, start_date, end_date, batch_size, samples_per_epoch, shuf
...
+                        dataset = AnemoiDataset
+                        datapath = cf.data_path_anemoi
+                    case "fesom":
+                        dataset = FesomDataset
+                        datapath = cf.data_path_fesom
+                    case "atmorep":
+                        dataset = AtmorepDataset
+                        datapath = cf.data_path_anemoi
+                    case _:
+                        msg = f"Unsupported stream type {stream_info[type]}"
```
tjhunter: typo: `stream_info['type']`

src/weathergen/datasets/obs_dataset.py
```diff
@@ -17,19 +18,14 @@
...
-        # normalize: bool = True,
-        # select: list[str] = None,
         start: int,
         end: int,
         len_hrs: int,
         step_hrs: int,
-        filename: str,
+        filename: Path,
         stream_info: dict,
+        # normalize: bool = True, # never called
```
tjhunter: please remove if unusde

src/weathergen/utils/config.py
tjhunter: But please put a comment in the comment with the explanation of accepting it (empty commented files). Othrewise,someone will have the temptation to remove it

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -66,49 +67,38 @@ def __init__(self, cf, start_date, end_date, batch_size, samples_per_epoch, shuf
...
+                        raise ValueError(msg)
+
+                datapath = pathlib.Path(datapath)
+                if datapath.is_file():
+                    kwargs["filename"] = datapath
                 else:
-                    assert False, "Unsupported stream type {}.".format(stream_info["type"])
+                    kwargs["filename"] = pathlib.Path(datapath) / fname
```
tjhunter: can you add a check that this file exists? it should a file if I am correct?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -66,49 +67,38 @@ def __init__(self, cf, start_date, end_date, batch_size, samples_per_epoch, shuf
...
+                ds = dataset(**kwargs)
```
tjhunter: can you log an info statement along the line of 'opening stream config: {filename} with constructor {dataset}'

also, rather adding filename to the kwargs dict, you can just use it:

```py
if ...:
  filename=...
else:
  filename=...
dataset(filename=filename, **kwargs)
```

this is cleaner in my opinion

tjhunter (approved):
It looks good on my side with a few comments, but please make sure that you are testing it is first before merging.

src/weathergen/datasets/obs_dataset.py
clessig: Please leave the ObsDataset class as is at the moment. I will take care of it when I implement the base class.

grassesi: I will remove it for now, it can be easily readded when there is need

src/weathergen/datasets/multi_stream_data_sampler.py
grassesi: I can only check if its a valid path, since input data that is stored as a zarr DirectoryStore will be a directory. Maybe this is something that should be delegated to each individual dataset class

clessig (approved):

Issue #185, opened 2025-04-19
Overwrite any config argument on the command line


clessig:
### Is your feature request related to a problem? Please describe.

Currently, only predefined config arguments can be specified/overwritten with command line args. This should be possible for all config args

### Describe the solution you'd like

E.g `uv run train --loader_num_workers=0` should work, analogously for any other config param

clessig:
@grassesi : Could you look into this?

grassesi:
Sure, I think it should be possible to collect anything that can not be interpreted by the CLI and see if it can be interpreted as overrides to config parameters instead

tjhunter:
Why would we need that compared to just modifying the config files? We already have multiple layers of configuration files.

Also, regarding the design of the format: we should not invent our own language to specify fields in a json/yaml. Look at the syntax of jq, which is designed for that: https://jqlang.org/tutorial/

I recommend `--param 'x.y.z=3'` this is more flexible

grassesi:
Im using [`OmegConf.from_cli`](https://omegaconf.readthedocs.io/en/2.3_branch/usage.html#from-command-line-arguments) for parsing arguments not recognized by argparse. The syntax used by OmegaConf is as far as I could see the same you mentioned

clessig:
> Why would we need that compared to just modifying the config files? We already have multiple layers of configuration files.
> 
> Also, regarding the design of the format: we should not invent our own language to specify fields in a json/yaml. Look at the syntax of jq, which is designed for that: https://jqlang.org/tutorial/
> 
> I recommend `--param 'x.y.z=3'` this is more flexible

My concrete use case is setting num_loader_workers=0 when I do model development on the data loader part. Yes, it could be changed in the config but it's more natural to do it on the command line since it flips forth and back between runs. Same for other parameters: it should be mainly useful for model development, but it's an important use case for me :)

PR #221, opened 2025-05-07
Sgrasse/develop/issue 185

grassesi:
## Type of Change

-   [x] New feature (non-breaking change which adds functionality) (hopefully)
-   [ ] Documentation update

## Issue Number

#185 

## Code Compatibility

### Dependencies

### Documentation

-   [ x I have added comments to my code, particularly in hard-to-understand areas


clessig:
@grassesi : there's still a failing check here. Can you fix this.

src/weathergen/utils/config.py
```diff
@@ -99,18 +113,41 @@ def load_config(
...
-    else:
-        _logger.info(f"Loading overwrite config from {overwrite_path}.")
-        return OmegaConf.load(overwrite_path)
+    Args:
+        arg_list: items in this list should be of the form: parent_obj.nested_obj=value
+    """
+    return OmegaConf.from_cli(arg_list)
+
+
+def _load_overwrite_conf(overwrite: Path | dict | OmegaConf) -> OmegaConf:
```
tjhunter: add `| None` to make explicit the empty option. Then the catch all `_` can throw a ValueError

src/weathergen/utils/config.py
```diff
@@ -87,10 +87,24 @@ def load_config(
...
+    Merge config information from multiple sources into one run_config.
+
+    Args:
+        private_home: Configuration file containing platform dependent information and secretes
+        run_id: Run/model id of pretrained WeatherGenerator model to continue training or evaluate
+        epoch: epoch of the checkpoint to load.
+        *overwrites: Additional overwrites from different sources
+
+    Note: The order of precendence for merging the final config is in ascending order:
+        - base config (either default config or loaded from previuos run)
```
tjhunter: typo: previous

src/weathergen/utils/config.py
```diff
@@ -87,10 +87,24 @@ def load_config(
...
+        - private config
+        - overwrites (also in ascending order)
```
tjhunter, line 117: this is starting to be very complicated. we should simplify our config system eventually: predictability is more important than convenience

src/weathergen/__init__.py
```diff
@@ -138,82 +68,51 @@ def evaluate_from_args(argl: list[str]):
...
-        help="Fine tune for forecasting. It overwrites some of the Config settings.",
-    )
-
-    args = parser.parse_args()
+    parser = cli.get_continue_parser()
+    args, additional_args = parser.parse_known_args()
 
-    cf = config.load_config(args.private_config, args.run_id, args.epoch, None)
+    if args.finetune_forecast:
+        finetune_overwrite = dict(
```
tjhunter, line 70: ok, thanks for the cleanup.

src/weathergen/utils/cli.py
```diff
@@ -0,0 +1,143 @@
+import argparse
+from pathlib import Path
+
+
+def get_train_parser():
+    parser = argparse.ArgumentParser(allow_abbrev=False)
+    parser.add_argument(
```
tjhunter: note that you can factor out the code here:

```

def _add_private_conf(parser: ...):
  parser.add_argument( "--private_config",
        type=Path,
        default=None,
        help="Path to private configuration file for paths",
    )

def get_train_parser():
  ...
  _add_private_conf(parser)

def get_eval_parser():
  ...
  _add_private_conf(parser)
```

tjhunter: (for run_id, private_config, config)

src/weathergen/utils/config.py
```diff
@@ -99,18 +113,41 @@ def load_config(
...
+    """
+    return OmegaConf.from_cli(arg_list)
+
+
+def _load_overwrite_conf(overwrite: Path | dict | OmegaConf) -> OmegaConf:
+    "Convert different sources into configs that can be used as overwrites."
+
+    "If source can not be converted return an empty config."
+    match overwrite:  # match the type
+        case Path():
```
tjhunter, line 152: interesting, I had not seen this `()` syntax to access the type.

tjhunter (approved):
@grassesi  it looks good, but
- i have some refactoring suggestions
- could you try first with the integration test.

clessig:
Since this is a config change that is affecting everyone and that is difficult to test, I think we should test more than the standard integration test and also make sure the workflow is right (especially concerning the merging of the configs). Happy to help with this.

src/weathergen/utils/config.py
grassesi: I know, it was the simplest solution I could think of without exploding the interface but still allowing the needed flexibilty. Do you have something different in mind? I am happy to discuss this.

src/weathergen/__init__.py
clessig: I think the better solution would actually be to specify this through an overwrite config. The specific settings such as forecast_policy are really options on chooses from an it shouldn't be in the code.

src/weathergen/utils/config.py
grassesi: As I have implemented it now, `None` is not an expected type and will raise an `ValueError` (the same as all unexpected/unknown types). As such I would not put in in the functions signature. Please tell how you feel about that.

src/weathergen/__init__.py
grassesi: Yes that would be conceptually cleaner (The question is also if we should store a set of default overwrites for forecasting => similar to default_config for user convenience). However this removes existing functionality and might break peoples workflow, so it should be maybe addressed in a separate issue. But I am open to remove it if people are ok with it :+1:

clessig: We should raise it in today's meeting. I don't think many people use it at the moment but we can and should check.

CC @tjhunter

src/weathergen/utils/cli.py
```diff
@@ -0,0 +1,115 @@
+import argparse
+from pathlib import Path
+
+import pandas as pd
+
+
+def get_train_parser():
```
tjhunter: output type please: `-> argparse.ArgumentParser`

src/weathergen/utils/cli.py
```diff
@@ -0,0 +1,115 @@
...
+    parser = argparse.ArgumentParser(allow_abbrev=False)
+    _add_general_arguments(parser)
+
+    return parser
+
+
+def get_continue_parser():
```
tjhunter: same here

src/weathergen/utils/cli.py
```diff
@@ -0,0 +1,115 @@
...
+        "--config",
+        type=Path,
+        nargs="+",
+        default=[],
+        help="Optional experiment specfic configuration files in ascending order of precedence.",
+    )
+    parser.add_argument(
+        "--run_id",
+        default=False,
+        help="Store training and evaluation artifacts under the given same run_id",
```
tjhunter: Let's make the help even more clear: `The run id for this run. All artifacts (models, metrics, ...) will be stored under this run_id. If not provided, a new run_id will be created.`

src/weathergen/utils/cli.py
```diff
@@ -0,0 +1,115 @@
...
+        nargs="+",
+        default=[],
+        help="Overwrite individual config options. This takes precedence over overwrites passed via --config or --finetune_forecast. Individual items should be of the form: parent_obj.nested_obj=value",
+    )
+
+
+def _add_model_loading_params(parser: argparse.ArgumentParser):
+    parser.add_argument(
+        "-id",
+        "--run_id_base",
```
tjhunter: I think this name could be more explicit about what we are trying to load. This is about starting the new work from a specific piece trained on an older run id. I would call it `--start-from-run-id` or `--from-run-id` to make it very clear. This is used right now from `evaluate` or `run_continue` so I think it captures this intent.

I would make the help more explicit (feel free to rewrite):

```The run id to start from. It is typically a run id corresponding to a model checkpoint.
```

src/weathergen/utils/config.py
```diff
@@ -46,20 +58,14 @@ def save(config: Config, epoch: int | None):
...
-    fname = dirname / f"model_{config.run_id}{epoch_str}.json"
+    fname = dirname / _get_model_config_file_name(config.run_id, epoch)
 
     json_str = json.dumps(OmegaConf.to_container(config))
     with fname.open("w") as f:
         f.write(json_str)
 
 
-def load_model_config(run_id: str, epoch: int | None, model_path: str | None) -> OmegaConf:
+def load_model_config(run_id: str, epoch: int | None, model_path: str | None) -> DictConfig:
```
tjhunter: you can change it to `-> Config` to make it resilient to future change

src/weathergen/utils/config.py
```diff
@@ -83,14 +86,38 @@ def load_model_config(run_id: str, epoch: int | None, model_path: str | None) ->
     return OmegaConf.create(json.loads(json_str))
 
 
+def _get_model_config_file_name(run_id: str, epoch: int | None):
```
tjhunter, line 89: thank you for doing that!

src/weathergen/utils/config.py
```diff
@@ -162,7 +214,7 @@ def _load_private_conf(private_home: Path | None) -> OmegaConf:
     return private_cf
 
 
-def _load_default_conf() -> OmegaConf:
+def _load_default_conf() -> DictConfig:
```
tjhunter: same comment

tests/test_cli.py
```diff
@@ -0,0 +1,121 @@
+import pathlib
```
tjhunter, line 1: thank you very much for the tests, this is great

src/weathergen/utils/cli.py
```diff
@@ -0,0 +1,115 @@
...
+    parser.add_argument(
+        "--finetune_forecast",
+        action="store_true",
+        help="Fine tune for forecasting. It overwrites some of the Config settings. Overwrites specified with --config take precedence.",
+    )
+
+    return parser
+
+
+def get_evaluate_parser():
```
tjhunter: `-> ArgumentParser`

src/weathergen/utils/cli.py
```diff
@@ -0,0 +1,115 @@
...
+    parser = argparse.ArgumentParser(allow_abbrev=False)
+
+    _add_model_loading_params(parser)
+    _add_general_arguments(parser)
+
+    parser.add_argument(
```
tjhunter, line 35: not for now, but since we have a more generic way to inject config values, we may not need these params anymore

tjhunter (changes requested):
@grassesi  thanks again for this refactoring (and apologies for the delay in the review). I have some comments on the names.
Also, could you make sure that you follow up with a change on `launch-slurm.py` in the private repo?

src/weathergen/utils/cli.py
grassesi: I think that should be addressed if we try to move the current datetime representation (`"%Y%m%d%H%M"`) to ISO 8601 strings (if I get around to implement #37 )

src/weathergen/utils/cli.py
grassesi: I also had trouble finding a good name. `from-run-id` is great but I would choose `from_run_id` to remain consistent with the formatting of other cli options

grassesi:
> @grassesi thanks again for this refactoring (and apologies for the delay in the review). I have some comments on the names. Also, could you make sure that you follow up with a change on `launch-slurm.py` in the private repo?

Done that: There is now an associated PR in the private repo

Issue #211, opened 2025-05-05
Problems with training on Alps


kctezcan:
### Is your feature request related to a problem? Please describe.

I run into problems when training on Alps. The exact same code with the exact same sbatch script (except running in container vs direct srun) is running on balfrin.

When I submit multiple jobs, 1 in 5 is working as expected, others are giving errors or hanging:

Same code (on 2 nodes – 8 GPUs, master_port=1345, 30 sec timeout): 5 different runs -> 3 different outcomes:
- Job_id=403404, 403408
  - Training normally
- Job_id=403405
  - ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/usr/local/lib/python3.12/dist-packages/weathergen-0.1.0.dist-info/METADATA'
- Job_id=403406, 403407
  - torch.distributed.DistNetworkError: The client socket has timed out after 30000ms while trying to connect to (172.28.31.224, 57323).
  
Same code (on 4 nodes – 16 GPUs, master_port=57323, 30 sec timeout): 
- job_id=403432
  - torch.distributed.DistStoreError: Timed out after 31 seconds waiting for clients. 12/16 clients joined.
 
Same code (on 4 nodes – 16 GPUs, master_port=57323, 120 sec timeout): 5 different runs -> 3 different outcomes:
- job_id= 403478, 403480, 403481
  - 403478: torch.distributed.DistNetworkError: The client socket has timed out after 120000ms while trying to connect to (172.28.31.156, 57323).
  - 403480: torch.distributed.DistStoreError: Timed out after 121 seconds waiting for clients. 8/16 clients joined.
  - 403481: torch.distributed.DistStoreError: Timed out after 121 seconds waiting for clients. 12/16 clients joined.
- Job_id=403479
  - Training normally
- Job_id=403482
  - Just hangs, no error

Not sure what is causing this problem, either there are two problems: i) connection between the nodes causing timeout errors and ii) access to the file system causing OS Errors. Or there is an underlying problem for both.

### Describe the solution you'd like

Ideally the training should work everytime we submit a job.

clessig:
This is not related to the weathergen code but we can use the issue to track the issue on ALPS.

kctezcan:
There were two problems:

1. pip install was being called N times on each of the N GPUs, which was causing the error: " OSError: [Errno 2] No such file or directory: '/usr/local/lib/python3.12/dist-packages/weathergen-0.1.0.dist-info/METADATA'"
[...]

PR #228, opened 2025-05-09
Ktezcan/dev/train mtm

kctezcan:
This is not a real PR. This is just to quickly share the WIP changes for training on clariden.

See the contents of the new folder: train_mtm

1. Notice that the instructions for making the venv is in the folder.
2. My sbatch scripts are also in the folder, you can see how we set up the environment variables for NCCL. These have my paths, these need to be adjusted.
3. the change "hpy_idxs_ord = np.argsort(hpy_idxs, kind='mergesort')" is because we have numpy<2.0 in the container we use.

Feel free to reject/delete this PR.

Related to issue https://github.com/ecmwf/WeatherGenerator/issues/211. I am available again in the evening for questions. Thanks

### DIFF
--- a/src/weathergen/datasets/tokenizer_utils.py
+++ b/src/weathergen/datasets/tokenizer_utils.py
@@ -98,7 +98,8 @@ def hpy_cell_splits(coords: torch.tensor, hl: int):
     posr3 = s2tor3(thetas, phis)
 
     # extract information to split according to cells by first sorting and then finding split idxs
-    hpy_idxs_ord = np.argsort(hpy_idxs, stable=True)
+    # hpy_idxs_ord = np.argsort(hpy_idxs, stable=True) # KCT
+    hpy_idxs_ord = np.argsort(hpy_idxs, kind='mergesort')
     splits = np.flatnonzero(np.diff(hpy_idxs[hpy_idxs_ord]))
 
     # extract per cell data
--- a/src/weathergen/train/trainer_base.py
+++ b/src/weathergen/train/trainer_base.py
@@ -71,19 +71,23 @@ class Trainer_Base:
             cf.rank = rank
             cf.num_ranks = num_ranks
             return
+        
+        master_port = os.environ.get("MASTER_PORT", "29514")
 
         local_rank = int(os.environ.get("SLURM_LOCALID"))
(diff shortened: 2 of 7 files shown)


src/weathergen/train/trainer.py
```diff
@@ -455,7 +455,7 @@ def compute_loss(
 
         # normalize by all targets and forecast steps that were non-empty
         # (with each having an expected loss of 1 for an uninitalized neural net)
-        loss /= ctr_ftarget
+        loss = loss / ctr_ftarget
```
clessig, line 464: This won't fix the error you are seeing, it just pushes it a few lines further down. The fix is to have no empty targets, which needs to be enforced in the data loader.

clessig:
> This is not a real PR. This is just to quickly share the WIP changes for training on clariden.
> 
> See the contents of the new folder: train_mtm
> 
> 1. Notice that the instructions for making the venv is in the folder.
> 2. My sbatch scripts are also in the folder, you can see how we set up the environment variables for NCCL. These have my paths, these need to be adjusted.
> 3. the change "hpy_idxs_ord = np.argsort(hpy_idxs, kind='mergesort')" is because we have numpy<2.0 in the container we use.
> 
> Feel free to reject/delete this PR.
> 
> Related to issue #211. I am available again in the evening for questions. Thanks

Could we move the instructions for Clariden (including how to build the container) to the private repo. That's the right place to track it.

src/weathergen/model/model.py
```diff
@@ -558,18 +558,19 @@ def predict(self, model_params, fstep, tokens, streams_data, target_coords_idxs)
...
-                        tc_embed,
-                        streams_data[i_b][ii].target_coords[fstep],
-                        use_reentrant=False,
-                    )
-                    if len(streams_data[i_b][ii].target_coords[fstep].shape) > 1
-                    else streams_data[i_b][ii].target_coords[fstep]
-                    for i_b in range(len(streams_data))
-                ]
-            )
+            with torch.amp.autocast('cuda', enabled=False):
```
clessig, line 561: I think it would be better to use:

torch.cuda.amp.autocast(enabled=True, dtype=torch.float32, device='cuda') 

This way the dtype to be used is explicit.

Issue #210, opened 2025-05-05
Metrics reading code does not handle columns with some NaNs


tjhunter:
The example file has mixtures of floats and nan's for `stream.meteosatxseviri.loss_mse.loss_avg` 
This should be handled properly. 
- not present -> null
- "nan" -> NaN
- number -> number

Given that json has no encoding for NaN, maybe it is worth encoding all numbers as strings.

clessig:
This leads to the correct behaviour:

```
    df = pl.read_ndjson(f"./results/{run_id}/metrics.json")
    dict = {k : pl.Float64 for k in df.schema.keys() if 'loss_mse.loss_avg' in k}
    dict["weathergen.timestamp"] = pl.Int64
    df = pl.read_ndjson(f"./results/{run_id}/metrics.json", schema_overrides=dict)
```

PR #238, opened 2025-05-12
Langguth issue0232 hard coded relative paths in plot training and trainlogger

### LINKED ISSUE #201: plot_training does not use a config file

The runs to be plotted a currently directly specified in plot_training.py. This is inconvenient.

The runs to be plotted (run_id, job id, description) should be read from a yaml file.

_No response_

### Additional context

_No response_

### Organisation

_No response_

### LINKED ISSUE #232: Hard-coded relative paths in plot_training and TrainLogger
### What happened?

The code to plot the loss-curves from training runs (`plot_training.py`) involves several hard-coded relative paths. 
This is error-prone and it does in particular not work if training runs are evaluated that are not located in the personal `models`- and `results`-directory of the repo, e.g. in a shared directory when collaborating with partners.

Run `plot_training.py` on a training run that is not located under the `models`-directory of the personal repo.

### Version

develop

OS-independant

### Relevant log output

```shell
File "/p/project1/hclimrep/langguth1/WeatherGenerator_ownfork/src/weathergen/utils/train_logger.py", line 177, in read
    log_train_df = read_metrics(cf, run_id, "train", cols1)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/p/project1/hclimrep/langguth1/WeatherGenerator_ownfork/src/weathergen/utils/train_logger.py", line 260, in read_metrics
    df = pl.read_ndjson(f"./results/{run_id}/metrics.json")
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/p/project1/hclimrep/langguth1/WeatherGenerator_ownfork/.venv/lib/python3.12/site-packages/polars/io/ndjson.py", line 196, in read_ndjson
    ).collect()
      ^^^^^^^^^
  File "/p/project1/hclimrep/langguth1/WeatherGenerator_ownfork/.venv/lib/python3.12/site-packages/polars/_utils/deprecation.py", line 88, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/p/project1/hclimrep/langguth1/WeatherGenerator_ownfork/.venv/lib/python3.12/site-packages/polars/lazyframe/frame.py", line 2188, in collect
    return wrap_df(ldf.collect(engine, callback))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: No such file or directory (os error 2): ./results/qlz6n9eg/metrics.json
```

### Accompanying data

_No response_

### Organisation

JSC

mlangguth89:
## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #232 
Closes #201 

## Code Compatibility

-   [x] I ran `plot_training.py` and it works now consistently  

### Dependencies

### Documentation

## Additional Notes

`plot_training.py` now runs from arbitrary directories since hard-coded relative paths are avoided and parsing is allowed instead. The default however makes use of the previous hard-coded relative paths for backward compatabilty.

### DIFF
--- a/src/weathergen/utils/plot_training.py
+++ b/src/weathergen/utils/plot_training.py
@@ -9,17 +9,18 @@
 
 import argparse
 import logging
-import yaml
 import subprocess
 from pathlib import Path
 
 import matplotlib.pyplot as plt
 import numpy as np
+import yaml
 
 import weathergen.utils.config as config
 from weathergen.utils.train_logger import Metrics, TrainLogger
 
 ####################################################################################################
+
 
 def yaml_or_path(str_or_path: str) -> dict:
     """
@@ -34,21 +35,23 @@ def yaml_or_path(str_or_path: str) -> dict:
     Returns
     dict
         The parsed YAML content as a dictionary.
-    """ 
+    """
     # Load from file if it's a valid path
     if Path(str_or_path).is_file():
-        with open(str_or_path, 'r') as f:
+        with open(str_or_path) as f:
             data_dict = yaml.safe_load(f)
     else:
         try:
             data_dict = yaml.safe_load(str_or_path)
         except yaml.YAMLError as e:
-            raise argparse.ArgumentTypeError(f"Invalid YAML string or path: {e}")
+            raise argparse.ArgumentTypeError(f"Invalid YAML string or path: {e}") from e
 
     # Validate the structure: {run_id: [job_id, experiment_name]}
     if not isinstance(data_dict, dict):
-        raise argparse.ArgumentTypeError("Input must be a dictionary mapping run_id to [job_id, experiment_name].")
-    
+        raise argparse.ArgumentTypeError(
+            "Input must be a dictionary mapping run_id to [job_id, experiment_name]."
+        )
+
     for k, v in data_dict.items():
         if not (isinstance(v, list) and len(v) == 2):
             raise argparse.ArgumentTypeError(
@@ -56,6 +59,7 @@ def yaml_or_path(str_or_path: str) -> dict:
             )
 
     return data_dict
+
 
 ####################################################################################################
 def clean_plot_folder(plot_dir: Path = "./plots/"):
--- a/src/weathergen/utils/plot_training.py
+++ b/src/weathergen/utils/plot_training.py
@@ -24,7 +24,7 @@ _logger = logging.getLogger(__name__)
 ####################################################################################################
 
 
-def check_run_id_dict(run_id_dict: dict) -> bool:
+def _check_run_id_dict(run_id_dict: dict) -> bool:
     """
     Check if the run_id_dict is valid.
 
@@ -45,51 +45,33 @@ def check_run_id_dict(run_id_dict: dict) -> bool:
             )
 
 
-def read_yaml_file(file_path: Path) -> dict:
+def _read_yaml_config(yaml_path_or_str: str | Path) -> dict:
     """
-    Read a YAML file and return its content as a dictionary.
+    Read a from a YAML file or a dictionary-like string to get a configuration dictionary.
 
     Parameters
     ----------
-    file_path : Path
-        Path to the YAML file.
+    yaml_path_or_str : string or Path
+        Path to the YAML file or dictonary string to read.
     Returns
     -------
     dict
         The content of the YAML file as a dictionary.
     """
-    with open(file_path) as f:
-        data_dict = yaml.safe_load(f)
+    if Path(yaml_path_or_str).is_file():
+        with open(yaml_path_or_str) as f:
+            data_dict = yaml.safe_load(f)
+    else:
+        data_dict = yaml.safe_load(yaml_path_or_str)
 
     # Validate the structure: {run_id: [job_id, experiment_name]}
-    check_run_id_dict(data_dict)
-
-    return data_dict
-
-
-def read_yaml_string(yaml_string: str) -> dict:
-    """
-    Read a YAML string and return its content as a dictionary.
-
-    Parameters
-    ----------
-    yaml_string : str
-        YAML string to read.
-    Returns
-    -------
-    dict
-        The content of the YAML string as a dictionary.
-    """
-    data_dict = yaml.safe_load(yaml_string)
-
-    # Validate the structure: {run_id: [job_id, experiment_name]}
-    check_run_id_dict(data_dict)
+    _check_run_id_dict(data_dict)
 
     return data_dict
 
 
 ####################################################################################################
-def clean_plot_folder(plot_dir: Path = "./plots/"):
+def clean_plot_folder(plot_dir: Path):
     """
     Clean the plot folder by removing all png-files in it.
 
@@ -129,8 +111,8 @@ def plot_lr(
     runs_ids: dict[str, list],
     runs_data: list[Metrics],
     runs_active: list[bool],
+    plot_dir: Path,
     x_axis: str = "samples",
-    plot_dir: Path = Path("./plots"),
 ):
     """
     Plot learning rate curves of training runs.
@@ -143,10 +125,10 @@ def plot_lr(
         list of Metrics objects containing the training data
     runs_active : list
         list of booleans indicating whether the run is still active
-    x_axis : str
-        x-axis strings used in the column names (options: "samples", "dtime")
     plot_dir : Path
         directory to save the plots
+    x_axis : str
+        x-axis strings used in the column names (options: "samples", "dtime")
     """
     prop_cycle = plt.rcParams["axes.prop_cycle"]
     colors = prop_cycle.by_key()["color"] + ["r", "g", "b", "k", "y", "m"]
@@ -199,8 +181,8 @@ def plot_utilization(
     runs_ids: dict[str, list],
     runs_data: list[Metrics],
     runs_active: list[bool],
+    plot_dir: Path,
     x_axis: str = "samples",
-    plot_dir: Path = Path("./plots"),
 ):
     """
     Plot compute utilization of training runs.
@@ -213,10 +195,10 @@ def plot_utilization(
         list of Metrics objects containing the training data
     runs_active : list
         list of booleans indicating whether the run is still active
-    x_axis : str
-        x-axis strings used in the column names (options: "samples", "dtime")
     plot_dir : Path
         directory to save the plots
+    x_axis : str
+        x-axis strings used in the column names (options: "samples", "dtime")
     """
     prop_cycle = plt.rcParams["axes.prop_cycle"]
     colors = prop_cycle.by_key()["color"] + ["r", "g", "b", "k", "y", "m"]
@@ -276,11 +258,11 @@ def plot_loss_per_stream(
     runs_data: list[Metrics],
     runs_active: list[bool],
     stream_names: list[str],
+    plot_dir: Path,
     errs: list[str] = ["loss_mse"],
     x_axis: str = "samples",
     x_type: str = "step",
     x_scale_log: bool = False,
-    plot_dir: Path = Path("./plots"),
 ):
     """
     Plot each stream in stream_names (using matching to data columns) for all run_ids
@@ -297,6 +279,8 @@ def plot_loss_per_stream(
         list of booleans indicating whether the run is still active
     stream_names : list
         list of stream names to plot
+    plot_dir : Path
+        directory to save the plots
     errs : list
         list of errors to plot (e.g. mse, stddev)
     x_axis : str
@@ -305,8 +289,6 @@ def plot_loss_per_stream(
         x-axis type (options: "step", "reltime")
     x_scale_log : bool
         whether to use log scale for x-axis
-    plot_dir : Path
-        directory to save the plots
     """
 
     modes = [modes] if type(modes) is not list else modes
@@ -404,10 +386,10 @@ def plot_loss_per_run(
     run_desc: str,
     run_data: Metrics,
     stream_names: list[str],
+    plot_dir: Path,
     errs: list[str] = ["mse"],
     x_axis: str = "samples",
     x_scale_log: bool = False,
-    plot_dir: Path = Path("./plots"),
 ):
     """
     Plot all stream_names (using matching to data columns) for given run_id
@@ -424,14 +406,14 @@ def plot_loss_per_run(
         Metrics object containing the training data
     stream_names : list
         list of stream names to plot
+    plot_dir : Path
+        directory to save the plots
     errs : list
         list of errors to plot (e.g. mse, stddev)
     x_axis : str
         x-axis strings used in the column names (options: "samples", "dtime")
     x_scale_log : bool
         whether to use log scale for x-axis
-    plot_dir : Path
-        directory to save the plots
     """
     plot_dir = Path(plot_dir)
 
@@ -518,13 +500,17 @@ if __name__ == "__main__":
     parser = argparse.ArgumentParser()
 
     parser.add_argument(
-        "-o", "--output_dir", default="./plots/", type=str, help="Directory where plots are saved"
+        "-o",
+        "--output_dir",
+        default="./plots/",
+        type=Path,
+        help="Directory where plots are saved"
     )
     parser.add_argument(
         "-m",
         "--model_base_dir",
         default="./models/",
-        type=str,
+        type=Path,
         help="Base-directory where models are saved",
     )
     parser.add_argument(
@@ -557,7 +543,7 @@ if __name__ == "__main__":
     run_id_group.add_argument(
         "-rs",
         "--run_ids_str",
-        type=read_yaml_string,
+        type=_read_yaml_config,
         dest="rs",
         help="Dictionary-string of form '{run_id: [job_id, experiment_name]}' for training runs to plot",
     )
@@ -566,7 +552,7 @@ if __name__ == "__main__":
         "-rf",
         "--run_ids_file",
         dest="rf",
-        type=read_yaml_file,
+        type=_read_yaml_config,
         help="YAML file configuring the training run IDS to plot",
     )
 
@@ -602,7 +588,7 @@ if __name__ == "__main__":
     plot_lr(runs_ids, runs_data, runs_active, plot_dir=out_dir)
 
     # plot performance
-    plot_utilization(runs_ids, runs_data, runs_active, plot_dir=out_dir)
+    #plot_utilization(runs_ids, runs_data, runs_active, plot_dir=out_dir)
 
     # compare different runs
     plot_loss_per_stream(
--- a/src/weathergen/utils/plot_training.py
+++ b/src/weathergen/utils/plot_training.py
@@ -19,47 +19,71 @@ import yaml
 import weathergen.utils.config as config
 from weathergen.utils.train_logger import Metrics, TrainLogger
 
+_logger = logging.getLogger(__name__)
+
 ####################################################################################################
 
-
-def yaml_or_path(str_or_path: str) -> dict:
+def check_run_id_dict(run_id_dict: dict) -> bool:
     """
-    Custom argparse type to handle both YAML strings and file paths.
-    If the input is a valid file path, it loads the YAML content from the file.
-    If the input is a valid YAML string, it loads the content directly.
+    Check if the run_id_dict is valid.
 
     Parameters
     ----------
-    value : str
-        The input string to be parsed as YAML or file path.
+    run_id_dict : dict
+        Dictionary to check.
     Returns
-    dict
-        The parsed YAML content as a dictionary.
+    -------
     """
-    # Load from file if it's a valid path
-    if Path(str_or_path).is_file():
-        with open(str_or_path) as f:
-            data_dict = yaml.safe_load(f)
-    else:
-        try:
-            data_dict = yaml.safe_load(str_or_path)
-        except yaml.YAMLError as e:
-            raise argparse.ArgumentTypeError(f"Invalid YAML string or path: {e}") from e
+    if not isinstance(run_id_dict, dict):
+        return False
 
-    # Validate the structure: {run_id: [job_id, experiment_name]}
-    if not isinstance(data_dict, dict):
-        raise argparse.ArgumentTypeError(
-            "Input must be a dictionary mapping run_id to [job_id, experiment_name]."
-        )
-
-    for k, v in data_dict.items():
-        if not (isinstance(v, list) and len(v) == 2):
+    for k, v in run_id_dict.items():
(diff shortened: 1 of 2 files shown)


src/weathergen/utils/plot_training.py
```diff
@@ -298,40 +409,70 @@ def plot_loss_per_run(
...
+        help="Base-directory where models are saved",
+    )
+    parser.add_argument(
+        "-d",
+        "--delete",
+        default=False,
+        action="store_true",
+        help="Delete all plots in the output directory before plotting",
+    )
+    parser.add_argument(
```
clessig, line 591: This needs to be specified via a yaml that is passed, not directly on the cmd line. There's an open issue on this.

mlangguth89: I was not aware of #201. Depending on the decision made there, the code should be adapted accordingly.

clessig: There needs to be an option for yaml files. I think passing this on the command line is not good, since it's not reproduible, but if people have strong feelings about it then fine.

mlangguth89: What about a parsing argument that can be both, a path to a yml-file or a dictionary-like string (current approach)?

src/weathergen/utils/plot_training.py
```diff
@@ -298,40 +450,70 @@ def plot_loss_per_run(
...
+        "-ids",
+        "--runs_ids",
+        type=yaml_or_path,
+        required=True,
+        help="YAML config file or dictionary-string of form {run_id: [job_id, experiment_name]}",
```
clessig: Let's have two separate command line arguments, one for yaml and one for list of run_ids.

src/weathergen/utils/plot_training.py
```diff
@@ -298,40 +450,70 @@ def plot_loss_per_run(
...
+    runs_data = [TrainLogger.read(run_id, model_path=model_base_dir) for run_id in runs_ids]
 
     # determine which runs are still alive (as a process, though they might hang internally)
     ret = subprocess.run(["squeue"], capture_output=True)
     lines = str(ret.stdout).split("\\n")
     runs_active = [np.array([str(v[0]) in l for l in lines[1:]]).any() for v in runs_ids.values()]
 
     x_scale_log = False
-    x_type = ("reltime",)  #'step'
+    x_type = ("step",)  #'reltime'
```
clessig: reltime is not working at the moment. Can we open a follow up issue.

src/weathergen/utils/plot_training.py
```diff
@@ -342,6 +524,7 @@ def plot_loss_per_run(
         ["era5", "METEOSAT", "NPP"],
```
clessig: This needs to be "ERA5" again when #210 is merged. This list should probably also be a command line argument

src/weathergen/utils/train_logger.py
```diff
@@ -203,7 +204,7 @@ def read(run_id, epoch=-1):
         except:
             print(f"Warning: no validation data loaded for run_id={run_id}")
             log_val = np.array([])
-        metrics_val_df = read_metrics(cf, run_id, "val", cols2)
+        metrics_val_df = read_metrics(cf, run_id, "val", cols2, results_path=result_dir_base)
```
clessig: This needs to rebased after #210 is merged.

src/weathergen/utils/plot_training.py
mlangguth89: I will implement this asap.

src/weathergen/utils/plot_training.py
mlangguth89: @clessig Yes, that's why I changed the default. It should probably also be a parsing argument, shouldn't it?

src/weathergen/utils/plot_training.py
mlangguth89: :+1:

mlangguth89: I've added `--streams` as a parsing argument. Furthermore, I made any access to the streams case-insensitive by using `.lower()` to the respective strings. This makes it quite fail-safe against changes to the naming convention of streams in the metrics-file.

src/weathergen/utils/train_logger.py
mlangguth89: What do you mean here?

clessig: Tim implemented some things in #210 that need adaptation here, I think.

src/weathergen/utils/plot_training.py
clessig: The streams name are (and should be) preserved now in the metrics file. Hence it should also be "ERA5" here

mlangguth89: Yep, it didn't work anyway in the end. The latest changes to the PR keep the streams names untouched.

src/weathergen/utils/train_logger.py
mlangguth89: Ah yes. That's also done with the latest changes.

mlangguth89:
@clessig 
Can you check the latest changes to the PR? Thanks :-)

src/weathergen/utils/plot_training.py
```diff
@@ -9,33 +9,145 @@
...
 import weathergen.utils.config as config
 from weathergen.utils.train_logger import Metrics, TrainLogger
 
-out_folder = pathlib.Path("./plots/")
+_logger = logging.getLogger(__name__)
+
+####################################################################################################
+
+
+def check_run_id_dict(run_id_dict: dict) -> bool:
```
tjhunter: for the functions inside a module that are not meant to be exposed to everyone, the python convention is to start with "_":  `_check_run_id_dict`

src/weathergen/utils/plot_training.py
```diff
@@ -9,33 +9,145 @@
...
+        return False
+
+    for k, v in run_id_dict.items():
+        if not isinstance(k, str) or not isinstance(v, list) or len(v) != 2:
+            raise argparse.ArgumentTypeError(
+                f"Each key must be a string and each value must be a list of [job_id, experiment_name], but got: {k}: {v}"
+            )
+
+
+def read_yaml_file(file_path: Path) -> dict:
```
tjhunter: same

src/weathergen/utils/plot_training.py
```diff
@@ -9,33 +9,145 @@
...
+    with open(file_path) as f:
+        data_dict = yaml.safe_load(f)
+
+    # Validate the structure: {run_id: [job_id, experiment_name]}
+    check_run_id_dict(data_dict)
+
+    return data_dict
+
+
+def read_yaml_string(yaml_string: str) -> dict:
```
tjhunter: you do not need to creae a separate function given that the safe_load function accepts both. I would have 1 function:

```
def read_yaml(c: str | Path)
```

src/weathergen/utils/plot_training.py
```diff
@@ -9,33 +9,145 @@
...
+    # Validate the structure: {run_id: [job_id, experiment_name]}
+    check_run_id_dict(data_dict)
+
+    return data_dict
 
 
 ####################################################################################################
-def clean_out_folder():
-    for image in out_folder.glob("*.png"):
+def clean_plot_folder(plot_dir: Path = "./plots/"):
```
tjhunter: thanks for adding the parametrization. There are 2 things we should change:
- we should not use hardcoded path values: when they are scattered around the code, they are hard to track and to change in the future
- let's limit the amount of default parameters to the bare minimum. they should nearly never be used for private functions in a python module. they usually bring confusion when trying to understand how code works and they limit the help from vscode (harder to track if a parameter is missing

In this case, I would just have:  `def _clean_plot_folder(plot_dir: Path):`

src/weathergen/utils/plot_training.py
```diff
@@ -9,33 +9,145 @@
...
+    plot_dir : Path
+        Path to the plot directory
+    """
+    for image in plot_dir.glob("*.png"):
         image.unlink()
 
 
 ####################################################################################################
-def get_stream_names(run_id):
+def get_stream_names(run_id: str, model_path: Path | None = "./model"):
```
tjhunter, line 148: same comment

src/weathergen/utils/plot_training.py
```diff
@@ -9,33 +9,145 @@
...
 
 
 ####################################################################################################
-def plot_lr(runs_ids, runs_data: list[Metrics], runs_active, x_axis="samples"):
+def plot_lr(
+    runs_ids: dict[str, list],
+    runs_data: list[Metrics],
+    runs_active: list[bool],
+    x_axis: str = "samples",
+    plot_dir: Path = Path("./plots"),
```
tjhunter: same comment: `plot_dir: Path,`

. `x_axis` also has a default but it was already there, let's not touch it in this PR.

src/weathergen/utils/plot_training.py
```diff
@@ -61,6 +173,9 @@ def plot_lr(runs_ids, runs_data: list[Metrics], runs_active, x_axis="samples"):
         ]
 
     if len(legend_str) < 1:
+        _logger.warning(
```
tjhunter, line 218: thanks!

src/weathergen/utils/plot_training.py
```diff
@@ -71,12 +186,38 @@ def plot_lr(runs_ids, runs_data: list[Metrics], runs_active, x_axis="samples"):
     plt.xlabel(x_axis)
     plt.tight_layout()
     rstr = "".join([f"{r}_" for r in runs_ids])
-    plt.savefig(f"./plots/{rstr}lr.png")
+
+    # save the plot
+    plt_fname = plot_dir / f"{rstr}lr.png"
+    _logger.info(f"Saving learning rate plot to '{plt_fname}'")
```
tjhunter, line 234: great that you put all the info logging

src/weathergen/utils/plot_training.py
```diff
@@ -71,12 +186,38 @@ def plot_lr(runs_ids, runs_data: list[Metrics], runs_active, x_axis="samples"):
...
 
 
 ####################################################################################################
-def plot_utilization(runs_ids, runs_data: list[Metrics], runs_active, x_axis="samples"):
+def plot_utilization(
+    runs_ids: dict[str, list],
+    runs_data: list[Metrics],
+    runs_active: list[bool],
+    x_axis: str = "samples",
+    plot_dir: Path = Path("./plots"),
```
tjhunter: same comment

src/weathergen/utils/plot_training.py
```diff
@@ -119,24 +261,52 @@ def plot_utilization(runs_ids, runs_data: list[Metrics], runs_active, x_axis="sa
...
+    _logger.info(f"Saving utilization plot to '{plt_fname}'")
+    plt.savefig(plt_fname)
     plt.close()
 
 
 ####################################################################################################
 def plot_loss_per_stream(
-    modes,
-    runs_ids,
+    modes: list[str],
```
tjhunter, line 316: thanks!

src/weathergen/utils/plot_training.py
```diff
@@ -119,24 +261,52 @@ def plot_utilization(runs_ids, runs_data: list[Metrics], runs_active, x_axis="sa
...
-    x_axis="samples",
-    x_type="step",
-    x_scale_log=False,
+    runs_active: list[bool],
+    stream_names: list[str],
+    errs: list[str] = ["loss_mse"],
+    x_axis: str = "samples",
+    x_type: str = "step",
+    x_scale_log: bool = False,
+    plot_dir: Path = Path("./plots"),
```
tjhunter: same comment. you will need to move it before the args with default arguments, this is fine

src/weathergen/utils/plot_training.py
```diff
@@ -215,28 +387,53 @@ def plot_loss_per_stream(
...
     run_data: Metrics,
-    stream_names,
-    errs=["mse"],
-    x_axis="samples",
-    x_scale_log=False,
+    stream_names: list[str],
+    errs: list[str] = ["mse"],
+    x_axis: str = "samples",
+    x_scale_log: bool = False,
+    plot_dir: Path = Path("./plots"),
```
tjhunter: same comment

src/weathergen/utils/plot_training.py
```diff
@@ -298,73 +498,159 @@ def plot_loss_per_run(
     sstr = "".join(
         [f"{r}_".replace(",", "").replace("/", "_").replace(" ", "_") for r in legend_str]
     )
-    plt.savefig(out_folder / "{}_{}{}.png".format(run_id, "".join([f"{m}_" for m in modes]), sstr))
+
+    # save the plot
+    plt_fname = plot_dir / "{}_{}{}.png".format(run_id, "".join([f"{m}_" for m in modes]), sstr)
+    _logger.info(f"Saving loss plot for {run_id}-run to '{plt_fname}'")
+    plt.savefig(plot_dir / "{}_{}{}.png".format(run_id, "".join([f"{m}_" for m in modes]), sstr))
```
tjhunter: `plt.savefig(plt_fname)` ?

src/weathergen/utils/plot_training.py
```diff
@@ -298,73 +498,159 @@ def plot_loss_per_run(
...
+    # python plot_training.py -rf eval_run.yml -m ./trained_models -o ./training_plots
+    # When providing a string for configuring the run IDs:
+    # python plot_training.py -rs "{run_id: [job_id, experiment_name]}" -m ./trained_models -o ./training_plots
+
     logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s")
     parser = argparse.ArgumentParser()
-    parser.add_argument("-d", "--delete")
+
+    parser.add_argument(
+        "-o", "--output_dir", default="./plots/", type=str, help="Directory where plots are saved"
```
tjhunter: I prefer we do not have short-hand arguments such as `-o`:
- it is confusing to have 2 ways to do the same thing
- the longer argument is more explicit
The programs such as `cp`, `mv` etc have shorthands because they are used my millions of people every day. In our case, I believe explicit is better than slightly faster at the risk of confusion.

src/weathergen/utils/plot_training.py
```diff
@@ -298,73 +498,159 @@ def plot_loss_per_run(
...
+    )
+    parser.add_argument(
+        "-m",
```
tjhunter, line 524: same

src/weathergen/utils/plot_training.py
```diff
@@ -298,73 +498,159 @@ def plot_loss_per_run(
...
+        "--delete",
+        default=False,
+        action="store_true",
+        help="Delete all plots in the output directory before plotting",
+    )
+    parser.add_argument(
+        "--streams",
+        "-s",
+        dest="streams",
+        default=["ERA5", "METEOSAT", "NPP"],
```
tjhunter: should it be only ERA5 for now? I do no think the other streams are enabled by default

src/weathergen/utils/plot_training.py
```diff
@@ -298,73 +498,159 @@ def plot_loss_per_run(
...
+        "--x_type",
+        "-x",
+        dest="x_type",
+        default="step",
+        type=str,
+        choices=["step", "reltime"],
+        help="Type of x-axis used in plots. Options: 'step' or 'reltime'",
+    )
+
+    run_id_group = parser.add_mutually_exclusive_group(required=True)
```
tjhunter, line 610: I was not aware of this feature, this is great!

src/weathergen/utils/plot_training.py
```diff
@@ -298,73 +498,159 @@ def plot_loss_per_run(
...
+    run_id_group.add_argument(
+        "-rs",
+        "--run_ids_str",
```
tjhunter: I would call it run-id-dict, it is always going to be a string anyway for the shell

src/weathergen/utils/plot_training.py
```diff
@@ -298,73 +498,159 @@ def plot_loss_per_run(
...
+        dest="rs",
+        help="Dictionary-string of form '{run_id: [job_id, experiment_name]}' for training runs to plot",
+    )
+
+    run_id_group.add_argument(
+        "-rf",
+        "--run_ids_file",
+        dest="rf",
+        type=read_yaml_file,
+        help="YAML file configuring the training run IDS to plot",
```
tjhunter, line 624: could you provide in the main doc of the program (in the parser description option) a small example of what this yaml file looks like?

```py
    parser = argparse.ArgumentParser(
        description="""Plot training metrics from the training log files. 
        
        An example YAML file is (chatgpt nonsense here right now)
        
       train:
           plot:
              - run_id: fb89k61l
                run_desc: ERA5 test
                log_file: /path/to/log_file
                log_dir: /path/to/log_dir
                log_cols:
                  - train_loss
                  - val_loss
                  - learning_rate
                  - system_utilization
        """
    )

```

src/weathergen/utils/plot_training.py
```diff
@@ -298,73 +498,159 @@ def plot_loss_per_run(
...
+    )
+
+    # parse the command line arguments
     args = parser.parse_args()
 
+    model_base_dir = Path(args.model_base_dir)
+    out_dir = Path(args.output_dir)
+    streams = list(args.streams)
+    x_type = args.x_type
+    if x_type == "reltime":
```
tjhunter: pro tip: always use whitelist rather than blacklist:

```
    x_types = ["reltime"]
    if args.x_type not in x_types:
        raise ValueError(f"x_type must be one of {x_types}, but got {args.x_type}")

```

src/weathergen/utils/plot_training.py
```diff
@@ -298,73 +498,159 @@ def plot_loss_per_run(
...
     lines = str(ret.stdout).split("\\n")
     runs_active = [np.array([str(v[0]) in l for l in lines[1:]]).any() for v in runs_ids.values()]
 
     x_scale_log = False
-    x_type = ("reltime",)  #'step'
-    # x_type = "step"
 
     # plot learning rate
-    plot_lr(runs_ids, runs_data, runs_active)
+    plot_lr(runs_ids, runs_data, runs_active, plot_dir=out_dir)
```
tjhunter, line 654: you see here that you never need the default arguments of the function since you use the command line values

src/weathergen/utils/train_logger.py
```diff
@@ -133,15 +133,16 @@ def add_val(self, samples, loss_avg, stddev_avg) -> None:
 
     #######################################
     @staticmethod
-    def read(run_id, epoch=-1):
+    def read(run_id, epoch=-1, model_path: str | None = None):
```
tjhunter: same no default argument

src/weathergen/utils/train_logger.py
```diff
@@ -202,9 +203,9 @@ def read(run_id, epoch=-1):
                 log_val = np.loadtxt(f, delimiter=",")
             log_val = log_val.reshape((log_val.shape[0] // len(cols_val), len(cols_val)))
         except:
-            print(f"Warning: no validation data loaded for run_id={run_id}")
+            _logger.warning(f"Warning: no validation data loaded for run_id={run_id}")
```
tjhunter: thanks!

src/weathergen/utils/train_logger.py
```diff
@@ -243,7 +248,11 @@ def by_mode(self, s: str) -> pl.DataFrame:
 
 
 def read_metrics(
-    cf: config.Config, run_id: RunId | None, stage: Stage | None, cols: list[str] | None
+    cf: config.Config,
+    run_id: RunId | None,
+    stage: Stage | None,
+    cols: list[str] | None,
+    results_path: Path = "./results/",
```
tjhunter: same

src/weathergen/utils/train_logger.py
```diff
@@ -258,7 +267,7 @@ def read_metrics(
         run_id = cf.run_id
 
     # TODO: this should be a config option
-    df = read_metrics_file(f"./results/{run_id}/metrics.json")
+    df = read_metrics_file(results_path / run_id / "metrics.json")
```
tjhunter, line 270: thanks for the cleanup!

tjhunter (changes requested):
@mlangguth89  my apologies for the delay. thank you for the cleanup changes here. I am happy to sync if you have any questions

src/weathergen/utils/plot_training.py
```diff
@@ -9,33 +9,127 @@
...
+    Parameters
+    ----------
+    yaml_path_or_str : string or Path
+        Path to the YAML file or dictonary string to read.
+    Returns
+    -------
+    dict
+        The content of the YAML file as a dictionary.
+    """
+    if Path(yaml_path_or_str).is_file():
```
tjhunter: I would change it to 

```
if isinstance(yaml_path_or_str, Path):
   # It's a path:
   assert yaml_path_or_str.is_file(), f"{yaml_path_or_str} is not a file"
  with ...
else:
  # It's a string to parse:
  ... safe_load ...
```
rather than checking if the string could be a file. Not a big issue though, you decide.

src/weathergen/utils/plot_training.py
tjhunter: noting this comment is still here

src/weathergen/utils/plot_training.py
tjhunter: still this comment

src/weathergen/utils/plot_training.py
tjhunter: Regarding your comment that this is training, not evaluation. Indeed. We should keep the plotting under `train` key in the yaml to ensure it will not conflict in the future with other plotting needs (finetuning, evaluation, ...)

src/weathergen/utils/plot_training.py
tjhunter: noting this comment is still here

src/weathergen/utils/plot_training.py
mlangguth89: I made adaptions to the code and added the info to the main doc of the program.
Thanks for assisting!

src/weathergen/utils/plot_training.py
mlangguth89: _read_yaml_config is a function that is directly applied to parsed argument from the command line. At this stage, it's just a string in any case, irrespectively if it points to a YAML-file or provides a dictionary string for configuration. Thus, checking if it's an instance of Path would not work.

tjhunter (approved):
@mlangguth89  it looks great! Thank you for the changes.

PR #248, opened 2025-05-13
[210] Handling nan's and other corner cases in the metrics file

### LINKED ISSUE #210: Metrics reading code does not handle columns with some NaNs

tjhunter:
## Description

Ensures that NaN's are clearly loaded. Currently, they are simply discarded as None. 

## Type of Change

## Issue Number

Closes #210 

## Code Compatibility

### DIFF
--- a/config/streams/streams_test/era5.yml
+++ b/config/streams/streams_test/era5.yml
@@ -21,11 +21,11 @@ ERA5 :
     net : transformer
     num_tokens : 1
     num_heads : 4
-    dim_embed : 128
+    dim_embed : 16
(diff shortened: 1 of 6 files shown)


clessig (approved):
Looks good to me. Should I test it (with the cases where I had problems before) or did you do it?

PR #252, opened 2025-05-16
[219] Ensures the run_id passed through the command line are propagated

tjhunter:
## Description

When using `--run_id=xxx` on train and evaluate, this changes ensure that this run_id is used (and that no other run_id are randomly generated)

## Type of Change

-   [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update

## Issue Number
Related to 219

### DIFF
--- a/src/weathergen/train/trainer.py
+++ b/src/weathergen/train/trainer.py
@@ -282,7 +282,7 @@ class Trainer(Trainer_Base):
         if cf.forecast_policy is not None:
             torch._dynamo.config.optimize_ddp = False
 
-        if self.cf.rank == 0:
+        if is_root():
             config.save(self.cf, None)
             config.print_cf(self.cf)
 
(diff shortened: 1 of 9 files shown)


src/weathergen/train/trainer.py
```diff
@@ -161,6 +161,7 @@ def run(self, cf, run_id_contd=None, epoch_contd=None, run_id_new=False):
         self.data_loader_validation = torch.utils.data.DataLoader(
             self.dataset_val, **loader_params, sampler=None
         )
+        _logger.info(f"Loaded {len(self.dataset)} training samples.")
```
clessig: The "Loaded" is misleading here since nothing is in general loaded at this point. Maybe "Starting training with ... samples" ?

src/weathergen/utils/logger.py
```diff
@@ -33,7 +33,10 @@ def init_loggers():
 
     This function can be called only once, so that it can be called repeatedly in multiprocessing pipelines.
```
clessig: I think this comment confusing.

clessig (approved):
Looks good, just some minor things.

tjhunter:
@clessig thanks for the review! I addressed your comments and tested on leonardo + atos + integration tests. I will merge.

Issue #256, opened 2025-05-20
Using `--finetune_forecast` doesn't disable masking


kacpnowak:
### What happened?

 If in pretraining masking mode was used, using `--finetune_forecast` trains model for forecasting with masking still enabled. 

### What are the steps to reproduce the bug?

Train model in masking mode and then continue training using `--finetune_forecast` argument. Example output: 

![Image](https://github.com/user-attachments/assets/9dc14aaf-5c6a-4610-a5ec-89e079d886c8)

 

### Version

0.1

### Platform (OS and architecture)

Levante

### Relevant log output

```shell

```

### Organisation

AWI

PR #257, opened 2025-05-20
Kacpnowak/develop/fix finetune forecast

kacpnowak:
## Description

This PR fixes small bug within FESOM dataset as well as bug reported here: https://github.com/ecmwf/WeatherGenerator/issues/256

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/256

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/__init__.py
+++ b/src/weathergen/__init__.py
@@ -184,6 +184,7 @@ def train_continue() -> None:
 
     #########################
     if args.finetune_forecast:
+        cf.training_mode = "forecast"
         cf.forecast_delta_hrs = 0  # 12
         cf.forecast_steps = 1  # [j for j in range(1,9) for i in range(4)]
         cf.forecast_policy = "fixed"  # 'sequential_random' # 'fixed' #'sequential' #_random'
--- a/src/weathergen/datasets/fesom_dataset.py
+++ b/src/weathergen/datasets/fesom_dataset.py
@@ -108,9 +108,9 @@ class FesomDataset:
 
         self.len = (self.end_idx - self.start_idx) // self.mesh_size
 
-        assert self.end_idx > self.start_idx, (
-            f"Abort: Final index of {self.end_idx} is the same of larger than start index {self.start_idx}"
-        )
+        assert (
+            self.end_idx > self.start_idx
+        ), f"Abort: Final index of {self.end_idx} is the same of larger than start index {self.start_idx}"
 
         self.colnames = list(self.ds.data.attrs["colnames"])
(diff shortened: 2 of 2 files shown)


src/weathergen/datasets/fesom_dataset.py
```diff
@@ -343,7 +342,7 @@ def normalize_target_channels(self, target: torch.tensor) -> torch.tensor:
         """
         assert target.shape[1] == len(self.target_idx)
         for i, ch in enumerate(self.target_idx):
-            target[..., i] = (target[..., i] - self.mean[ch + 2]) / self.stdev[ch + 2]
+            target[..., i] = (target[..., i] - self.mean[ch]) / self.stdev[ch]
```
jpolz, line 345: was the original +2 meant for lat and lon?

kacpnowak: Yes

src/weathergen/__init__.py
```diff
@@ -184,6 +184,7 @@ def train_continue() -> None:
 
     #########################
     if args.finetune_forecast:
+        cf.training_mode = "forecast"
```
clessig, line 187: @grassesi : I would merge this quickly. Can you add the missing training argument to the overwrite config for forecasting.

src/weathergen/datasets/fesom_dataset.py
clessig: The means and var were recomputed?

clessig (approved):
Looks good.

src/weathergen/datasets/fesom_dataset.py
kacpnowak: No, it's just nicer solution that doesn't require shift with `+2` as it's all handled earlier

jpolz: You still need to know that means and vars have the same length and order as the channels. Would a dictionary with channel names be easier to handle, i.e. require even less knowledge about the shape of the data?

kacpnowak: It would be nicer to store it, but we can't filter dictionary just like numpy array. We would need to run dict -> list -> array conversions back and forth

src/weathergen/__init__.py
kctezcan: Oh, I have been also training incorrectly then :(. Is it fixed with this line @kacpnowak ?

kacpnowak: Yes

PR #265, opened 2025-05-24
Adding base class for data readers and proper handling of time windows

### LINKED ISSUE #100: Base class and unit-tests for dataloaders

We have data-loaders for multiple datasets. They all need to handle a set of special cases, which most of them do not do at the moment. These include:
- time range of dataset does not include training/validation range
- time range of dataset only partially overlaps training/validation range
- time step of dataset (if applicable) is not aligned with window length (e.g. window length of 3h and time step of 2h). Then more some complexity is required to determine the correct index for a temporal window.

All data-loaders should be derived from a common base class. It might also be useful to have a second hierarchy level with a base class for dataset with a uniform time step (which applies to most or all simulation and reanalysis products but also some observations) and one for scattered-in-time-data (typically obs). The functionality to handle special/corner cases as the ones described above should come as much as possible from the base classes. These should also define all expected class members and initialize them with default values for empty datasets (e.g. when the dataset time range does not overlap the training/val range).

_No response_

### Additional context

_No response_

### Organisation

ECMWF

### LINKED ISSUE #266: Incorrect info message with type of data reader class
### What happened?

https://github.com/ecmwf/WeatherGenerator/blob/bfc669d27251548f66c6c6ef20d86b6656873710/src/weathergen/datasets/multi_stream_data_sampler.py#L111 :

>>> type(dataset)
<class 'type'>

expected return: the class of the data reader

Run any model.

### Version

develop

Linux

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

clessig:
## Description

Implements base class for data readers and adds missing functionality for proper sampling of data for time windows

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #100 
Closes #266 
Closes #265

## Code Compatibility

### Dependencies

### Documentation


clessig:
Status: works for anemoi with arbitrary length windows and offsets.

clessig:
> > Status: works for anemoi with arbitrary length windows and offsets.
> 
> Very good! I will have a look and implement it for the RADKLIM data reader. Thanks for the effort 👍

This are the variables the concrete class is expected to implement:

```
        # variables that need to be set / properly initialized by child classes that provide
        # concrete implementation

        self.len = 0
        self.source_idx = []
        self.target_idx = []
        self.geoinfo_idx = []
        self.source_channels = []
        self.target_channels = []

        self.mean = np.zeros(0)
        self.stdev = np.ones(0)
        self.mean_geoinfo = np.zeros(0)
        self.stdev_geoinfo = np.ones(0)

        self.frequency = np.datetime64(0, "s")
        self.ds_start_time = self.time_window_handler.zero_time
        self.ds_end_time = self.time_window_handler.zero_time
```

Please be in touch if anything is unclear or looks suspicious.

src/weathergen/datasets/data_reader_base.py
```diff
@@ -0,0 +1,497 @@
...
+        data :
+            data to be normalized
+
+        Returns
+        -------
+        Normalized data
+        """
+        assert source.shape[-1] == len(self.source_idx), "incorrect number of source channels"
+        for i, ch in enumerate(self.source_idx):
+            source[..., i] = (source[..., i] - self.mean[ch]) / self.stdev[ch]
```
jpolz, line 489: Would it hurt performance to check here if mean and stdev are in allowed ranges? Possibly with a warning and falling back to 0 and 1 if outside the range.

 The other possible location would be in the initialization of the child class when loading stats from STAC or elsewhere.

clessig: If it happens in the constructor then it wouldn't affect performance. The question I have is how we define "allowed ranges"? \mu_T \approx 273; \mu_u \approx 0; \sigma_{tp} \approx ? depending on the unit that is used.

jpolz: I thought less about physical plausibility and more about numerical issues when numbers get too small or too large. At least std>0 should be checked IMO.

clessig: Ok, that's sensible, yes.

pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
  "flash-attn; sys_platform == 'linux'",
  "polars~=1.25.2",
  "omegaconf~=2.3.0",
+ "pdbpp>=0.11.6", # TODO restore default
```
clessig: Any reason not to use the better pdb?

src/weathergen/datasets/data_reader_anemoi.py
```diff
@@ -0,0 +1,240 @@
...
+        # open  dataset to peak that it is compatible with requested parameters
+        ds0: Dataset = anemoi_datasets.open_dataset(filename)
+        # If there is no overlap with the time range, the dataset will be empty
+        if tw_handler.t_start >= ds0.dates[-1] or tw_handler.t_end <= ds0.dates[0]:
+            super().__init__(tw_handler)
+            self.init_empty()
+            return
+
+        kwargs = {}
+        if "frequency" in stream_info:
```
clessig, line 66: Since anemoi datasets have the frequency parameter to handle subsampling, I opted for using this one here. Consequently the subsampling handling is not the responsibility of the timestepped base class but of the concrete implementation.

For the format, I am suggesting to use HH:MM:SS for all deltatimes in the config. @grassesi , this would also affect step_hrs and len_hrs that should be changed to `time_window_len` and `time_window_step` and use this format.

src/weathergen/datasets/data_reader_base.py
```diff
@@ -0,0 +1,679 @@
...
+        Index of the time window.
+    tw_handler : TimeWindowHandler
+        Handler for time windows.
+
+    Returns
+    -------
+    NDArray[np.int32]
+        Array of dataset indexes corresponding to the time window.
+
+    dataset_start_time and period must be aligned with the time window handler.
```
clessig: What do you mean by this? I don't think we have such an assumption below

src/weathergen/datasets/data_reader_base.py
```diff
@@ -0,0 +1,679 @@
...
+    data_start_time: NPDT64,
+    data_end_time: NPDT64 | None,
+    period: NPTDel64,
+    idx: TIndex,
+    tw_handler: TimeWindowHandler,
+) -> (NDArray[np.int32], DTRange):
+    """
+    Get dataset indexes for a given time window index, when the dataset is periodic.
+
+    Keeping this function separate for testing purposes.
```
clessig, line 640: Do we want to do this also in the long term?

src/weathergen/datasets/data_reader_anemoi.py
```diff
@@ -0,0 +1,240 @@
...
+        -------
+        None
+        """
+
+        # open  dataset to peak that it is compatible with requested parameters
+        ds0: Dataset = anemoi_datasets.open_dataset(filename)
+        # If there is no overlap with the time range, the dataset will be empty
+        if tw_handler.t_start >= ds0.dates[-1] or tw_handler.t_end <= ds0.dates[0]:
+            super().__init__(tw_handler)
+            self.init_empty()
```
clessig, line 62: Needed to add this for CERRA, which does not overlap validation period currently.

The `init_empty()` is not very elegant but if we want to have abstract properties then we cannot initialize them in the base class. From my point of view, I rather initalize them to the default and the concrete implementations overwrite what is needed, but it's a trade-off here and I can see pros and cons either way.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -269,60 +301,62 @@ def __iter__(self):
...
+                for stream_info, stream_ds in zip(self.streams, self.streams_datasets, strict=True):
                     stream_data = StreamData(
                         forecast_dt + self.forecast_offset, nhc_source, nhc_target
                     )
+                    name = stream_info["name"]
 
                     # for all sources for current stream
                     for _, ds in enumerate(stream_ds):
                         # source window (of potentially multi-step length)
-                        (coords, geoinfos, source, times) = ds.get_source(idx)
```
clessig, line 302: I think we should remove it but it was used by @kacpnowak so we need to discuss. The cleanest solution would be too have independent window lengths for source and target and then this could be handled cleanly (but would require yet more changes in this PR, not much but still).

src/weathergen/datasets/data_reader_base.py
```diff
@@ -0,0 +1,679 @@
...
+            self.time_window_handler,
+        )
+
+
+# to avoid rounding issues
+# The basic time precision is 1 second.
+t_epsilon = np.timedelta64(1, "s")
+
+
+def get_dataset_indexes_timestep(
```
clessig, line 630: renamed `_periodic`; it's not what is used ESM and most people would have an incorrect association.

src/weathergen/datasets/data_reader_obs.py
```diff
@@ -0,0 +1,247 @@
...
+
+        rdata = ReaderData(
+            coords=coords,
+            geoinfos=geoinfos,
+            data=data,
+            datetimes=datetimes,
+        )
+
+        dtr = self.time_window_handler.window(idx)
+        check_reader_data(rdata, dtr)
```
clessig, line 245: Do we want to keep this in the life code? Most checks are cheap but some of the datetime correctness test can be somewhat expensive (but maybe best to just do something about it when it shows up in the profiler).

src/weathergen/datasets/data_reader_obs.py
```diff
@@ -0,0 +1,247 @@
...
+    DataReaderBase,
+    ReaderData,
+    TimeWindowHandler,
+    check_reader_data,
+)
+
+_logger = logging.getLogger(__name__)
+
+
+class DataReaderObs(DataReaderBase):
```
clessig, line 28: @kacpnowak : can you test this with your life dataset. My test one might be outdated; it works there.

src/weathergen/datasets/data_reader_base.py
```diff
@@ -0,0 +1,679 @@
...
+    # adjust start_idx if not exactly on start time
+    if (delta_t_start % period) > np.timedelta64(0, "s"):
+        # empty window in between two timesteps
+        if start_didx == end_didx:
+            return (np.array([], dtype=np.int32), dtr)
+        start_didx += 1
+
+    end_didx = start_didx + int((dtr.end - dtr.start - t_epsilon) / period)
+
+    return (np.arange(start_didx, end_didx + 1, dtype=np.int32), dtr)
```
clessig, line 690: Returns now also the `dtr`. This is for `check_reader_data()`. I wouldn't rename the function or anything but happy to discuss

src/weathergen/datasets/data_reader_anemoi.py
```diff
@@ -0,0 +1,240 @@
...
+
+        period = np.timedelta64(ds.frequency)
+        data_start_time = ds.dates[0]
+        data_end_time = ds.dates[-1]
+        assert data_start_time is not None and data_end_time is not None, (
+            data_start_time,
+            data_end_time,
+        )
+        super().__init__(
+            tw_handler,
```
clessig, line 86: I think it would be useful to also collect the stream_info in the base class. The reflexivity would help and make it future-proof, though we don't use it currently.

clessig (commented):
Code working now for Anemoi, Obs, and FESOM, as far as I can tell. Things that require attention have been flagged as comments.

src/weathergen/datasets/data_reader_anemoi.py
grassesi: The decision will be if we are trying to be backward compatible to old configs: Since we do not guarantee backward compatibility yet but it would surely disturb current efforts if we break it, I would suggest a compromise: Add `time_window_len` and `time_window_step` but do not remove `step_hrs` and `len_hrs` yet. Instead fail over to use/convert the latter if the former is not in the config and raise a `DeprecationWarning`. After a month or so remove `step_hrs` and `len_hrs` breaking backward compatibility. The code to handle the failover and deprecation warning should be implemented in a separate method (something like: `get_time_window_len(config) -> np.timedelta`). `step_hrs` and `len_hrs` have to be replaced by `time_window_len` and `time_window_step` in `default_config.yml`.

clessig: It would be great to have transition period to not break things. 

Do we agree to use HH:MM:SS for all deltatimes?

src/weathergen/datasets/data_reader_obs.py
kacpnowak: It's not working with FESOM dataset. There's no way to get `self.hrly_index` to work without some major changes

clessig: The FESOM dataset that I adapted didn't get committed, instead an empty file. Will need to recover it.

src/weathergen/datasets/data_reader_fesom.py
```diff
@@ -0,0 +1,236 @@
...
+    DataReaderTimestep,
+    ReaderData,
+    TimeWindowHandler,
+    check_reader_data,
+)
+
+_logger = logging.getLogger(__name__)
+
+
+class DataReaderFesom(DataReaderTimestep):
```
clessig, line 28: @kacpnowak : so here's the version to test. Can you in particular have a look at the logic in _get that we get the correct time steps. I had a quick look but a second pair of eyes and some proper testing would be good. (check_reader_data does sanity check but we wouldn't catch missing time steps).

pyproject.toml
tjhunter: this is a dev dependency, it should not pollute the official deps

src/weathergen/datasets/data_reader_anemoi.py
tjhunter: I agree

src/weathergen/datasets/data_reader_anemoi.py
tjhunter: that sounds good to me

src/weathergen/datasets/data_reader_anemoi.py
```diff
@@ -0,0 +1,242 @@
...
+        if tw_handler.t_start >= ds0.dates[-1] or tw_handler.t_end <= ds0.dates[0]:
+            name = stream_info["name"]
+            _logger.warning(f"{name} is not supported over data loader window. Stream is skipped.")
+            super().__init__(tw_handler)
+            self.init_empty()
+            return
+
+        kwargs = {}
+        if "frequency" in stream_info:
+            t = datetime.datetime.strptime(stream_info["frequency"], "%H:%M:%S")
```
tjhunter: let's collect these 2 lines in a utility functions, it will be used in multiple readers

src/weathergen/datasets/data_reader_fesom.py
kacpnowak, line 1: I tested it with FESOM and got this error:
```
2: 2025-06-11 15:11:33,802 2488518 trainer.py:241 : DEBUG    : steps_decay=-4342 lr_steps=10
2: 2025-06-11 15:11:33,802 2488518 trainer.py:250 : WARNING  : cf.lr_steps_warmup and cf.lr_steps_cooldown were larger than cf.lr_steps=10. The value have been adjusted to cf.lr_steps_warmup=1 and  cf.lr_steps_cooldown=0 so that steps_decay=9.
2: Traceback (most recent call last):
2:   File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/__init__.py", line 156, in train_with_args
2:     trainer.run(cf)
2:   File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/train/trainer.py", line 251, in run
2:     self.lr_scheduler = LearningRateScheduler(
2:                         ^^^^^^^^^^^^^^^^^^^^^^
2:   File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/train/lr_scheduler.py", line 89, in __init__
2:     self.scheduler_warmup = OneCycleLR(
2:                             ^^^^^^^^^^^
2:   File "/work/ab0995/a270225/WeatherGenerator2/.venv/lib/python3.12/site-packages/torch/optim/lr_scheduler.py", line 2120, in __init__
2:     super().__init__(optimizer, last_epoch, verbose)
2:   File "/work/ab0995/a270225/WeatherGenerator2/.venv/lib/python3.12/site-packages/torch/optim/lr_scheduler.py", line 149, in __init__
2:     self._initial_step()
2:   File "/work/ab0995/a270225/WeatherGenerator2/.venv/lib/python3.12/site-packages/torch/optim/lr_scheduler.py", line 154, in _initial_step
2:     self.step()
2:   File "/work/ab0995/a270225/WeatherGenerator2/.venv/lib/python3.12/site-packages/torch/optim/lr_scheduler.py", line 241, in step
2:     values = self.get_lr()
2:              ^^^^^^^^^^^^^
2:   File "/work/ab0995/a270225/WeatherGenerator2/.venv/lib/python3.12/site-packages/torch/optim/lr_scheduler.py", line 2162, in get_lr
2:     pct = (step_num - start_step) / (end_step - start_step)
2:           ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
2: ZeroDivisionError: float division by zero
```

clessig: This should have nothing to do with the refactor. I assume your time period is too short. That's the config I used:

```
streams_directory: "./config/streams/streams_ocean/"

embed_orientation: "channels"
embed_local_coords: True
embed_centroids_local_coords: False
embed_size_centroids: 0
embed_unembed_mode: "block"

target_cell_local_prediction: True

ae_local_dim_embed: 1024
ae_local_num_blocks: 2
ae_local_num_heads: 16
ae_local_dropout_rate: 0.1
ae_local_with_qk_lnorm: True

ae_local_num_queries: 2
ae_local_queries_per_cell: False
ae_adapter_num_heads: 16
ae_adapter_embed: 128
ae_adapter_with_qk_lnorm: True
ae_adapter_dropout_rate: 0.1

ae_global_dim_embed: 2048
ae_global_num_blocks: 10
ae_global_num_heads: 32
ae_global_dropout_rate: 0.1
ae_global_with_qk_lnorm: True
ae_global_att_dense_rate: 0.2
ae_global_block_factor: 64
ae_global_mlp_hidden_factor: 2

pred_adapter_kv: False
pred_self_attention: True
pred_dyadic_dims: False
pred_mlp_adaln: True

forecast_offset : 1
forecast_delta_hrs: 0
forecast_steps: 0
forecast_policy: null
forecast_freeze_model: False
forecast_att_dense_rate: 0.25
fe_num_blocks: 0
fe_num_heads: 16
fe_dropout_rate: 0.1
fe_with_qk_lnorm: True

healpix_level: 5

with_mixed_precision: True
with_flash_attention: True
compile_model: False
with_fsdp: True

loss_fcts:
  -
    - "mse"
    - 1.0
loss_fcts_val:
  -
    - "mse"
    - 1.0

batch_size: 1
batch_size_validation: 1

# training mode: "forecast" or "masking" (masked token modeling)
# for "masking" to train with auto-encoder mode, forecast_offset should be 0
training_mode: "forecast"
# masking rate when training mode is "masking"; ignored in foreacast mode
masking_rate: 0.6
masking_rate_sampling: True
sampling_rate_target: 1.0

num_epochs: 24
samples_per_epoch: 4096
samples_per_validation: 512
shuffle: True

lr_scaling_policy: "sqrt"
lr_start: 0.000001
lr_max: 0.00003
lr_final_decay: 0.000001
lr_final: 0.0
lr_steps_warmup: 256
lr_steps_cooldown: 4096
lr_policy_warmup: "cosine"
lr_policy_decay: "linear"
lr_policy_cooldown: "linear"

grad_clip: 5.0
weight_decay: 0.1
norm_type: "LayerNorm"
nn_module: "te"

# start_date: 197901010000
# end_date: 202012310000
start_date: 195001010000
end_date: 195012310000
start_date_val: 202101010000
end_date_val: 202201010000
len_hrs: 6
step_hrs: 6
input_window_steps: 1

val_initial: False

loader_num_workers: 8 # TODO restore default
log_validation: 0

istep: 0
run_history: []

desc: ""
data_loader_rng_seed: ???
run_id: ???

train_log:
  log_interval: 20

```

src/weathergen/datasets/data_reader_base.py
tjhunter: did you push your changes?

src/weathergen/datasets/data_reader_base.py
tjhunter: I would, because it is pure and it will probably get more complex. This way it is straightforward to test.

src/weathergen/datasets/data_reader_fesom.py
tjhunter: I think FESOM is high frequency? I changed t_epsilon, which should help

src/weathergen/datasets/data_reader_fesom.py
```diff
@@ -0,0 +1,236 @@
...
+
+        super().__init__(
+            tw_handler,
+            start_ds,
+            end_ds,
+            period,
+        )
+        self.colnames = list(self.ds.data.attrs["colnames"])
+        self.cols_idx = list(np.arange(len(self.colnames)))
+        self.lat_index = list(self.colnames).index("lat")
```
tjhunter, line 123: don't store in self extra variables that you do not need beyond the constructor. It is hard to understand what is internal constructor detail and what is relevant for the rest of the class

src/weathergen/datasets/data_reader_fesom.py
```diff
@@ -0,0 +1,236 @@
...
+        self.lon_index = list(self.colnames).index("lon")
+        self.colnames.remove("lat")
+        self.colnames.remove("lon")
+        self.cols_idx.remove(self.lat_index)
+        self.cols_idx.remove(self.lon_index)
+        self.cols_idx = np.array(self.cols_idx)
+
+        # Ignore step_hrs, idk how it supposed to work
+        # TODO, TODO, TODO:
+        self.step_hrs = 1
```
tjhunter, line 133: it should be a timedelta or numpy.timedelta64, not an integer of arbitrary units

src/weathergen/datasets/data_reader_fesom.py
```diff
@@ -0,0 +1,236 @@
...
+        )
+        self.colnames = list(self.ds.data.attrs["colnames"])
+        self.cols_idx = list(np.arange(len(self.colnames)))
+        self.lat_index = list(self.colnames).index("lat")
+        self.lon_index = list(self.colnames).index("lon")
+        self.colnames.remove("lat")
+        self.colnames.remove("lon")
+        self.cols_idx.remove(self.lat_index)
+        self.cols_idx.remove(self.lon_index)
+        self.cols_idx = np.array(self.cols_idx)
```
tjhunter, line 129: please don't do that, it is hard to reason about the type of the variable:

You can simplify:

```py
        col_idxs: list[tuple[int, str]] = [(i, n) 
                                           for (i, n) in enumerate(self.ds.data.attrs["colnames"]) if n not in ["lat", "lon"]
                                           ]
        self.colnames: list[str] = [n, for (_, n) in col_idxs]
        self.cols_idx: NDArray[np.int32] = np.array([i for (i, _) in col_idxs], dtype=np.int32)

```

config/streams/streams_big/cerra.yml
```diff
@@ -0,0 +1,35 @@
+# # (C) Copyright 2024 WeatherGenerator contributors.
```
tjhunter: should we remove these files? I am neutral

config/streams/streams_big/geos_1.yml
```diff
@@ -0,0 +1,38 @@
+# # (C) Copyright 2024 WeatherGenerator contributors.
+# #
+# # This software is licensed under the terms of the Apache Licence Version 2.0
+# # which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# #
+# # In applying this licence, ECMWF does not waive the privileges and immunities
+# # granted to it by virtue of its status as an intergovernmental organisation
+# # nor does it submit to any jurisdiction.
+
+# METEOSAT-X, SEVIRI :
```
tjhunter: we really need to have simple names here, this breaks the metrics pipeline

config/streams/streams_big/npp_atms.yml
```diff
@@ -0,0 +1,36 @@
+# # (C) Copyright 2024 WeatherGenerator contributors.
+# #
+# # This software is licensed under the terms of the Apache Licence Version 2.0
+# # which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# #
+# # In applying this licence, ECMWF does not waive the privileges and immunities
+# # granted to it by virtue of its status as an intergovernmental organisation
+# # nor does it submit to any jurisdiction.
+
+# NPP, ATMS :
```
tjhunter: same thing

config/streams/streams_big/synop.yml
```diff
@@ -0,0 +1,33 @@
+# # obs_types
+# # 0 : polar orbiting satellites
+# # 1 : geostationay satellites
+# # 2 : conventional observations
+
+# Surface Combined :
```
tjhunter: same thing

tjhunter (commented):
I am good the way it is. I have concerns about FESOP but I would rather fix it in a follow-up. We are starting to add too much to this PR.

config/streams/streams_big/geos_1.yml
clessig: That's pretty standard: platform, instrument. But we can discuss.

PR #272, opened 2025-05-27
clean up assignment of run_ids

### LINKED ISSUE #261: Evaluation not returning new run_id by default

Before when I ran an evaluation, a new run_id was generated automatically. This is not the case anymore. How can I get a new run_id? 

There is the CLI parameter
```python
parser.add_argument(
        "--eval_run_id",
        type=str,
        required=False,
        dest="eval_run_id",
        help="(optional) if specified, uses the provided run id to store the evaluation results",
    )
``` 
but this needs me to choose the run_id.

Generate a new run_id automatically unless otherwise specified.

_No response_

### Additional context

_No response_

### Organisation

_No response_

grassesi:
## Type of Change

-   [ ] Documentation update

## Issue Number

closes #261 

## Code Compatibility

### Dependencies

### Documentation


src/weathergen/utils/config.py
```diff
@@ -118,16 +118,33 @@ def load_config(
...
     else:
-        base_config = load_model_config(run_id, epoch, private_config["model_path"])
+        base_config = load_model_config(from_run_id, epoch, private_config["model_path"])
 
     # use OmegaConf.unsafe_merge if too slow
     return OmegaConf.merge(base_config, private_config, *overwrite_configs)
 
 
+def set_run_id(config: Config, run_id: str | None, reuse_run_id: bool):
+    """
```
tjhunter, line 126: can you explain in broad strokes the logic in the docstring? I am confused myself :-)

src/weathergen/utils/config.py
```diff
@@ -118,16 +118,33 @@ def load_config(
...
-        base_config.run_id = get_run_id()
     else:
-        base_config = load_model_config(run_id, epoch, private_config["model_path"])
+        base_config = load_model_config(from_run_id, epoch, private_config["model_path"])
 
     # use OmegaConf.unsafe_merge if too slow
     return OmegaConf.merge(base_config, private_config, *overwrite_configs)
 
 
+def set_run_id(config: Config, run_id: str | None, reuse_run_id: bool):
```
tjhunter: could you make this function pure and return a new Config object? Performance is definitely not an issue here, and purity should always be preferred, it makes composing behavior easier to understand. You can start the code with 

```
config = config.copy()
...
return config
```

src/weathergen/utils/config.py
```diff
@@ -118,16 +118,33 @@ def load_config(
...
+    """
+    Determine run_id of current run.
+
+    Args:
+        config: Base configuration loaded from previous run or default.
+        run_id: Id assigned to this run. If None a new one will be generated.
+        reuse_run_id: Reuse run_id from base configuration instead.
+    """
+    if not reuse_run_id:
```
tjhunter: I would do it a bit different:

```
if reuse_run_id:
  assert config.run_id is not None, "..."
else:
  config.run_id = run_id or get_run_id()
return config
```

I would also add a few logging statements, this is an important piece in the code

tjhunter (commented):
A few comments. Does it work with `submit-slurm.py` ?

pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
  "flash-attn; sys_platform == 'linux'",
  "polars~=1.25.2",
  "omegaconf~=2.3.0",
+ "pytest-mock>=3.14.1",
```
tjhunter: this should go in dev dependencies

clessig:
> * run_id): run `train`, `train_continue` or `evaluate` with `--run_id <RUNID>` flag => assign a run_id manually to this run

Thanks for the clarifications. I am skeptical we want to support option 2 since we would need conflict resolution then, which will be very, very complex in our (locally and globally) distributed environment: everyone will name their runs run1, exp1, era5, ... and experiment tracking will rely on a unique identified. There's a separate description field. wandb also separates the unique, auto-generated id and a not necessarily unique name and description.

tjhunter (commented):
I just tried this new logic with the slurm submit script and it works. Waiting for the requested changes below

grassesi (commented):
> could you make this function pure and return a new Config object? Performance is definitely not an issue here, and purity should always be preferred, it makes composing behavior easier to understand. You can start the code with 
> 
> ```
> config = config.copy()
> ...
> return config
> ``` 

good point

src/weathergen/utils/config.py
grassesi: Thanks that is much nicer logic

src/weathergen/utils/config.py
```diff
@@ -118,16 +118,50 @@ def load_config(
...
-        base_config.run_id = get_run_id()
     else:
-        base_config = load_model_config(run_id, epoch, private_config["model_path"])
+        base_config = load_model_config(from_run_id, epoch, private_config["model_path"])
 
     # use OmegaConf.unsafe_merge if too slow
     return OmegaConf.merge(base_config, private_config, *overwrite_configs)
 
 
+def set_run_id(config: Config, run_id: str | None, reuse_run_id: bool):
```
tjhunter: small nit: `-> Config` return type

src/weathergen/utils/config.py
```diff
@@ -118,16 +118,50 @@ def load_config(
...
+        assert config.run_id is not None, "run_id loaded from previous run should not be None."
+        _logger.info(f"reusing run_id from previous run: {config.run_id}")
+    else:
+        if run_id is None:
+            # generate new id if run_id is None
+            config.run_id = run_id or get_run_id()
+            _logger.info(f"using generated run_id: {config.run_id}")
+        else:
+            config.run_id = run_id
+            _logger.info(f"using assigned run_id: {config.run_id}")
```
tjhunter: Based on conversation with @clessig , maybe make it more clear: f"using assigned run_id: {config.run_id}. If you manually selected this run_id, this is an error."

clessig:
> @clessig indeed option 2 is for the slurm wrapper. Without setting the run id at slurm level, it is hard to track the files being generated and/or the source code being copied. That script already assumes a unique run id and uses the same logic to generate a new run_id.
> 
> What I would suggest: make clear in the documentation of the CLI that the run id should not be manually set.

Yes, you are completely right--this is why this was always there. And agreed, let's make it clear in the documentation that this is not to be used manually.

src/weathergen/utils/config.py
```diff
@@ -118,16 +118,50 @@ def load_config(
...
 
 
+def set_run_id(config: Config, run_id: str | None, reuse_run_id: bool):
+    """
+    Determine and set run_id of current run.
+
+    Determining the run id should follow the following logic:
+
+    1. (default case): run train, train_continue or evaluate without any flags => generate a new run_id for this run.
+    2. (assign run_id): run train, train_continue or evaluate with --run_id <RUNID> flag => assign a run_id manually to this run
```
clessig: Add "For train this should *not* be used manually.

clessig:
When running experiments at the weekend, I realized that train_continue misses an option --new_run_id : bool. We can manually assign a new run_id but by default it should be auto-generated. Does this PR add this option. Do we have it for evaluate?

@grassesi : could be please add it before merging if not already there. Thanks!

src/weathergen/utils/config.py
```diff
@@ -106,7 +106,7 @@ def load_config(
 
     Args:
         private_home: Configuration file containing platform dependent information and secretes
-        run_id: Run/model id of pretrained WeatherGenerator model to continue training or evaluate
+        from_run_id: Run/model id of pretrained WeatherGenerator model to continue training or evaluate
```
clessig: This is run id and not model id. It's a bit of a convention but we do not use the plain weights but load all the config parameters, and potentially only overwrite selected ones. Thus, run id seems more appropriate (and this is also what the command line arg is called)

src/weathergen/utils/cli.py
```diff
@@ -113,3 +113,8 @@ def _add_model_loading_params(parser: argparse.ArgumentParser):
         default=-1,
         help="Epoch of pretrained WeatherGenerator model used (Default -1 corresponds to the last checkpoint).",
     )
+    parser.add_argument(
+        "--reuse_run_id",
+        action="store_true",
+        help="Use the id given via --from_run_id also for the current run. The storage location for artifacts will be reused as well. This might overwrite artifacts from previous runns.",
```
clessig: Typo runns -> runs. 

There should be no overwrite of artifacts. This happens (except for #280 and potentially other bugs) for *_latest.ckpt and *_latest.json. We should avoid this. It breaks reproducibility (I typically continue from the *_latest chkpt). Can we open a PR for this.

grassesi: I would say overwriting the `*_latest.*` files is expected behavior: For me using `--reuse_id_run_id` means that this run should be indistinguishable to the previous. What would the semantics otherwise? The alternative would be simply to disallow reusing run ids.

src/weathergen/utils/config.py
grassesi: I will also add a disclaimer in the CLI :)

src/weathergen/utils/config.py
grassesi: yes this is better for consistincey :)

src/weathergen/utils/config.py
```diff
@@ -113,16 +113,52 @@ def load_config(
...
     return OmegaConf.merge(base_config, private_config, *overwrite_configs)
 
 
+def set_run_id(config: Config, run_id: str | None, reuse_run_id: bool) -> Config:
+    """
+    Determine and set run_id of current run.
+
+    Determining the run id should follow the following logic:
+
+    1. (default case): run train, train_continue or evaluate without any flags => generate a new run_id for this run.
```
tjhunter, line 131: this is serious doc, thank you very much for spending the time writing it.

Issue #267, opened 2025-05-26
Masking is being repeated in `masking` training mode


kacpnowak:
### What happened?

When evaluating model in `masking` training mode, each sample should be masked randomly. However the masks are reused by what looks like 7 or 8 consecutive samples. I attach gif for illustration. 

![Image](https://github.com/user-attachments/assets/e4d0e38d-334c-47ad-99b7-4634a1927bea)

### What are the steps to reproduce the bug?

Train model with training mode set to `masking` and evaluate it on data having more than 100 samples

### Version

0.1

### Platform (OS and architecture)

Levante Linux

### Relevant log output

```shell

```

### Organisation

AWI

clessig:
@kacpnowak : Can you try this patch

```
--- a/src/weathergen/datasets/multi_stream_data_sampler.py
+++ b/src/weathergen/datasets/multi_stream_data_sampler.py
@@ -418,4 +418,7 @@ class MultiStreamDataSampler(torch.utils.data.IterableDataset):
                 + f" : dataset [{local_start},{local_end}) : [{iter_start},{iter_end})"
             )
 
+        # ensure the tokenizers use different seeds
+        self.tokenizer.reset()
+
         return iter_start, iter_end
 
--- a/src/weathergen/datasets/tokenizer_masking.py
+++ b/src/weathergen/datasets/tokenizer_masking.py
@@ -130,7 +130,9 @@ class TokenizerMasking:
         return self.size_time_embedding
 
     def reset(self) -> None:
-        self.rng = np.random.default_rng(int(time.time()))
+        worker_info = torch.utils.data.get_worker_info()
+        div_factor = worker_info.num_workers if worker_info is not None else 1
+        self.rng = np.random.default_rng(int(time.time() / div_factor))
 
     def batchify_source(
         self,
```

clessig:
Yes, open a PR.

You mean the loss is increases by 5%? At the beginning of training?

kacpnowak:
I mean the loss computed on the evaluation. I have it evaluated on one year of data to test MTM over entire seasonal cycle

PR #277, opened 2025-05-28
Fix repeating masks in masking training mode

kacpnowak:
## Description

Fixes issue with mask being repeated by each worker by setting TokenizerMasking rng based on worker's ID

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/267

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/datasets/tokenizer_masking.py
+++ b/src/weathergen/datasets/tokenizer_masking.py
@@ -122,8 +122,8 @@ class TokenizerMasking:
         )
 
         worker_info = torch.utils.data.get_worker_info()
-        id = worker_info.id if worker_info is not None else 1
-        self.rng = np.random.default_rng(int(time.time()) + id)
+        div_factor = worker_info.id if worker_info is not None else 1
+        self.rng = np.random.default_rng(int(time.time() / div_factor))
 
         self.size_time_embedding = 6
 
@@ -133,8 +133,8 @@ class TokenizerMasking:
 
     def reset(self) -> None:
         worker_info = torch.utils.data.get_worker_info()
-        id = worker_info.id if worker_info is not None else 1
-        self.rng = np.random.default_rng(int(time.time()) + id)
+        div_factor = worker_info.id if worker_info is not None else 1
+        self.rng = np.random.default_rng(int(time.time() / div_factor))
 
     def batchify_source(
         self,
--- a/src/weathergen/datasets/tokenizer_masking.py
+++ b/src/weathergen/datasets/tokenizer_masking.py
@@ -121,7 +121,9 @@ class TokenizerMasking:
             .to(torch.float32)
         )
 
-        self.rng = np.random.default_rng(int(time.time()))
+        worker_info = torch.utils.data.get_worker_info()
+        id = worker_info.id if worker_info is not None else 1
+        self.rng = np.random.default_rng(int(time.time()) + id)
 
         self.size_time_embedding = 6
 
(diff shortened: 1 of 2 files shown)


src/weathergen/datasets/tokenizer_masking.py
```diff
@@ -288,9 +292,11 @@ def id(arg):
         # select masked tokens for network input
         # TODO: implement sampling rate target
         target_tokens = [
-            torch.cat([c if p else torch.tensor([]) for c, p in zip(cc, pp, strict=True)])
-            if len(cc) > 0
-            else torch.tensor(cc)
+            (
```
clessig, line 295: There still seems to be a problem with your ruff settings

src/weathergen/datasets/tokenizer_masking.py
```diff
@@ -130,7 +132,9 @@ def get_size_time_embedding(self) -> int:
         return self.size_time_embedding
 
     def reset(self) -> None:
-        self.rng = np.random.default_rng(int(time.time()))
+        worker_info = torch.utils.data.get_worker_info()
+        id = worker_info.id if worker_info is not None else 1
+        self.rng = np.random.default_rng(int(time.time()) + id)
```
clessig: What is the advantage compared to the version that I proposed in the issue?

clessig (commented):
Many thanks for the quick fix. Just a minor question

src/weathergen/datasets/tokenizer_masking.py
kacpnowak: In your case, seed was exactly the same for each worker. By using ID it's unique to each of them

src/weathergen/datasets/tokenizer_masking.py
kacpnowak: I really don't know how it keeps happening. I'm always using `uv run --no-project --with "ruff==0.9.7" ruff format --check -n src/` before commiting

src/weathergen/datasets/tokenizer_masking.py
clessig: But the bug was in `num_workers` here instead of `id`:

div_factor = worker_info.num_workers if worker_info is not None else 1

Why the change in the second line. I would prefer the version with div_factor

tjhunter: rather than trying some math formulas, how about a hash:

```py
seed = abs(hash((worker_info.id, time.time())))
self.rng = np.random.default_rng(seed=seed)
```

Also, I would rather we have an index incremented at each reset() rather than using `time()`. This would be deterministic.

tjhunter: (and this works for None by the way)

clessig: I prefer something random; you don't want to lose 5% because you always run the optimization/training along the same sub-optimal path. There are other sources of randomness but the masking is a critical part. We can ensure that problems are repeatable by providing the sample index in stream_info. 

No strong feeling on hash versus math :)

kacpnowak: > rather than trying some math formulas, how about a hash:
> 
> ```python
> seed = abs(hash((worker_info.id, time.time())))
> self.rng = np.random.default_rng(seed=seed)
> ```
> 
> Also, I would rather we have an index incremented at each reset() rather than using `time()`. This would be deterministic.

I think both solutions work just as good.

Just one thing needs fixing, accessing the `id` in case the `worker_info` in `None` will throw `AttributeError` so the entire object needs to be hashed

tjhunter: I would not hash the entire object and just the id (or None). the reason being that there is no guarantee that other subfields of the object would be themselves hashable or even stable between repeated calls.

kacpnowak: Okay, so if you and @clessig approve this solution I'll commit it

clessig: Can we use

```
div_factor = worker_info.id if worker_info is not None else 1
self.rng = np.random.default_rng(int(time.time() / div_factor))
```

I feel more comfortable with seeds that are not so clustered.

Issue #215, opened 2025-05-06
Tokens get doubled when forecasting engine is deactivated.


mlangguth89:
### What happened?

When setting num_fe_blocks = 0 (e.g., during pretraining with a deactivated forecasting engine), the tokens from the global assimilation engine are added to themselves. This occurs because self.forecast simply returns the input tokens in this configuration.

**Code Reference:**
In [model.py](https://github.com/ecmwf/WeatherGenerator/blob/22af0121f54fc6e0fb2ae1cb2ef36ddf228b7416/src/weathergen/model/model.py#L403), the following operation is performed:
```
token = token + self.forecast(token)
```
When num_fe_blocks = 0, self.forecast returns token unchanged, leading to:
```
token = token + token
```
**Expected Behavior:**
When the forecasting engine is deactivated (num_fe_blocks = 0), the forecast_step loop should not alter the tokens.

### What are the steps to reproduce the bug?

Run training with its default configuration (cf. [here](https://github.com/ecmwf/WeatherGenerator/blob/22af0121f54fc6e0fb2ae1cb2ef36ddf228b7416/config/default_config.yml#L47)).

### Version

develop

### Platform (OS and architecture)

OS-independant

### Relevant log output

```shell

```

### Organisation

JSC

mlangguth89:
@clessig So, we won't do pretraining of the global assimilation engine for one-step forecasting in the future anymore?
If yes, we can close this issue-branch I guess

clessig:
No, the forecasting engine is not supposed to do forecasting.

mlangguth89:
Will open a PR on it soon. However, I will rather raise a `ValueError` than doing an assertion.

PR #281, opened 2025-05-29
Throw error on bad forecast engine config when forecast_steps > 0.

mlangguth89:
## Type of Change

-   [ ] Documentation update

## Issue Number

#215 

## Code Compatibility

### Dependencies

### Documentation

## Additional Notes

Introduces a check that forecasting engine is activated if `forecast_steps > 0`.

### DIFF
--- a/src/weathergen/model/model.py
+++ b/src/weathergen/model/model.py
@@ -172,7 +172,7 @@ class Model(torch.nn.Module):
         # forecasting engine
         if cf.forecast_steps > 0 and cf.fe_num_blocks == 0:
             raise ValueError("Empty forecast engine (fe_num_blocks = 0), but forecast_steps > 0")
-        
+
         self.fe_blocks = ForecastingEngine(cf, self.num_healpix_cells).create()
 
         ###############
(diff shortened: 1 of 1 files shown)


tjhunter (approved):
I have not checked deeply the logic but it looks reasonable to me.

src/weathergen/model/model.py
```diff
@@ -170,6 +170,9 @@ def create(self):
 
         ###############
         # forecasting engine
+        if cf.forecast_steps > 0 and cf.fe_num_blocks == 0:
+            raise ValueError("Empty forecast engine (fe_num_blocks = 0), but forecast_steps > 0")
```
clessig, line 174: That's now the only place where we exceptions in the whole code. I think we should be consistent with error checking and flagging.

mlangguth89: In accordance with following style guidelines of our project, I followed the Google Python Style Guide to 'not use assert statements in place of conditionals or validating preconditions' (see https://google.github.io/styleguide/pyguide.html#s2.4-exceptions).

clessig:
For the time being we will use asserts to validate assumptions about inputs/outputs/program states.

________________________________

Issue #282, opened 2025-05-30
Implement logging loss per-chanel


kacpnowak:
### Is your feature request related to a problem? Please describe.

Currently there's no way to monitor how individual chanels and loss functions are changing during training. This is critical for debugging multi-stream models during training (e.g., isolating unstable channels, finiding channels that perform worse). We can make this optional via a flag to reduce clutter. This was part of issue https://github.com/ecmwf/WeatherGenerator/issues/29

### Describe the solution you'd like

Loss is computed individually [here](https://github.com/ecmwf/WeatherGenerator/blob/65f9c85b7f34f1ac7adcc44d467e8bd3f4377442/src/weathergen/train/trainer.py#L437C49-L437C68) and averaged over channels to be stored in [`losses_all`](https://github.com/ecmwf/WeatherGenerator/blob/65f9c85b7f34f1ac7adcc44d467e8bd3f4377442/src/weathergen/train/trainer.py#L442C25-L442C54). 

The solution is to add another dimension to `losses_all` and save the individual losses there. As this array is used anyway for logging, minimal changes is required within the code. This change would fully backwards compatible.

### Organisation

AWI

clessig:
@kacpnowak , thanks for picking this up again. Would you be able to work on this and open a PR?

When this is implemented, it might be worthwhile to change losses_all into a dictionary or a data class to make it better structured. 

@tjhunter : Is there any refactoring needed to handle potential O(100) channels in TrainLogger and in the json files that track the metrics?

kacpnowak:
I'll happily do it

PR #283, opened 2025-05-30
Implement per-channel logging

kacpnowak:
## Description

Implements logging losses for each channel individually. 

Additionally fixes bugs:
- Crash when worker's ID is 0
- Crash when `private_cf` doesn't have `secrets`

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/282

## Code Compatibility

### Dependencies

### Documentation


src/weathergen/datasets/tokenizer_masking.py
```diff
@@ -134,7 +134,7 @@ def get_size_time_embedding(self) -> int:
     def reset(self) -> None:
         worker_info = torch.utils.data.get_worker_info()
         div_factor = worker_info.id if worker_info is not None else 1
-        self.rng = np.random.default_rng(int(time.time() / div_factor))
+        self.rng = np.random.default_rng(int(time.time() / (div_factor + 1)))
```
kacpnowak: @clessig fixes issue https://github.com/ecmwf/WeatherGenerator/issues/288

src/weathergen/utils/config.py
```diff
@@ -210,7 +210,10 @@ def _load_private_conf(private_home: Path | None) -> DictConfig:
     private_cf["model_path"] = (
         private_cf["model_path"] if "model_path" in private_cf.keys() else "./models"
     )
-    del private_cf["secrets"]
+
+    if "secrets" in private_cf:
+        del private_cf["secrets"]
+
```
grassesi: Thanks :)

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -63,6 +63,8 @@ def __init__(self, cf, start_date, end_date, batch_size, samples_per_epoch, shuf
         self.len = 100000000
 
         self.streams_datasets = []
+        self.max_channels = 0  # Maximum number of target channels across all streams
```
clessig: This is needed since losses_all is a tensor, and not a list / dict / data class. I think we should make the change there, and then avoid having this not so clean variable here.

src/weathergen/train/trainer.py
```diff
@@ -434,11 +434,10 @@ def compute_loss(
                                     )
                                     val_uw += temp.item()
```
clessig: `val_uw` is no longer used, is it?

src/weathergen/train/trainer.py
```diff
@@ -491,7 +492,10 @@ def train(self, epoch):
             forecast_steps = batch[-1]
             batch = self.batch_to_device(batch)
 
-            losses_all = torch.ones((len(self.loss_fcts_val), len(cf.streams))) * torch.nan
+            losses_all = (
```
clessig: See above. This should be a list / dict / data class

src/weathergen/train/trainer.py
```diff
@@ -716,7 +730,7 @@ def log(self, bidx):
                 }
                 for i_obs, rt in enumerate(self.cf.streams):
                     loss_dict["training {}".format(rt["name"].replace(",", ""))] = float(
-                        l_avg[0, i_obs]
+                        l_avg[0, i_obs, :].nanmean(0)
```
clessig: Why not `l_avg[0, i_obs].nanmean(0)` ?

src/weathergen/train/trainer.py
```diff
@@ -752,7 +766,7 @@ def log_terminal(self, bidx, epoch):
                         bidx,
                         len_dataset,
                         self.cf.istep,
-                        np.nanmean(l_avg[0]),
+                        np.nanmean(l_avg[0].detach()),
```
clessig: l_avg() is a tensor so it should be torch.nanmean( )

Why the `detach` here?

src/weathergen/train/trainer.py
clessig: Because of problems when the number of channels is 1. But `torch.nanmean( l_avg[0, i_obs])` would avoid the issue.

src/weathergen/utils/train_logger.py
```diff
@@ -69,13 +69,13 @@ def add_train(self, samples, lr, loss_avg, stddev_avg, perf_gpu=0.0, perf_mem=0.
...
         """
-
         metrics = dict(num_samples=samples)
+        loss_avg = loss_avg.detach()
 
         log_vals = [int(datetime.datetime.now().strftime("%Y%m%d%H%M%S"))]
         log_vals += [samples]
 
-        metrics["loss_avg_0_mean"] = np.nanmean(loss_avg[0])
+        metrics["loss_avg_0_mean"] = np.nanmean(loss_avg[0, :, :])
```
clessig: `, :, :]` to avoid issues when shape loss_avg is [*, 1,1]?

clessig (commented):
Thanks for making progress on this so quickly. I think having a better data structure for losses_all would simplify things. 

I leave comments for train_logger to @tjhunter .

src/weathergen/datasets/multi_stream_data_sampler.py
kacpnowak: I see the point, but I'm afraid it causes more mess downstream. To get mean values we can't simply use `nanmean` without stacking the tensors together

clessig: Having one tensor per stream seems like a good compromise. And then we wouldn't need this variable, I think?

kacpnowak: Unfortunately it also generates issues. In such case `losses_all` will have different shapes based on individual streams. Then it wont work with stacking here: 
```python
self.losses_hist += [losses_all]

.... 

losses_all = self.ddp_average(
    torch.stack(self.losses_hist).to(torch.float64).nanmean(0)
)
```

src/weathergen/train/trainer.py
kacpnowak: Should I switch it in every case for consistency?

src/weathergen/train/trainer.py
kacpnowak: I forgot to use detach loss before putting it there. Will fix it

src/weathergen/train/trainer.py
kacpnowak: It seems so, will delete it

src/weathergen/train/trainer.py
clessig: Yes, I think this would be better

src/weathergen/utils/train_logger.py
clessig: As above, don't think we need the slicing.

src/weathergen/datasets/multi_stream_data_sampler.py
clessig: There are two options:
1. perform the ddp_average separately
2. build one tensor temporarily, ddp_average, and then split again to recover the per stream data; likely will require 1d tensor to be able to stack/cat
The second option makes sure that there is no adverse performance effect from many calls to ddp_average (not sure if they are blocking), although it might have it's own perf problems to build the the joint tensor. The first option is simpler to implement, but we should test that it doesn't lead to a perf degradation (I can help with this).

kacpnowak: I'm a bit concerned about the added complexity versus the immediate benefit. The current implementation is in my opinion more straightforward to understand and maintain.

tjhunter (changes requested):
@kacpnowak  thank you for looking at this part. It is a much needed feature and a complicated part of the code base. 

I have some concerns about building up on the existing matrix of `loss_fun x streams`:

- it requires extra indexing and bookkeeping (the max_channels)
- figuring out if a NaN in losses_all is actually a NaN or just a filler is not obvious -> harder to debug

Rather than trying to put everything in a tensor at all costs, why not indeed have 
`losses_all: Dict[(str: stream name), Dict[(str:channel name), tensor<len(loss_fcts_val)>]]` (my preference) , or

`losses_all: Dict[(str: channel name), tensor<len(channels for this stream), len(loss_fcts_val)>]`

When thinking about readability and debugability, I would rather have `compute_loss` eventually have no side effects of updating various mutable tensors, as it makes reasoning harder for more junior people, hence returning parts of the losses rather than updating in place.

I believe this part is not going to be a bottleneck in performance given the complexity of assembling a prediction to begin with. Happy to have a call if it helps to move forward. What do you think?

kacpnowak:
@tjhunter @clessig thank you very much for your review. Here is revised version. 

Changes in Trainer class:
- `compute_loss` is now pure function that creates dictionaries with losses and stds for each stream
- now instead of logging only the first loss function (`log_vals += [np.nanmean(loss_avg[0])]`), the global loss is being saved for logging
- aggregation of losses across all processes has been moved to `_prepare_losses_for_logging` method
- all loging to files has been move to `log` method
- all logging to terminal has been moved to `log_terminal`
- loss function that need stds computed are now inside `_stat_loss_fcts` list 
- loss functions are being checked for computing std using `__name__` as `self.cf.loss_fcts` is might not be the same as `self.cf.loss_fcts_val`
- fixed small bug in line 437, where tensor was created on CPU causing a crash when any ensemble based loss function was used 

Changest in TrainLogger class:

- modify `add_train` and `add_val` to accomodate recieving dictionaries 
- change logging first loss function to logging the global loss

src/weathergen/train/trainer.py
```diff
@@ -34,6 +34,8 @@
 
 _logger = logging.getLogger(__name__)
 
+_stat_loss_fcts = ["stats", "kcrps"]  # Names of loss functions that need std computed
```
clessig: I think this should be defined in losses.py

src/weathergen/train/trainer.py
```diff
@@ -476,38 +490,33 @@ def compute_loss(
...
         self.ddp_model.train()
+        log_interval = self.cf.train_log.log_interval
 
         dataset_iter = iter(self.data_loader)
 
         self.optimizer.zero_grad()
-        self.losses_hist, self.stddev_hist = [], []
+
+        # Unweighted loss, real weighted loss, std for losses that need it
+        self.losses_hist, self.real_losses_hist, self.stddev_hist = [], [], []
```
clessig: Let's be more verbose here with the variable names:

self.losses_hist -> self.losses_unweighted_hist
self.real_loesses_hist -> self.losses_model_hist
self.stdev_hist -> self.stdev_unweighted_hist

I would probably also drop the plural and have `loss_` instead of `losses_`

src/weathergen/train/trainer.py
```diff
@@ -615,52 +640,22 @@ def validate(self, epoch):
...
-                if self.cf.rank == 0:
-                    print(
-                        f"validation ({cf.run_id}) : {epoch:03d} : loss = {torch.nanmean(losses_all[0].nanmean(0)):.4E}",
-                        flush=True,
-                    )
-                    for i_obs, rt in enumerate(cf.streams):
-                        print(
-                            "{}".format(rt["name"]) + f" : {losses_all[0, i_obs].nanmean(0):0.4E}"
-                        )
+                self.log(val=True)
```
clessig: We should use the stages (TRAIN, VAL) that are introduced in #265

clessig (commented):
Overall looks good to me. I didn't run it, though; let me know if I should. 

I left some smaller comments that should also still be addressed.

src/weathergen/train/trainer.py
kacpnowak: I'm not sure how to do that. I guess I can wait for you to merge that branch and then update here or the other way round

clessig: We can merge your PR and then open issues to track that the flagged change happens.

clessig: Can you take care of the PR once it's merged?

clessig:
There are some checks failing (ruff). Can you please fix.

src/weathergen/train/trainer.py
kacpnowak: Sure

config/streams/streams_ocean/fesom.yml
```diff
@@ -9,7 +9,7 @@
 
 FESOM :
   type : fesom
-  filenames : ['fesom_ifs_awi']
+  filenames : ['test4.zarr']
```
tjhunter, line 12: is it correct?

.gitignore
```diff
@@ -204,5 +204,5 @@ output/
 logs/
 models/
 results/
-
+plots/
```
tjhunter, line 207: ah yes, good point

src/weathergen/train/trainer.py
```diff
@@ -356,7 +355,14 @@ def compute_loss(
 
         ctr_ftarget = 0
         loss = torch.tensor(0.0, device=self.devices[0], requires_grad=True)
-
+        # Create list storing losses for each stream
+        losses_all = {
```
tjhunter: could you add the type as well?

```
from torch import Tensor
...
losses_all:dict[str, Tensor] = {
```

src/weathergen/train/trainer.py
```diff
@@ -356,7 +355,14 @@ def compute_loss(
...
+            st.name: torch.zeros((len(st[str(stage) + "_target_channels"]), len(loss_fcts)))
+            for st in self.cf.streams  # No nan here as later it's divided so any remaining 0s become nans
+        }  # Create tensor for each stream
+        stddev_all = {
```
tjhunter: same here

src/weathergen/train/trainer.py
```diff
@@ -356,7 +355,14 @@ def compute_loss(
 
         ctr_ftarget = 0
         loss = torch.tensor(0.0, device=self.devices[0], requires_grad=True)
-
+        # Create list storing losses for each stream
+        losses_all = {
+            st.name: torch.zeros((len(st[str(stage) + "_target_channels"]), len(loss_fcts)))
```
tjhunter, line 379: I am actually surprised this works. I see this is how we do it further below. Ok. we should clean that later

src/weathergen/train/trainer.py
```diff
@@ -470,8 +473,13 @@ def compute_loss(
         # (with each having an expected loss of 1 for an uninitalized neural net)
         loss = loss / ctr_ftarget
 
+        losses_all.update((k, v / ctr_ftarget) for k, v in losses_all.items())
```
tjhunter: At this rate, you are fully rebuilding a dict, no need for `update`:  

```py
losses_all = {(k, v / ...) for ... }
```

src/weathergen/train/trainer.py
```diff
@@ -553,13 +561,40 @@ def train(self, epoch):
...
+        Aggregates and prepares loss and standard deviation data for logging.
+
+        Returns:
+            real_loss (float): The averaged scalar loss used for backpropagation.
+            losses_all (dict[str, torch.Tensor]): Dictionary mapping each stream name to its averaged
+                per-channel loss tensor.
+            stddev_all (dict[str, torch.Tensor]): Dictionary mapping each stream name to its averaged
+                per-channel standard deviation tensor.
+        """
+        losses_all = {}
```
tjhunter: can you put the type here as well?

src/weathergen/train/trainer.py
```diff
@@ -553,13 +561,40 @@ def train(self, epoch):
 
         self.dataset.advance()
 
+    def _prepare_losses_for_logging(self):
```
tjhunter: can you return the type? `-> tuple[float, float, float]`

src/weathergen/train/trainer.py
```diff
@@ -711,44 +713,44 @@ def save_model(self, epoch: int, name=None):
...
-                loss_dict = {
-                    "training mse": float(torch.nanmean(l_avg[0])),
-                    "lr": self.lr_scheduler.get_lr(),
-                }
-                for i_obs, rt in enumerate(self.cf.streams):
-                    loss_dict["training {}".format(rt["name"].replace(",", ""))] = float(
-                        l_avg[0, i_obs]
-                    )
+        Args:
+            val (bool, optional): If True, logs are treated as validation logs.
```
tjhunter: can you put a TODO and open a ticket to convert this to a proper stage?

src/weathergen/train/trainer.py
```diff
@@ -711,44 +713,44 @@ def save_model(self, epoch: int, name=None):
...
+                                If False (default), logs are treated as training logs.
+
+        Notes:
+            - This method only executes logging on the main process (rank 0).
+            - After logging, historical loss and standard deviation records are cleared.
+        """
+        avg_loss, losses_all, stddev_all = self._prepare_losses_for_logging()
+        samples = self.cf.istep * self.cf.batch_size * self.cf.num_ranks
+
+        if self.cf.rank == 0:
```
tjhunter: I am going to have a regex against this. use utils.distributed.is_root instead

src/weathergen/train/trainer.py
```diff
@@ -711,44 +713,44 @@ def save_model(self, epoch: int, name=None):
...
                     self.perf_gpu,
                     self.perf_mem,
                 )
 
-            self.losses_hist, self.stddev_hist = [], []
+            self.loss_unweighted_hist, self.loss_model_hist, self.stdev_unweighted_hist = [], [], []
 
     ###########################################
-    def log_terminal(self, bidx, epoch):
+    def log_terminal(self, bidx: int, epoch: int, val: bool = False):
```
tjhunter: no default arg please, it is hard to track them down

src/weathergen/utils/train_logger.py
```diff
@@ -69,33 +69,39 @@ def log_metrics(self, stage: Stage, metrics: dict[str, float]) -> None:
             f.write(s.encode("utf-8"))
 
     #######################################
-    def add_train(self, samples, lr, loss_avg, stddev_avg, perf_gpu=0.0, perf_mem=0.0) -> None:
+    def add_train(
+        self, samples, lr, avg_loss, losses_all, stddev_all, perf_gpu=0.0, perf_mem=0.0
```
tjhunter: ok here for default arg because we need to clean this later and they were already there

src/weathergen/utils/train_logger.py
```diff
@@ -122,14 +128,18 @@ def add_val(self, samples, loss_avg, stddev_avg) -> None:
...
 
-        for i_obs, st in enumerate(self.cf.streams):
-            for j, (l_n, _) in enumerate(self.cf.loss_fcts):
-                metrics[_key_loss(st["name"], l_n)] = loss_avg[j, i_obs]
-                log_vals += [loss_avg[j, i_obs]]
-        if len(stddev_avg) > 0:
-            for i_obs, st in enumerate(self.cf.streams):
-                metrics[_key_stddev(st["name"])] = stddev_avg[i_obs]
-                log_vals += [stddev_avg[i_obs]]
+        for st in self.cf.streams:
```
tjhunter, line 163: we should revisit this cleaning logic later, but let's keep it as is for now

tjhunter:
@kacpnowak  my apologies! I noticed that my review had been written but not submitted.

config/streams/streams_ocean/fesom.yml
kacpnowak: It's current name of my dataset. It's going to change once we have full dataset ready.

src/weathergen/train/trainer.py
kacpnowak: Done

src/weathergen/train/trainer.py
kacpnowak: Also done

src/weathergen/train/trainer.py
kacpnowak: Done

src/weathergen/train/trainer.py
kacpnowak: Done

src/weathergen/train/trainer.py
kacpnowak: I've converted it to use stage

src/weathergen/train/trainer.py
kacpnowak: Done

src/weathergen/train/trainer.py
kacpnowak: Done

src/weathergen/train/trainer.py
```diff
@@ -568,13 +576,40 @@ def train(self, epoch):
 
         self.dataset.advance()
 
+    def _prepare_losses_for_logging(self) -> tuple[float, float, float]:
```
tjhunter: note the type does not match your explanations. By the way, when return a tuple and you need to provide an elaborate explanation, it is very useful to return a small dataclass. This way you can change the results without issues with unpacking the tuple.

```
@dataclasses.dataclass
class LoggingLosses:
   real_loss: float
   losses: dict[str, torch.Tensor] 

def _prepare_losses_for_logging ... -> LoggingLosses
```

src/weathergen/utils/train_logger.py
```diff
@@ -71,33 +92,46 @@ def log_metrics(self, stage: Stage, metrics: dict[str, float]) -> None:
...
             for j, (lf_name, _) in enumerate(self.cf.loss_fcts):
                 lf_name = _clean_name(lf_name)
-                metrics[_key_loss(st_name, lf_name)] = loss_avg[j, i_obs]
-                if len(stddev_avg) > 0:
-                    metrics[_key_stddev(st_name)] = stddev_avg[i_obs]
-                log_vals += [loss_avg[j, i_obs]]
-        if len(stddev_avg) > 0:
-            for i_obs, _rt in enumerate(self.cf.streams):
-                log_vals += [stddev_avg[i_obs]]
+                metrics[_key_loss(st["name"], lf_name)] = loss[:, j].nanmean().item()
```
tjhunter, line 126: could the .nanmean().item() happen before we supply to the logger?

src/weathergen/utils/train_logger.py
```diff
@@ -71,33 +92,46 @@ def log_metrics(self, stage: Stage, metrics: dict[str, float]) -> None:
...
+        metrics: dict[str, float] = dict(num_samples=samples)
 
-        metrics = dict(num_samples=samples)
-
-        log_vals = [int(datetime.datetime.now().strftime("%Y%m%d%H%M%S"))]
+        log_vals: list[float] = [int(datetime.datetime.now().strftime("%Y%m%d%H%M%S"))]
         log_vals += [samples]
 
-        metrics["loss_avg_0_mean"] = np.nanmean(loss_avg[0])
+        metrics["loss_avg_mean"] = avg_loss
```
tjhunter, line 113: renaming keys can cause all sorts of issues for people who depend on it downstream. In this specific name, the new name seems better and I do not see any references to the old name. Maybe it is safe to rename in this case.

@clessig any opinion on it?

src/weathergen/utils/train_logger.py
```diff
@@ -71,33 +92,46 @@ def log_metrics(self, stage: Stage, metrics: dict[str, float]) -> None:
             f.write(s.encode("utf-8"))
 
     #######################################
-    def add_train(self, samples, lr, loss_avg, stddev_avg, perf_gpu=0.0, perf_mem=0.0) -> None:
+    def add_train(
+        self,
+        samples: int,
+        lr: float,
+        avg_loss: float,
+        losses_all: dict[str, Tensor],
```
tjhunter, line 100: my only last comment would be to convert the tensors to scalars before passing them here. This logger code will eventually be moved to commons and it should not depend on pytorch. 

In the interest of time, I would be open to leave it as is and cleaning later.

tjhunter (approved):
@kacpnowak thank you very much for looking into this part of the code. I have a few small comments but it looks good otherwise.

i tried it with the integration test and it seems to work as expected. Before you merge, could you please merge the changes I had to do to make the tests work? It is small:
https://github.com/ecmwf/WeatherGenerator/tree/tjh/develop/per-channel-logginig

src/weathergen/utils/train_logger.py
clessig: I wouldn't. If we start to do even finer grain logging, knowing where the NaNs happened will likely be useful information.

src/weathergen/utils/train_logger.py
clessig: As far as I can tell, `loss_avg[0]` was just considering the first sample. We should rename if avg_loss is now properly computed across all samples.

The MLflow and plot_train code might break. But rather changing it now than later. 

I did a quick grep and this came back negative, i.e. we don't use the value elsewhere. @kacpnowak maybe you can check more carefully.

kacpnowak: I've checked and as far as I was able to check (I don't see the code for mlflow) nothing is relying on the json file. The `plot_training.py` uses txt file, where only the order matters.

src/weathergen/train/trainer.py
kacpnowak: Thanks for pointing this error. I fixed it 

I see the value that the usage of dataclass brings here, but I'm sceptical about applying it in this case as it would be an exception in the entire project. I'd be very happy to talk over a ticket with cleaning up types  and introducing things like you mentioned

src/weathergen/utils/train_logger.py
clessig: The mlflow upload code is in the private repo. plot_training uses TrainLogger, which reads the json, see [here](https://github.com/ecmwf/WeatherGenerator/blob/6f831c33ac3d18401757b022dacc752bfad80e9e/src/weathergen/utils/plot_training.py#L661) and [here](https://github.com/ecmwf/WeatherGenerator/blob/6f831c33ac3d18401757b022dacc752bfad80e9e/src/weathergen/utils/train_logger.py#L142)

src/weathergen/train/trainer.py
clessig: We use dataclasses in a few places already, e.g. [here](https://github.com/ecmwf/WeatherGenerator/blob/6f831c33ac3d18401757b022dacc752bfad80e9e/src/weathergen/datasets/data_reader_base.py#L166) (more in the same file, actually) and [here](https://github.com/ecmwf/WeatherGenerator/blob/6f831c33ac3d18401757b022dacc752bfad80e9e/src/weathergen/datasets/stream_data.py#L15).

clessig: But we should leave it here now and get the PR merged. This can be cleaned up a bit later.

Issue #294, opened 2025-06-03
Change the default network parameters


SavvasMel:
### Is your feature request related to a problem? Please describe.

Many people are training WGen on ERA5 by using the default parameters which leads to suboptimal results. I have raised this issue to suggest some network options which can offer much better performance.

### Describe the solution you'd like

I will have a PR on the new network parameters.

### Additional context

This also relates to issue https://github.com/ecmwf/WeatherGenerator/issues/126. The network parameters will be chosen based on that issue.

Issue #242, opened 2025-05-12
Fine-tune forecasting after MTM training


kctezcan:
### Is your feature request related to a problem? Please describe.

i have the MTM model trained for 84 epoches, so in the next step I want to fine-tune the model for forecasting some steps. For this I will continue training with the parameters set as :

training_mode: "forecast"
forecast_offset : 1
istep : 0

kctezcan:
Started a run here, waiting for the results

kctezcan:
See https://github.com/ecmwf/WeatherGenerator/issues/255 for a first step problem

kctezcan:
I fine-tuned the MTM model for 10 eopches (after fixing the bug with lacking mode change in finetune_forecasting mode), so now we get the full globe grid points.

Here are the results:


1-step forecasting with the MTM trained model (i.e. 0-shot forecasting):
-----------------------------------------
target:

![Image](https://github.com/user-attachments/assets/d87c2dbb-970d-46eb-94ed-4cc4e2f4d148)

prediction:

![Image](https://github.com/user-attachments/assets/5f1fdba7-9b4d-44f0-888a-7053a375336a)

difference:

![Image](https://github.com/user-attachments/assets/ded789b5-f378-46e7-8755-5ee6ae18bbac)

1-step forecasting with the MTM trained model + 10 epoches 1-step forecasting finetuning:
target:

![Image](https://github.com/user-attachments/assets/0a4427c1-1219-452a-a405-c5790367c2cd)

prediction:

![Image](https://github.com/user-attachments/assets/bb7a0be3-293f-4aa7-bc6b-13f9715e9c56)

difference:

![Image](https://github.com/user-attachments/assets/238bb6c9-1d6d-41a9-b33b-5d2c5f19888d)

Observations:
1. Notice that the observation (see the comments [here](https://github.com/ecmwf/WeatherGenerator/issues/241#issuecomment-2893919535)  in the difference maps (the difference showing `target-pred`) is still there that the forecast underestimates the sun heating from the east.
2. The RMSE error is higher for the model trained for forecasting, which should be the other way around.

Both of these require further investigation.

clessig:
Thanks for the update. There's clearly an issue with the temporal encoding that I haven't seen with obs. Maybe it's because it is constant?

2. Could you explain in detail. Is it trained versus zero-shot?

clessig:
[...]

Issue #126, opened 2025-03-25
Improve forecast skill for pure ERA5 training


clessig:
### Is your feature request related to a problem? Please describe.

The current WeatherGenerator has ok skill for training with just ERA5. The forecast skill should be improved to reach roughly Pangu-Weather skill.

### Describe the solution you'd like

A systematic ablation study should be performed to optimize the performance of the WeatherGenerator for ERA5. Parameters that should be considered need to be discussed in more detail.

CC @mlangguth89 @kctezcan

mlangguth89:
Initial experiments in an auto-encoder setting has been conducted, showing that increasing the embedding dimensions in the embedding network and the local assimilation engine can improve the performance of the model. 
However, the overall set-up was found to bee too simplistic for actually be useful for pretraining. In particular, the model has to learn the identity function in ths setting (only ERA5 in- and output, no forecasting, no masked modeling) without any real bottleneck in the data processing. Consequently, the achieved loss has been very low with test loss values of O(10^-3). 

Later experiments now focus on training the model for one-step forecasting. However, the set-up of the model is not yet established.
In a first trial, the model has been trained on forecasting, but with an deactivated forecasting engine (`fe_num_block = 0`). The attained results are rather weak with a test loss in the order of 10^-1. This translates, for instance, to an RMSE of 1.45 K of the 2m temperature for 6h forecasting. Furthermore, it has been found that the 'forecast' for 0h are actually even a bit worse, with some signature of grid artefacts in the output (see `10u` or `q_850`). 
Conceptionally, the above model set-up is also questionable since the atmospheric state at 0h and 6h lead time is sampled from the same latent space. 

Furthermore, experiments question the role of the local assimilation engine, as well as the forecasting engine. 
Below, the loss curves for three experiments are shown that has been run on two nodes of Juwel Booster (=8 A100 GPUs):

![Image](https://github.com/user-attachments/assets/1133cf96-70a8-47ac-bebc-9f276951206f)

While the experiment without a forecasting engine

mlangguth89:
Numerous experiments with varying the following hyperparameters have been performed for the one-step forecast pretraining (no forecast engine, but `forecast_offset: 1`:
- `embed_dim`
- `ae_local_dim_embed`
- `ae_global_dim_embed`
- 'ae_global_num_blocks'

The set of experiments is summarized in the following table:
[...]

PR #296, opened 2025-06-03
Minor changes to the default network parameters

SavvasMel:
## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/294

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -38,7 +38,7 @@ pred_mlp_adaln: True
 
 # number of steps offset applied to first target window; if set to zero and forecast_steps=0 then
 # one is training an auto-encoder
-forecast_offset : 1
+forecast_offset : 0
 forecast_delta_hrs: 0
 forecast_steps: 0
 forecast_policy: null
@@ -70,9 +70,9 @@ batch_size_validation: 1
 
 # training mode: "forecast" or "masking" (masked token modeling)
 # for "masking" to train with auto-encoder mode, forecast_offset should be 0
-training_mode: "forecast"
+training_mode: "masking"
 # masking rate when training mode is "masking"; ignored in foreacast mode
-masking_rate: 0.0
+masking_rate: 0.6
 # sample the masking rate (with normal distribution centered at masking_rate)
 masking_rate_sampling: True
 # sample a subset of all target points, useful e.g. to reduce memory requirements
@@ -99,7 +99,7 @@ weight_decay: 0.1
 norm_type: "LayerNorm"
 nn_module: "te"
 
-start_date: 201301010000
+start_date: 197901010000
 end_date: 202012310000
 start_date_val: 202101010000
 end_date_val: 202201010000
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -8,7 +8,7 @@ embed_unembed_mode: "block"
 
 target_cell_local_prediction: True
 
-ae_local_dim_embed: 1024
+ae_local_dim_embed: 2048
 ae_local_num_blocks: 2
 ae_local_num_heads: 16
 ae_local_dropout_rate: 0.1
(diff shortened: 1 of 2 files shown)


tjhunter:
@mlangguth89 this is a larger discussion that is worth to have soon indeed. We are building a sophisticated training pipeline and we should be able to represent it. My suggestion would be to delay this discussion a few days since it has many consequences on how we train and share models.

config/default_config.yml
```diff
@@ -38,7 +38,7 @@ pred_mlp_adaln: True
 
 # number of steps offset applied to first target window; if set to zero and forecast_steps=0 then
 # one is training an auto-encoder
-forecast_offset : 1
```
clessig, line 41: This is unrelated to improving the skill--we ask a different question when we change from forecasting to masking. I think the changes in the network parameter configs that you want to implement were with forecasting, right?

clessig (approved):
Thanks for picking this up. Could you clarify why you want to switch from forecasting to masking. Also if possible, we should document the skill change through the change in the parameters, ideally with headline scores but at least with mean MSE loss value.

config/default_config.yml
SavvasMel: So yes, this is unrelated and apologies that I forgot to mention it. The introduced changes in the network parameters will work well for both cases. I changed it since pretraining on autoencoder-decoder style with masking is the first step now anyway. But I can change it back of course.

mlangguth89:
I think we should also change the default value of `token_size` in the ERA5 stream file.
With the used dataset, we just have 2-5 input items per healpix cell as documented in the [hedgedoc](https://gitlab.jsc.fz-juelich.de/hedgedoc/xuYwPs8qRNa5PRvoK_oTyg#) related to #126. So, to avoid excessive zero-padding, we should lower `token_size`. The hedgedoc in #126 used 16, but one could probably also set to 8.

mlangguth89:
> > I think we should also change the default value of `token_size` in the ERA5 stream file. With the used dataset, we just have 2-5 input items per healpix cell as documented in the [hedgedoc](https://gitlab.jsc.fz-juelich.de/hedgedoc/xuYwPs8qRNa5PRvoK_oTyg#) related to #126. So, to avoid excessive zero-padding with the [current default](https://github.com/ecmwf/WeatherGenerator/blob/3bb9b3858739d8842a79f847f6296f0ad0cf4c74/config/streams/streams_anemoi/era5.yml#L19), we should lower `token_size`. The hedgedoc in #126 used 16, but one could probably also set to 8.
> 
> Did we do experiments that compare token size versus healpix level? The latter is likely the better option, but we should base the decision on numbers.

Not really. Kerem documented the statistics for input items per cell for the ERA5 O96 and N320 data in the hedgedoc. This shows that the default value of  `token_size: 32` is quite large for the ERA5 O96 dataset that is utilized in the default experiments. Even with Healpix level 4 grid, one has mostly less or equal than 16 input items per grid cell, and thus we decided to stick to `token_size: 16`. Yet, a systematic analysis of token size vs. Healipix level remains to be done, but for the standard experiments, the current default value does not make much sense. The input tensors to the embedding network are just inflated with many padded zeros (typically about 28 zeros per cell/channel).

SavvasMel:
@clessig @mlangguth89 @tjhunter @kctezcan The introduced changes are motivated by the recent runs that I had based on issue https://github.com/ecmwf/WeatherGenerator/issues/126 (and the respective hedgedoc) and that were replicated by @kctezcan very recently, look for example the training loss plot below that was uploaded in the "ERA5 training optimization" channel:

image

This is related also to issue and comments https://github.com/ecmwf/WeatherGenerator/issues/242#issuecomment-2909927558 and https://github.com/ecmwf/WeatherGenerator/issues/242#issuecomment-2919824673 where we can see that the performance by just using the "default" parameters (reported by @kctezcan) is not as good as the performance that is reported later on (by me). In my case, the parameters that have been introduced in this PR have been used.

I should note that the above is not related to the the `offset` parameter. We can still leave this to be `1`.

PR #306, opened 2025-06-06
[100] Refactor of data readers into clessig's branch

tjhunter:
## Description
Merging work into the branch of @clessig 

This also includes the latest pulls from the develop branch -> the refactor_data_readers_100 should be updated first


src/weathergen/datasets/data_reader_anemoi.py
```diff
@@ -153,23 +154,34 @@ def _get(self, idx: int, channels_idx: np.array) -> ReaderData:
...
+        )
+        # TODO: this would not work with sub-sampling
+        didx_start = t_idxs[0]
+        # End is inclusive
+        didx_end = t_idxs[-1] + 1
 
-        if self.len == 0 or len(t_idxs) == 0:
-            return rdata
+        if self.ds is None or self.len == 0 or len(t_idxs) == 0:
+            return ReaderData.empty()
```
clessig: This should return empty with correct number of channels.

src/weathergen/datasets/data_reader_base.py
```diff
@@ -100,45 +176,99 @@ class ReaderData:
...
+    data: NDArray[DType]
+    datetimes: NDArray[NPDT64]
 
-    def __init__(self):
-        self.coords = np.zeros([0, 0], dtype=np.float32)
-        self.geoinfos = np.zeros([0, 0], dtype=np.float32)
-        self.data = np.zeros([0, 0], dtype=np.float32)
-        self.datetimes = np.zeros([0], dtype=np.datetime64)
+    @staticmethod
+    def empty() -> "ReaderData":
```
clessig: For me it's sufficiently clear that the default constructor creates an empty object, no need for the extra function here. And dataclasses have a default constructor, which anyway create empty instances

src/weathergen/utils/better_abc.py
```diff
@@ -0,0 +1,40 @@
+"""
```
clessig, line 1: Do we really this if it's not really a python concept? In terms of functionality, we could achieve it with 

```
raise NotImplementedError()
```

src/weathergen/datasets/data_reader_base.py
```diff
@@ -9,32 +9,107 @@
...
+
+    start: NPDT64
+    end: NPDT64
+
+    def __post_init__(self):
+        assert self.start < self.end, "start time must be before end time"
+        assert self.start > _DT_ZERO, "start time must be after 1850-01-01T00:00"
+
+
+def str_to_datetime64(s: str | int | NPDT64) -> NPDT64:
```
clessig, line 68: This should happen earlier at config level. There's also an issues on this for quite a while.

src/weathergen/datasets/data_reader_base.py
tjhunter: I don't agree. The empty datasets should be provided with the shape of data to be properly aligned when stacking. This is a separate constructor.

src/weathergen/utils/better_abc.py
tjhunter: we cannot because it is a field (evaluated at initialization of the object). I tried first the same functionality with properties and functions. Then you need to create all sorts of overriding functions. 

If you look at the discussion below in SO, all alternatives involve either heavy machinery with abstract functions or do not work as expected
 https://stackoverflow.com/questions/23831510/abstract-attribute-not-property

From having a tried a few solutions, I think this keeps the spirit of the original code (fields) without introducing too much magic. The 'pythonic' alternative would be to define functions to override for each of the fields.

src/weathergen/datasets/data_reader_base.py
tjhunter: I agree but the changes are massive enough. I would make the config change as a followup?

Issue #307, opened 2025-06-06
Crash when `stats` loss is used


kacpnowak:
### What happened?

Model crashes when `stats` loss function is being used

### What are the steps to reproduce the bug?

Train model with config:
loss_fcts:
  -
    - "stats"
    - 1.0

### Version

0.1

### Platform (OS and architecture)

Levante 

### Relevant log output

```shell

```

### Organisation

AWI

PR #308, opened 2025-06-06
[Hotfix] Crash when stats loss is used

kacpnowak:
## Description

Fixes three bugs:
- crash when `stats` loss is used
- crash when `secrets` key is not present in private config
- crash when worker's ID is 0

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/307

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/datasets/tokenizer_masking.py
+++ b/src/weathergen/datasets/tokenizer_masking.py
@@ -122,7 +122,7 @@ class TokenizerMasking:
         )
 
         worker_info = torch.utils.data.get_worker_info()
-        div_factor = worker_info.id if worker_info is not None else 1
+        div_factor = (worker_info.id + 1) if worker_info is not None else 1
         self.rng = np.random.default_rng(int(time.time() / div_factor))
 
         self.size_time_embedding = 6
--- a/src/weathergen/train/trainer.py
+++ b/src/weathergen/train/trainer.py
@@ -400,7 +400,11 @@ class Trainer(Trainer_Base):
                                             target[mask, i],
                                             pred[:, mask, i],
                                             pred[:, mask, i].mean(0),
-                                            (pred[:, mask, i].std(0) if ens else torch.zeros(1)),
+                                            (
+                                                pred[:, mask, i].std(0)
+                                                if ens
+                                                else torch.zeros(1, device=pred.device)
+                                            ),
                                         )
                                         val_uw += temp.item()
                                         val = val + channel_loss_weight[i] * temp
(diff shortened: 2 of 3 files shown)


src/weathergen/utils/config.py
```diff
@@ -241,7 +241,10 @@ def _load_private_conf(private_home: Path | None) -> DictConfig:
     private_cf["model_path"] = (
         private_cf["model_path"] if "model_path" in private_cf.keys() else "./models"
     )
-    del private_cf["secrets"]
+
+    if "secrets" in private_cf:
+        del private_cf["secrets"]
```
tjhunter, line 246: I missed it. We should not use `del`. 

`private_cf.pop("secrets", None)` is all you need

src/weathergen/train/trainer.py
```diff
@@ -451,15 +455,17 @@ def compute_loss(
...
-            None
-            if not log_data
-            else [
-                preds_all,
-                targets_all,
-                targets_coords_raw_rt,
-                targets_times_raw_rt,
-                targets_lens,
-            ],
+            (
```
tjhunter, line 458: strange that the formatter gives different results here.

tjhunter (approved):
one style comment. looks good.

src/weathergen/datasets/tokenizer_masking.py
```diff
@@ -122,7 +122,7 @@ def __init__(self, healpix_level: int):
         )
 
         worker_info = torch.utils.data.get_worker_info()
-        div_factor = worker_info.id if worker_info is not None else 1
+        div_factor = (worker_info.id + 1) if worker_info is not None else 1
```
clessig, line 125: Did this actually get triggered for you? It really shouldn't

src/weathergen/train/trainer.py
clessig: We had this before with Kacper. Any idea what might trigger this behavior?

clessig (approved):

src/weathergen/datasets/tokenizer_masking.py
kacpnowak: It doesn't get triggered but I changed it just in case

src/weathergen/train/trainer.py
kacpnowak: Tim pointed out that I'm using Python 3.12.0 instead of 3.12.9

clessig: @tjhunter  Should we fix the python version more tightly then?

https://github.com/ecmwf/WeatherGenerator/blob/f0a28bd35089ad5ab570181cfe8dea008e64e8ef/pyproject.toml#L10

src/weathergen/datasets/tokenizer_masking.py
clessig: Needs to be fixed, but just wanted to make sure there were no side effects that I was not aware of / understanding.

Good exercise: why didn't it get triggered?

PR #317, opened 2025-06-10
recover num_ranks from previous run to calculate epoch_base

### LINKED ISSUE #280: epoch calculation incorrect when number of GPUs change in train_continue
### What happened?

When I start a training run with 265 GPUs and train it for 2 epoches, then continue this training on 1 GPU, it should print the loss starting from epoch 2, but it is starting from epoch 0. This is due to the calculation of the epoch in the trainer.py:
https://github.com/ecmwf/WeatherGenerator/blob/a40f502aa950e2c2907f3a9f53835fcac2833ebf/src/weathergen/train/trainer.py#L281
the cf.istep value is coorectly read from the config, but the `len(self.data_loader)` value is a lot smaller for 256 GPUs than for 1 GPU, so the `epoch_base` is caluclated incorrectly.

run a  continued training with significantly different number of GPUs than the previous training run,

### Version

0.1

linux

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

Jubeku:
## Description

When continue a training run, previously used `num_ranks` is used to calculate `epoch_base`. 

## Type of Change

-   [ ] Documentation update

## Issue Number

This is supposed to fix #280 .

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/train/trainer.py
+++ b/src/weathergen/train/trainer.py
@@ -54,6 +54,10 @@ class Trainer(Trainer_Base):
 
         self.devices = self.init_torch()
 
+        # Get num_ranks of previous run which is to be continued before
+        # num_ranks gets overwritten by current setting during init_ddp()
+        self.num_ranks_original = cf.num_ranks if "num_ranks" in cf.keys() else None
+
         self.init_ddp(cf)
 
         # read configuration of data streams
@@ -264,7 +268,15 @@ class Trainer(Trainer_Base):
         self.loss_fcts_val = [[getattr(losses, name), w] for name, w in cf.loss_fcts_val]
 
         # recover epoch when continuing run
-        epoch_base = int(self.cf.istep / len(self.data_loader))
+        if self.num_ranks_original is None:
+            epoch_base = int(self.cf.istep / len(self.data_loader))
+        else:
+            len_per_rank = (
+                (len(self.dataset) // self.num_ranks_original) // cf.batch_size
+            ) * cf.batch_size
+            epoch_base = int(
+                self.cf.istep / (min(len_per_rank, cf.samples_per_epoch) * self.num_ranks_original)
+            )
 
         # torch.autograd.set_detect_anomaly(True)
         if cf.forecast_policy is not None:


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/WeatherGenerator?pullRequest=317) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/WeatherGenerator?pullRequest=317) before we can accept your contribution.<br/><hr/>**Julian Kuehnert** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/WeatherGenerator?pullRequest=317) it.</sub>

src/weathergen/train/trainer.py
```diff
@@ -54,6 +54,10 @@ def init(
 
         self.devices = self.init_torch()
 
+        # Get num_ranks of previous, to be continued run before
+        # num_ranks gets overwritten by current setting during init_ddp()
+        self.num_ranks_original = cf.num_ranks if "num_ranks" in cf.keys() else None
```
tjhunter: it is so cheap to access we should not add it as an extra state in the class

src/weathergen/train/trainer.py
```diff
@@ -264,7 +268,15 @@ def run(self, cf, run_id_contd=None, epoch_contd=None):
         self.loss_fcts_val = [[getattr(losses, name), w] for name, w in cf.loss_fcts_val]
 
         # recover epoch when continuing run
-        epoch_base = int(self.cf.istep / len(self.data_loader))
+        if self.num_ranks_original is None:
```
tjhunter, line 271: I would calculate it here (and note the more pythonic way):

```py
num_ranks_original = self.cf.get("num_ranks", None)

```

src/weathergen/train/trainer.py
```diff
@@ -264,7 +268,15 @@ def run(self, cf, run_id_contd=None, epoch_contd=None):
...
+            epoch_base = int(self.cf.istep / len(self.data_loader))
+        else:
+            len_per_rank = (
+                (len(self.dataset) // self.num_ranks_original) // cf.batch_size
```
tjhunter: just for my sanity, can ew have `(len(self.dataset) // (self.num_ranks_original*  cf.batch_size))`

tjhunter (changes requested):
@Jubeku  thanks! I put some suggestions. As @clessig  mentioned in the ticket, we should switch to optim steps anyway later on.

src/weathergen/train/trainer.py
Jubeku: The problem is that the next line will overwrite the "num_ranks" of the original run and adapt it to the current system. That's why it needs to be captured here.

tjhunter: ok. seem my comment below about style, but looks good otherwise.

Jubeku:
Thanks for the style suggestion @tjhunter. Changed accordingly.

Issue #340, opened 2025-06-14
Evaluation package


clessig:
### Is your feature request related to a problem? Please describe.

We currently don't have an evaluation python package

### Describe the solution you'd like

Create initial python package for evaluation/scoring/diagnostics `weathergen-analysis` that uses simple plot.py that exists.

PR #359, opened 2025-06-17
[357] Sub-package for evaluation

### LINKED ISSUE #357: Investigate sub-packages with uv
The current code is structured to run on GPU clusters only (the dependencies download a CUDA environment). The newer parts of the code (evaluation) should run without GPU dependencies, but should also include their own dependencies on plotting libraries such as cartopy plotly etc.

The goal of this task is to investigate how much change is necessary so that:

- weathergen.(train, dataset) has GPU dependencies
- weathergen.evaluation has no GPU dependencies but plotting dependencies

tjhunter:
## Description

Adds a separate `weathergen-evaluate` and `weathergen-common` packages in the same source tree, while leaving the current model code intact. This has been tested for the usual commands `uv run evaluate/train/...` .

Here is the tree of dependency (showing all the dependencies are in lockstep and `weathergen-common` being a dependency of `weathergen` and `weathergen-evaluate`):

```
weathergen v0.1.0
├── anemoi-datasets v0.5.16
├── astropy-healpix v1.1.2
├── flash-attn v2.7.4.post1
├── matplotlib v3.10.1
├── numpy v2.2.4
├── omegaconf v2.3.0
├── packaging v24.2
├── pandas v2.2.3
├── polars v1.25.2
├── psutil v7.0.0
├── pynvml v12.0.0
├── torch v2.6.0+cu124
├── tqdm v4.67.1
├── weathergen-common v0.1.0
├── wheel v0.45.1
├── zarr v2.17.0
├── pytest v8.3.5 (group: dev)
├── pytest-mock v3.14.1 (group: dev)
└── ruff v0.9.7 (group: dev)
weathergen-evaluate v0.1.0
├── cartopy v0.24.1
├── plotly v6.1.2
├── weathergen-common v0.1.0 (*)
├── pytest v8.3.5 (group: dev) (*)
├── pytest-mock v3.14.1 (group: dev) (*)
└── ruff v0.9.7 (group: dev)
```

What works:
- everything. There should be no change for people working on the model/data.
- pylance / vscode has no issue navigating, hinting, type checking. It needs a refresh of the cache though.

Points of questions:
- each sub-package needs to define its own dependencies, including the dev ones such as ruff etc. There is a bit of duplication.
- each sub-package has its own version number etc. We do not use versions right and there seems to be a way to synchronize all the versions if we wanted to.

There is an example of a self-contained script called `plot.py` which depends on `weathergen-evaluate` (and all its dependencies). It can directly be called with `./plot.py` and `uv` will take care of creating the appropriate venv.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #357


src/weathergen/__init__.py
tjhunter, line 1: this file needs to be removed. it was already causing issues for nv perf tools, and in general if it exists, it prevents python from discovering sub-packages such as `weathergen.common`

tjhunter: Also, this change would need to happen in any case if we want to run any code on CPU-only code. The training code depends on flash-attn (GPU-only), so removing any reference to training or model in `__init__.py` is necessary to run a sub-package in CPU-only code

packages/common/src/weathergen/common/__init__.py
```diff
@@ -0,0 +1,2 @@
+def common_function():
+    return "This is a common function for weather generation."
```
clessig, line 9: This refers to code that is shared by the model and evaluation code? The config is something that is needed on both sides.

packages/evaluate/src/weathergen/evaluate/plot.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+from weathergen.common import common_function
+
+
+def plot():
+    print(common_function())
+
+
+if __name__ == "__main__":
```
clessig, line 18: Can we use the plot.py function that we have.

src/weathergen/run_train.py
```diff
@@ -1,9 +1,169 @@
+# (C) Copyright 2025 WeatherGenerator contributors.
```
clessig, line 1: This file should now go to the train/ directory

clessig (commented):
It seems a bit inconsistent for me to have packages/common and packages/evaluation and the "main" code is under src/. But I can live with it. config.py and validation_io.py (which is currently revised!) should be part of the common package. Anything else?

Issue #357 is also more or less a duplicate of #340.

packages/common/src/weathergen/common/__init__.py
tjhunter: Exactly, this is for common code. 

I think it is better to put skeleton first and make the move of these pieces later because it is a large change and it will probably have many conflicts. I put a TODO here for the time being and I will create a separate ticket (which can be handled by other people). How does that sound to you?

packages/evaluate/src/weathergen/evaluate/plot.py
tjhunter: yes we can. I was not sure if you wanted to make changes there first? You were saying it was not ready yet. But happy to move here first.

src/weathergen/run_train.py
tjhunter: Some slrum scripts directly refer to this file. I would rather first change these scripts to use the `train` / `evaluate` / ... commands first before moving this file. Is this good for you?

clessig (approved):

Issue #251, opened 2025-05-15
Adapt and rework validation_io module to enable more efficient evaluation workflows


grassesi:
### What maintenance does this project need?

The current `validation_io` module used for writing output/predictions has proven to be insufficient for upcoming evaluation needs. To address this the following hierarchy has been proposed:
```
\
  \<sample_idx>
     \<stream_x>
        \<forecast_step_n>
           \source
              \coords
              \geoinfo
              \data
              \times
           \target
              ...
           \prediction
              ...
     \<stream_y>
   \<sample_idx>
      ...
```
reanalysis, forecst, GEOs, SYNOP datasets will have time-stepped data that should be preserved and accessible.

The implementation should orient itself as much as possible towards the integration of existing frameworks such as xarray or pandas to keep the data accessible. 

### Organisation

JSC

clessig:
Sorry, I missed this. Some minor clarification: forecast_step_n will be a k-hour window and this might have multiple time steps in it that should be preserved.

grassesi:
So for a Stream that has regular hourly values (e.g. reanalysis) and a time window of 12 hours, I would expect the above structure where the `times` dataset will contain 12 unique values for each forecast step?

clessig:
There will be as many `times` as there are datapoint (spatial values, with X channels). But `np.unique( times)` would give 12 values.

clessig:
Is there a reason to not use zarr3 for the intermediate format? Separating the chunking and number of files is an important objective.

grassesi:
I am happy to use zarr3 although I have not collected any experience with regard to its performance/stability.

Issue #343, opened 2025-06-15
Global paths for models, results per HPC


clessig:
### Is your feature request related to a problem? Please describe.

Currently there is no consistent usage of models/, results/ dirs. 

### Describe the solution you'd like

The dirs should be shared per HPC to facilitate collaboration and ensure consistency. They should be sourced from the private repo, analogous to how data/ is obtained.

Unclear: logs, output are created on demand. The submissions script creates softlinks to these on first run? What happens when the first run is with the submission script? How can we ensure this is always local per user wrt to the WeatherGenerator git dir?

sophie-xhonneux:
environment.env can be in gitlab private, you would just have to specify the path when calling load_dotenv in python

clessig:
> environment.env can be in gitlab private, you would just have to specify the path when calling load_dotenv in python

I don't have strong feelings about which solution is used. But it doesn't make sense to me to have the paths to the data in one way and the paths to models/, results/ in another way.

grassesi:
> environment.env can be in gitlab private, you would just have to specify the path when calling load_dotenv in python

Can we store the variables in the private config eg:
```
env:
  - "dotenv_value_1"
  - "dotenv_value_2"
  - ...
```
and then load it via
```
stream = StringIO("\n".join(private_config.env))
load_dotenv(stream=stream)
```

That way we can use an existing mechanism to pass platform dependent information from WeatherGenerator-private to the core Code

sophie-xhonneux:
I think at the end of the day it does not matter that much and the most important thing is to simply make a decision and live with it in the medium term. What would you like me to implement?

tjhunter:
I have some questions regarding user behavior:

- a share directory would be read+write by all project members, which means we would not be constrained by the home quotas and garbage can accumulate. What is your experience in previous projects? I am unsure if we will need to implement our home grown GC logic (which I would much rather not do)

- that could already work right now if each user manually creates a symlink from WG/plots -> $SHARED/plots . The submit script would pick up the link. Should it be part of the onboarding instructions for each cluster? It seems a one-off operation to me, so I am not sure how much we want to automate it.

PR #365, opened 2025-06-18
Sgrasse/develop/issue 251

grassesi:
## Type of Change

-   [ ] Documentation update

## Issue Number

#251 

## Code Compatibility

### Dependencies

dask => provide parallel loading of output data

### Documentation

### DIFF
--- a/src/weathergen/evaluate/score.py
+++ b/src/weathergen/evaluate/score.py
@@ -8,7 +8,7 @@
 # nor does it submit to any jurisdiction.
 
 from dataclasses import dataclass
-from typing import List, Any
+from typing import List
 import json
 import logging
 import dask.array as da
@@ -68,27 +68,6 @@ def to_json(func):
     return wrapper
 
 
-def to_list(obj: Any) -> List:
-    """
-    Converts given object to list if obj is not already a list. Sets are also transformed to a list.
-
-    Parameters
-    ----------
-    obj : Any
-        Object to convert to list
-
-    Returns
-    ----------
-    list
-        List containing obj, or obj itself if obj was already a list
-    """
-    if isinstance(obj, (set, tuple)):
-        obj = list(obj)
-    elif not isinstance(obj, list):
-        obj = [obj]
-    return obj
-
-
 @dataclass(frozen=True)
 class VerifiedData:
     prediction: xr.DataArray
@@ -113,6 +92,44 @@ class VerifiedData:
             xr.broadcast(self.prediction, self.ground_truth)
         except ValueError as e:
             raise ValueError(f"Forecast and truth are not broadcastable: {e}")
+        
+
+def get_score(data: VerifiedData, score_name: str, agg_dims: str | List[str] = "all",
+              ens_dim: str = "ens", compute: bool = False, **kwargs) -> xr.DataArray:
+    """
+    Get the score for the given data and score name.
+    Note that the scores are aggregated over all dimensions of the prediction data by default.
+
+    Parameters
+    ----------
+    data : VerifiedData
+        VerifiedData object containing prediction and ground truth data.
+    score_name : str
+        Name of the score to calculate.
+    agg_dims : str | List[str]
+        List of dimension names over which the score will be aggregated (most often averaged).
+        If set to 'all', aggregation will be performed over all dimensions of the forecast data.
+    ens_dim : str
+        Name of the ensemble dimension in the forecast data. Only used for probabilistic scores.
+    compute : bool
+        If True, the score will be computed immediately. If False, the score will be returned
+        as a lazy xarray DataArray, which allows for efficient graph construction and execution
+    kwargs : dict
+        Additional keyword arguments to pass to the score function.
+
+    Returns
+    -------
+    xr.DataArray
+        Calculated score as an xarray DataArray.
+    """
+    sc = Scores(agg_dims=agg_dims, ens_dim=ens_dim)
+
+    score_data = sc.get_score(data, score_name, **kwargs)
+    if compute:
+        # If compute is True, compute the score immediately
+        return score_data.compute()
+
+    return score_data
 
 
 # scores class
@@ -138,8 +155,8 @@ class Scores:
         Returns
         -------
         """
-        self._agg_dims_in = agg_dims
-        self._ens_dim = ens_dim
+        self._agg_dims_in = self._validate_agg_dims(agg_dims)
+        self._ens_dim = self._validate_ens_dim(ens_dim)
 
         self.det_metrics_dict = {
             "ets": self.calc_ets,
@@ -164,7 +181,7 @@ class Scores:
             "spread": self.calc_spread,
         }
 
-    def __call__(self, data: VerifiedData, score_name: str, **kwargs):
+    def get_score(self, data: VerifiedData, score_name: str, compute: bool = False, **kwargs):
         """
         Calculate the score for the given data and score name.
 
@@ -186,6 +203,9 @@ class Scores:
             VerifiedData object containing prediction and ground truth data.
         score_name : str
             Name of the score to calculate.
+        compute : bool
+            If True, the score will be computed immediately. If False, the score will be returned
+            as a lazy xarray DataArray, which allows for efficient graph construction and execution.
         kwargs : dict
             Additional keyword arguments to pass to the score function.
 
@@ -235,30 +255,19 @@ class Scores:
 
         return result
 
-    @property
-    def agg_dims_in(self):
-        return self._agg_dims_in
-
-    @agg_dims_in.setter
-    def agg_dims(self, dims):
+    def _validate_agg_dims(self, dims):
         if dims == "all":
-            self._agg_dims_in = dims
-        else:
-            agg_dims = to_list(dims)
-            assert all([isinstance(dim) for dim in dims]), (
-                "All elements of the agg_dims must be strings."
-            )
+            return dims
+        if isinstance(dims, str):
+            return [dims]
+        if isinstance(dims, list) and all(isinstance(d, str) for d in dims):
+            return dims
+        raise ValueError("agg_dims must be 'all', a string, or list of strings.")
 
-            self._agg_dims_in = agg_dims
-
-    @property
-    def ens_dim(self):
-        return self._ens_dim
-
-    @ens_dim.setter
-    def ens_dim(self, ens_dim):
-        assert isinstance(ens_dim, str), "ens_dim must be a string."
-        self._ens_dim = ens_dim
+    def _validate_ens_dim(self, dim):
+        if not isinstance(dim, str):
+            raise ValueError("ens_dim must be a string.")
+        return dim   
 
     def _sum(self, data: xr.DataArray) -> xr.DataArray:
         """


src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -146,6 +146,7 @@ def __init__(
                 if len(ds) > 0:
                     self.len = min(self.len, len(ds) - (self.len_hrs * (fsm + 1)) // self.step_hrs)
 
+                # MODIFIES config !!!
```
tjhunter, line 150: can you put a `TODO: modifiess config. Remove that part later` . The goal is that we can do a `grep TODO` later

src/weathergen/utils/config.py
```diff
@@ -308,3 +309,19 @@ def load_streams(streams_directory: Path) -> list[Config]:
             continue
 
     return list(streams.values())
+
+
+def set_pathes(config: Config) -> Config:
```
tjhunter: grammar: `paths`

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,258 @@
...
-class MockDataChunk:
-    def __init__(self, sample, stream, forecast_step, datapoints=10, channels=1, ens_size=1):
-        # dims: sample, stream, forecast_time, valid_time, ipoint, variable, ens
-        shape = (sample, 1, 1, datapoints, channels, ens_size)
-        sampler = np.arange(np.prod(shape)).reshape(shape)
-        darr = da.from_array(sampler, chunks=tuple(max(1, s // 2) for s in shape))
-        self._xr = xr.DataArray(
-            darr,  # 1
+    # (datapoints, ???)
+    geoinfo: np.ndarray | None
```
tjhunter: for typing numpy, I believe this is recommended: `NDArray[np.float32]` , see: https://github.com/ecmwf/WeatherGenerator/blob/develop/src/weathergen/datasets/data_reader_base.py#L601

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,258 @@
...
-        return self._xr
+
+@dataclasses.dataclass
+class OutputItem:
+    target: OutputDataset
+    prediction: OutputDataset
+    source: OutputDataset | None = None
+
+    # pseudo-field (init-only variable) => kept out of most dataclasses logic
+    _meta: dataclasses.InitVar[ItemMeta | None] = None
```
tjhunter: I think that you make your life more complicated than you need to. What you really want is a constructor that has custom logic. Here is an example courtesy of chatgpt. This way, you do not need properties, you just create everything extra in that constructor methods.

```
@dataclass
class User:
    name: str
    age: int

    @classmethod
    def create(cls, name: str, birth_year: int):
        current_year = 2025
        age = current_year - birth_year
        return cls(name=name, age=age)

user = User.create("Alice", 1995)
print(user)

```

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,258 @@
...
+
+# experimental value, should be inferred more intelligently
+CHUNK_N_SAMPLES = 16392
+
+# TODO: extract geoinfo metadata, include it in xarray
+
+
+@dataclasses.dataclass
+class ItemMeta:
+    sample: str
```
tjhunter: what is this field?

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,258 @@
...
+    _meta: dataclasses.InitVar[ItemMeta | None] = None
+
+    def __post_init__(self):
+        self._meta = self.target.meta
+        if self._meta.with_source and not self.source:
+            msg = f"Missing source dataset for item: {self._meta.key}"
+            raise ValueError(msg)
+
+    @property
+    def datasets(self):
```
tjhunter: with return type annotation. Also, I would return a list, not a tuple.

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,258 @@
...
+
+    @property
+    def forecast_steps(self) -> list[int]:
+        # assume stream/samples/forecast_steps are orthogonal
+        _, example_stream = next(next(self.data_root.groups()).groups())
+        return list(example_stream.group_keys())
+
+
+@dataclasses.dataclass
+class OutputBatchData:
```
tjhunter, line 234: instead of having a list for each field, why not having a single list of inner structures? All the sources, targes, predicitons, ... lists are supposed to be aligned.

tjhunter (commented):
@grassesi some comments. I am fine if we merge it to the validation branch. I do not fully understand the logic, I would need to see it in the full context.

clessig:
Please let me have a look before we merge this PR (will try today but cannot promise).

src/weathergen/datasets/multi_stream_data_sampler.py
grassesi: Its more of a warning, but if possible I think such non-intuitive modifications of the config should be avoided if possible or at least be made more transparent. In this case I think the modification was done because in order to know `source_channels` and `target_channels` currently we rely on reading the dataset. I will add a TODO. Should I also raise an Issue for that?

src/weathergen/utils/io.py
grassesi: It should be `sample:int`, it is a unique index across all batches under which the sample will be saved.

src/weathergen/datasets/multi_stream_data_sampler.py
clessig: > Its more of a warning, but if possible I think such non-intuitive modifications of the config should be avoided if possible or at least be made more transparent. In this case I think the modification was done because in order to know `source_channels` and `target_channels` currently we rely on reading the dataset. I will add a TODO. Should I also raise an Issue for that?

It's correct that we need to think about if and when we allow modifications to the config. But it's a separate issue and PR.

src/weathergen/utils/io.py
grassesi: Unfortunatly sources is not aligned with the other lists. I know its not optimal, but having the unaltered lists passed from `write_validation` provides a clear interface to the old code and helps documenting the semantics of these lists.

pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
  "polars~=1.25.2",
  "omegaconf~=2.3.0",
  "weathergen-common",
+ "dask"
```
tjhunter, line 29: put some bounds

src/weathergen/evaluate/score.py
```diff
@@ -254,7 +254,7 @@ def calc_rmse(self, **kwargs):
         return rmse
 
     @to_json
-    def calc_acc(self, clim_mean: xr.DataArray, spatial_dims: List = ["lat", "lon"]):
+    def calc_acc(self, clim_mean: xr.DataArray, spatial_dims: list = ["lat", "lon"]):
```
tjhunter: technically you could say "list[str]" to be accurate

src/weathergen/train/trainer.py
```diff
@@ -702,9 +694,9 @@ def save_model(self, epoch: int, name=None):
                     ("_" + name) if name is not None else "",
                 ]
             )
-            base_path = Path(self.cf.model_path) / self.cf.run_id
-            file_out: Path = base_path / (filename + ".chkpt")
-            file_tmp: Path = base_path / (filename + "_tmp.chkpt")
+            base_path = config.get_path_model(self.cf)
+            file_out = base_path / (filename + ".chkpt")
```
tjhunter, line 718: these are already reviewed elsewhere, right? could you update both branches with the latest develop? I am not sure what should be reviewed

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
 
-    now = np.datetime64("now", "ns")
-    end = now + np.timedelta64(24, "h")
+    channels: list[str]
 
-    # Represent times as integers (ns since epoch)
-    now_int = now.astype("int64")
-    end_int = end.astype("int64")
+    @property
+    def datapoints(self) -> NDArray[np.int_]:
```
tjhunter, line 75: don't make it a property, it is a potentially complex calculation.

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+        return np.arange(self.data.shape[0])
 
-    # Sample n random integers in [now_int, end_int)
-    rand_ints = np.random.randint(now_int, end_int, size=n)
+    def as_xarray(self, chunk_nsamples=CHUNK_N_SAMPLES) -> xr.Dataset:
+        """Convert raw dask arrays into chunked dask-aware xarray dataset."""
+        chunks = (chunk_nsamples, *self.data.shape[1:])
```
tjhunter, line 80: what are the performance implications of having so many chunks?

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+    forecast_step: int
+    stream: str
+
+    @property
+    def key(self):
+        """Unique identifyier for one output item."""
+        return f"{self.sample}/{self.stream}/{self.forecast_step}"
+
+    @property
+    def with_source(self):
```
tjhunter, line 35: should it be rather `is_source`?

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+
+# experimental value, should be inferred more intelligently
+CHUNK_N_SAMPLES = 16392
+DType: typing.TypeAlias = np.float32
+
+# TODO: extract geoinfo metadata, include it in xarray
+
+
+@dataclasses.dataclass
+class ItemMeta:
```
tjhunter: If it is used to identify a specific chunk of data, should it be rather called `DataItemKey` ? `Meta` assumes many extra info.

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+        # TODO: is this valid for the adjusted (offsetted) forecast steps?
+        return self.forecast_step == 0
+
+
+@dataclasses.dataclass
+class OutputDataset:
+    """Access source/target/prediction zarr data contained in one output item."""
+
+    name: str
+    item: ItemMeta
```
tjhunter: key?

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
-    rand_ints = np.random.randint(now_int, end_int, size=n)
+    def as_xarray(self, chunk_nsamples=CHUNK_N_SAMPLES) -> xr.Dataset:
+        """Convert raw dask arrays into chunked dask-aware xarray dataset."""
+        chunks = (chunk_nsamples, *self.data.shape[1:])
 
-    # Convert back to datetime64
-    return list(rand_ints.astype("datetime64[ns]"))
+        # maybe do dask conversion earlier? => usefull for parallel writing?
+        data = da.from_zarr(self.data, chunks=chunks)  # dont call compute to lazy load
+        coords = da.from_zarr(self.coords).compute()
```
tjhunter, line 88: why compute the times and coords?

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+    @property
+    def with_source(self):
+        """Decide if output item should contain source dataset."""
+        # TODO: is this valid for the adjusted (offsetted) forecast steps?
+        return self.forecast_step == 0
+
+
+@dataclasses.dataclass
+class OutputDataset:
+    """Access source/target/prediction zarr data contained in one output item."""
```
tjhunter, line 44: The name is very generic, and output is always the input of something else. ForecastDataset ?

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+class OutputItem:
+    def __init__(
+        self, target: OutputDataset, prediction: OutputDataset, source: OutputDataset | None = None
+    ):
+        """Collection of possible datasets for one output item."""
+        self.target = target
+        self.prediction = prediction
+        self.source = source
+
+        self._meta = self.target.item
```
tjhunter: I would call it key or something along these lines

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+        for array_name, array in dataset:  # suffix is eg. data or coords
+            self._create_dataset(dataset_group, array_name, array)
+
+    def _create_dataset(self, group: zarr.Group, name: str, array: NDArray | None):
+        if array:  # TODO guard against geoinfo = None
+            chunks = (CHUNK_N_SAMPLES, array.shape[1:])
+            group.create_dataset(name, data=array, chunks=chunks)
+
+    @property
+    def samples(self) -> list[int]:
```
tjhunter, line 213: don't use properties, it really assumes it is a quick operation. You can cache the results with https://docs.python.org/3/library/functools.html#functools.lru_cache

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+
+    stream_names: list[str]
+
+    # stream, channel name
+    channels: list[list[str]]
+
+    sample_start: int
+    forecast_offset: int
+
+    def __post_init__(self):
```
tjhunter: you could make them functions or properties

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+        _, example_stream = next(next(self.data_root.groups()).groups())
+        return list(example_stream.group_keys())
+
+
+@dataclasses.dataclass
+class OutputBatchData:
+    """Provide convenient access to adapt existing output data structures."""
+
+    # sample, stream, (datapoint, channel) => datapoints is accross all datasets per stream
+    sources: list[list[torch.Tensor]]
```
tjhunter: do you need to have a pytorch type here? The problem is that it may resolve to a GPU-based type. It would be easier if the evaluation code did not have pytorch. I have not fully grasped what happens in this case.

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+    stream_names: list[str]
+
+    # stream, channel name
+    channels: list[list[str]]
+
+    sample_start: int
+    forecast_offset: int
+
+    def __post_init__(self):
+        self.samples = np.arange(len(self.sources)) + self.sample_start
```
tjhunter: sorry, samples and fsteps are rather generic and I struggle to understand what they mean

tjhunter (approved):
@grassesi  I have a few comments. I am approving, but I have not been able to try out the whole changes yet so I will have to do a second pass later. Happy to have a call if it helps.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -382,13 +383,8 @@ def __iter__(self):
...
 
                 # skip completely empty batch item or when all targets are empty -> no grad
-                if (
-                    np.array([s.empty() for s in streams_data]).all()
-                    or np.array([s.target_empty() for s in streams_data]).all()
-                ):
-                    continue
-
-                batch += [streams_data]
+                if not (all(s.empty() or s.target_empty() for s in streams_data)):
```
clessig, line 383: Has this been tested?

src/weathergen/evaluate/score.py
```diff
@@ -254,7 +254,7 @@ def calc_rmse(self, **kwargs):
...
         """
         Calculate anomaly correlation coefficient (ACC).
         :param clim_mean: climatological mean of the data
```
clessig: This is not the style we use for docstrings

src/weathergen/utils/config.py
```diff
@@ -22,6 +22,7 @@
 _REPO_ROOT = Path(__file__).parent.parent.parent.parent  # TODO use importlib for resources
 _DEFAULT_CONFIG_PTH = _REPO_ROOT / "config" / "default_config.yml"
 _DEFAULT_MODEL_PATH = "./models"
+_DEFAULT_RESULT_PATH = "./results"
```
clessig, line 25: We should remove these with #343

src/weathergen/utils/io.py
clessig: Our output won't always be a forecast

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+    name: str
+    item: ItemMeta
+
+    # (datapoints, channels, ens)
+    data: zarr.Array
+
+    # (datapoints,)
+    times: zarr.Array
 
+    # (datapoints, 2) => maybe more??
```
clessig: no, the rest is in geoinfos. But we should capture these as well. We can in principle cat these to coords.

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+    coords: zarr.Array
 
-def _random_datetimes_next_24h(n, seed=None):
-    """
-    Generate an array of `n` random numpy.datetime64[ns] values uniformly distributed
-    between now and 24 hours from now.
-    """
-    if seed is not None:
-        np.random.seed(seed)
+    # (datapoints, ???)
```
clessig: (datapoints, X) where X is some value >= 0

src/weathergen/utils/io.py
clessig: With the above CHUNK_N_SAMPLES it should be fast

src/weathergen/utils/io.py
clessig: compute() just means to materialize the computations?

src/weathergen/utils/mock_io.py
```diff
@@ -0,0 +1,83 @@
...
+    end_int = end.astype("int64")
+
+    # Sample n random integers in [now_int, end_int)
+    rand_ints = np.random.randint(now_int, end_int, size=n)
+
+    # Convert back to datetime64
+    return list(rand_ints.astype("datetime64[ns]"))
+
+
+class MockDataChunk:
```
clessig, line 28: Is this for testing purposes?

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+    # sample, stream, (datapoint, channel) => datapoints is accross all datasets per stream
+    sources: list[list[torch.Tensor]]
+
+    # fstep, stream, redundant dim (size 1), (sample x datapoint, channel)
+    targets: list[list[list[torch.Tensor]]]
+
+    # fstep, stream, redundant dim (size 1), (ens, sample x datapoint, channel)
+    predictions: list[list[list[torch.Tensor]]]
+
+    # fstep, stream, (sample x datapoint, 105)
```
clessig: This should be sample x datapoint x 2 (or 2 + len(geoinfos))

src/weathergen/utils/io.py
```diff
@@ -1,83 +1,281 @@
...
+    """Provide convenient access to adapt existing output data structures."""
+
+    # sample, stream, (datapoint, channel) => datapoints is accross all datasets per stream
+    sources: list[list[torch.Tensor]]
+
+    # fstep, stream, redundant dim (size 1), (sample x datapoint, channel)
+    targets: list[list[list[torch.Tensor]]]
+
+    # fstep, stream, redundant dim (size 1), (ens, sample x datapoint, channel)
+    predictions: list[list[list[torch.Tensor]]]
```
clessig: These lists of lists of lists of tensors are for backward compatibility? Then we should rather remove it.

src/weathergen/utils/io.py
grassesi: no this decides if the associated OutputItem should contain also a source dataset. I added a docstring now, I hope that clears things up

src/weathergen/utils/io.py
grassesi: Thats a good name thanks :)

src/weathergen/utils/io.py
grassesi: Yes, by default any dask operation just adds operations to the dask-DAG. Compute actually makes the results of a operation available

src/weathergen/utils/io.py
grassesi: thanks, the shape of geoinfo is stream dependent i assume?

src/weathergen/utils/io.py
grassesi: I see your point, from a packaging perspective (esp. with the new subpackages) this is really inconvenient. Maybe there is a way to load only torch types for the sake of annotation, otherwise I will just mention that it is a torch tensor in the comment above.

src/weathergen/utils/io.py
grassesi: For now I dont want to touch the compute loss logic. I think this is better done in a secondary step.

src/weathergen/utils/io.py
grassesi: Thanks for the clarification there, the 105 seemed quite magic. I took it from the docstring in `weathergen.datasets.stream_data.StreamData.add_target` Maybe it should be changed there as well?

src/weathergen/utils/io.py
clessig: But what's the reason that this is torch.tensor? It's not clear to me why we cannot cast it to np.array before.

src/weathergen/utils/io.py
clessig: Yes. the shape depends on the stream. It's 0 for ERA5 and a lot of gridded datasets and >0 for most obs.

src/weathergen/train/trainer.py
grassesi: Yes, this only replaced how `base_path` is calculated: Instead of doing the path calculation itself, this method delegates to the new `config.get_path_model` method that should serve as a system wide encapsulation of this behavior (how to correctly obtain the run-specific directories).

src/weathergen/utils/io.py
grassesi: Sure, we can do that. But for now this is just how the data is received from `compute_loss`. It would make sense to rework the logic there, but I first want to get this up and running.

src/weathergen/utils/io.py
grassesi: `samples` is a continous index of all samples accross all batches. `fsteps` is similiarly a index of all forecast steps adjusted by the forecast offset. Does that clear things up?

src/weathergen/utils/io.py
grassesi: If I understand it correctly, tensors in the `target_coords` lists are actually concatenated coords and geoinfo, where coords is just the first two columns. To obtain both correctly I need to split them.

src/weathergen/utils/io.py
grassesi: numpy arange is usually really fast, is it ok for you if I cache the result? (using eg. `functools.cached_property`)

src/weathergen/utils/mock_io.py
grassesi: Yes this is for integration purposes, it will be removed before merging the hackathon branch to develop

src/weathergen/datasets/multi_stream_data_sampler.py
grassesi: It is functionally equivalent: `AND{stream in streams}(stream is empty) OR AND{stream in streams}(stream target is empty) <=> AND{stream in streams}(stream is empty OR stream target is empty)`. Instead of using `continue` we only append to the batch if this condition is false

clessig: I see it locally but in the web interface it's difficult to see what the indentation of the different blocks is and if it is really equivalent

src/weathergen/utils/io.py
tjhunter: ok

src/weathergen/evaluate/score.py
```diff
@@ -63,79 +153,162 @@ def __init__(
...
+                }"
+            )
+
+        if self._agg_dims_in == "all":
+            # Aggregate over all dimensions of the prediction data
+            self._agg_dims = list(data.prediction.dims)
+        else:
+            # Check if _agg_dims is in prediction data
+            for dim in self._agg_dims:
+                if dim not in data.prediction.dims:
```
tjhunter, line 240: def score_metric(met:str, data, ...):
  return Score(data.dimension.dims).`met`(data)

PR #383, opened 2025-06-24
Introduce masking class and incorporate in TokenizerMasking

### LINKED ISSUE #380: Masking class and separation of tokenizer_masking

Currently masking of data happens in tokenizer_masking in batchify_source and batchify_target, and there is not a class currently for masking, which is needed to implement different masking strategies, vary them through training and so on.

The first solution is a bare bones implementation of this. Create a masker class, which implements a couple of simple masking strategies, and is instantiated in multi_stream_data_sampler before the tokenizer is instantiated. This class should take input data, and return the masked data. This needs to occur for both source and target data.

tokenizer_masking should then be adjusted to do only tokenisation, and no longer do masking at the same time.

_No response_

### Additional context

_No response_

### Organisation

_No response_

### LINKED ISSUE #408: RNGs in tokenizers are not properly initalized
### What happened?

we use time.time() to initialize rngs, which makes runs unreproducible

Run code

### Version

develop

Linux

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

shmh40:
## Description

WIP: create and instantiate a Masker class which implements masking strategies and is called in TokenizerMasking to produce masked source and target tokens. This class should allow the implementation of different masking strategies (e.g. random, per healpix cell, inpainting etc.) and should ensure the masking of source and target are properly aligned. 

Current questions:

1. The Masker returns masked source_data, and the mask perm_sel. This perm_sel is then used in tokenizer_masking, outside of the Masker. Perhaps both the source data and the target data should be produced using this Masker class? We could include, in Masker.mask, if target: use ~perm_sel to produce masked target data.
2. I haven't quite worked out how this should link with different streams, if we want to mask a whole stream and predict from another stream.
3. The second masking strategy, "block" is just a placeholder. Ideally this would be implemented with healpix cells. I need to work out/implement how to use the healpix cells in this Masker class too. Just pass them (produced in TokenizerMasking?) and use them? 

## Type of Change

-   [ ] Documentation update

## Issue Number

Resolves #380 
Resolves #408 

## Code Compatibility

### Dependencies

### Documentation


src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,160 @@
...
+        "block", MORE TO BE IMPLEMENTED...).
+        TO BE IMPLEMENTED: masking_combination (str): The strategy for combining masking
+        strategies through training (e.g., "sequential").
+        masking_rate_sampling (bool): Whether to sample the masking rate from a distribution.
+    """
+
+    def __init__(
+        self,
+        masking_rate: float,
+        masking_strategy: str = "random",
```
tjhunter: no default args please

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,160 @@
...
+        self.masking_strategy = masking_strategy
+        self.masking_combination = masking_combination
+        self.masking_rate_sampling = masking_rate_sampling
+
+    def mask(
+        self,
+        tokenized_data: list[torch.Tensor],
+        rng: np.random.Generator,
+        masking_rate: float = None,
+    ) -> tuple[list[torch.Tensor], list[np.ndarray]]:
```
tjhunter: see the PR I am asking you to review.

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,160 @@
...
+        split_indices = np.cumsum(token_lens)[:-1]
+        perm_sel = np.split(flat_mask, split_indices)
+
+        # Apply the mask to get the source data (where mask is False)
+        source_data = [data[~p] for data, p in zip(tokenized_data, perm_sel, strict=True)]
+
+        return source_data, perm_sel
+
+
+class Masker_archived:
```
tjhunter: CamelCase

tjhunter (commented):
a few style comments

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,238 @@
...
+    ):
+        self.masking_rate = masking_rate
+        self.masking_strategy = masking_strategy
+        # self.masking_combination = masking_combination
+        self.masking_rate_sampling = masking_rate_sampling
+
+    def mask_source(
+        self,
+        tokenized_data: list[torch.Tensor],
+        rng: np.random.Generator,
```
clessig: This should be part of the state, potentially passed in the constructor

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,238 @@
...
+        masking_rate: float,
```
clessig: It's passed in the constructor. Why is it passed here again? any sampling should also happen in this class

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,238 @@
+import numpy as np
+import torch
+
+
+class Masker:
+    """Class to generate boolean masks for token sequences and apply them.
```
clessig: I think the class can also be used for BERT-type masking + noising.

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,238 @@
...
+                                               out and will become the target.
+        """
+        token_lens = [len(t) for t in tokenized_data]
+        num_tokens = sum(token_lens)
+
+        # If there are no tokens, return empty lists.
+        if num_tokens == 0:
+            return tokenized_data, []
+
+        # Determine the masking rate to use for this call
```
clessig: Remove, see above

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,238 @@
...
+        rate = self.masking_rate if masking_rate is None else masking_rate
+
+        if self.masking_rate_sampling:
+            rate = np.clip(
+                np.abs(rng.normal(loc=rate, scale=1.0 / (2.5 * np.pi))),
```
clessig: We should parametrize this. But better to do it in a separate PR

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,238 @@
...
+            flat_mask = np.zeros(num_tokens, dtype=bool)
+            block_size = int(np.round(rate * num_tokens))
+            if block_size > 0 and num_tokens > 0:
+                start_index = rng.integers(0, max(1, num_tokens - block_size + 1))
+                flat_mask[start_index : start_index + block_size] = True
+        else:
+            assert False, f"Unknown masking strategy: {self.masking_strategy}"
+
+        # Ensure the mask is not degenerate (all True or all False) if there's more than one token.
+        if num_tokens > 1:
```
clessig: This should not be necessary. We should see where things break when the corner cases are hit. Maybe this has already been fixed. Otherwise it's also something for a follow-up PR.

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,238 @@
...
+            if flat_mask.all():
+                flat_mask[rng.integers(low=0, high=num_tokens)] = False
+            elif not flat_mask.any():
+                flat_mask[rng.integers(low=0, high=num_tokens)] = True
+
+        # Split the flat mask to match the structure of the tokenized data (list of lists)
+        # This will be our perm_sel
+        split_indices = np.cumsum(token_lens)[:-1]
+        perm_sel = np.split(flat_mask, split_indices)
```
clessig: Keep perm_sel internal to the class

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,238 @@
...
+                processed_target_tokens.append(torch.cat(selected_tensors))
+            else:
+                processed_target_tokens.append(
+                    torch.empty(0, feature_dim, dtype=coords.dtype, device=coords.device)
+                )
+
+        return processed_target_tokens
+
+
+class MaskerArchived:
```
clessig: What is this class about?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -336,6 +339,8 @@ def __iter__(self):
 
                             (ss_cells, ss_lens, ss_centroids) = self.tokenizer.batchify_source(
                                 stream_info,
+                                # NOTE: two unused arguments in TokenizerMasking,
+                                # still used in TokenizerForecast?
                                 self.masking_rate,
```
clessig: No, they should be removed

src/weathergen/datasets/tokenizer_utils.py
```diff
@@ -178,6 +178,7 @@ def tokenize_window_space(
     if len(source) < 2:
         return
 
+    # idx_ord_lens is length...
```
clessig: # idx_ord_lens is length is number of tokens per healpix cell

src/weathergen/datasets/tokenizer_masking.py
```diff
@@ -289,16 +286,40 @@ def id(arg):
...
-        # TODO: implement sampling rate target
-        target_tokens = [
-            (
-                torch.cat([c if p else torch.tensor([]) for c, p in zip(cc, pp, strict=True)])
-                if len(cc) > 0
-                else torch.tensor(cc)
-            )
-            for cc, pp in zip(target_tokens_cells, self.perm_sel, strict=True)
-        ]
+        ######################
```
clessig: Remove before we merge

src/weathergen/datasets/tokenizer_masking.py
```diff
@@ -280,7 +278,6 @@ def id(arg):
         )
 
         # tokenize
-        # TODO: properly set stream_id; don't forget to normalize
```
clessig, line 283: This still needs to be done

src/weathergen/datasets/tokenizer_masking.py
```diff
@@ -177,46 +171,49 @@ def batchify_source(
...
+        # NOTE: Is it already handled in the masker?
 
-        # ensure that masking is not degenerate i.e. it is not all true or false
-        if not mask.any():
-            mask[self.rng.integers(low=0, high=len(mask))] = True
-        if mask.all():
-            mask[self.rng.integers(low=0, high=len(mask))] = False
+        # if masking rate is 1.0, all tokens are masked, so the source is empty
+        # but we must compute perm_sel for the target function
+        if masking_rate == 1.0:
```
clessig: This case should be handled in the Masker and not here

src/weathergen/datasets/masking.py
shmh40: Agreed thank you - done with new commit

src/weathergen/datasets/masking.py
shmh40: Agreed

src/weathergen/datasets/masking.py
shmh40: Updated the docstring, remove boolean

src/weathergen/datasets/masking.py
shmh40: Done

src/weathergen/datasets/masking.py
shmh40: Sorry, done

src/weathergen/datasets/masking.py
shmh40: Older modified version of the class for my reference, now removed

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,153 @@
...
+    def __init__(
+        self,
+        masking_rate: float,
+        masking_strategy: str,
+        # masking_combination: str,
+        masking_rate_sampling: bool,
+    ):
+        self.masking_rate = masking_rate
+        self.masking_strategy = masking_strategy
+        # self.masking_combination = masking_combination
```
clessig: Can we remove this line before we merge.

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,153 @@
...
+        self.masking_rate_sampling = masking_rate_sampling
+        
+        # Initialize the random number generator.
+        worker_info = torch.utils.data.get_worker_info()
+        div_factor = (worker_info.id + 1) if worker_info is not None else 1        
+        self.rng = np.random.default_rng(int(time.time() / div_factor))
```
clessig, line 36: The rng seed should be passed from the cf.seed, to ensure we can reproduce maskings if we want, while ensuring it is different for the different parallel workers. 

I would suggest to keep it as is to not overload the PR but open once merged, open an issue to address the problem.

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,153 @@
...
+        # If masking_rate_sampling is enabled, sample the rate from a normal distribution.
+        if self.masking_rate_sampling:
+            rate = np.clip(
+                np.abs(self.rng.normal(loc=rate, scale=1.0 / (2.5 * np.pi))),
+                0.0,
+                1.0,
+            )
+
+        # Handle the special case where all tokens are masked
+        # NOTE: not going to handle different streams correctly.
```
clessig: I don't fully understand the comment. Can you please elaborate.

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,153 @@
...
+            source_data = [data[~p] for data, p in zip(tokenized_data, self.perm_sel, strict=True)]
+            return source_data
+
+        # Implementation of different masking strategies.
+        # Generate a flat boolean mask based on the strategy.
+
+        if self.masking_strategy == "random":
+            flat_mask = self.rng.uniform(0, 1, num_tokens) < rate
+
+        elif self.masking_strategy == "block":
```
clessig, line 95: Have you visualized this and checked correctness?

src/weathergen/datasets/masking.py
```diff
@@ -0,0 +1,153 @@
...
+            coords (torch.Tensor): Coordinates tensor, used to determine feature dimension.
+            geoinfos (torch.Tensor): Geoinfos tensor, used to determine feature dimension.
+            source (torch.Tensor): Source tensor, used to determine feature dimension.
+
+        Returns:
+            list[torch.Tensor]: The target data with masked tokens, one tensor per cell.
+        """
+
+        # check that self.perm_sel is set with an assert statement
+        assert hasattr(self, 'perm_sel'), "Masker.perm_sel must be set (in mask_source) before calling mask_target."
```
clessig: You set it in the constructor (to None) so the hasattr will alway be true. You need to test for not-None

src/weathergen/datasets/masking.py
shmh40: Yes, visualised and looks as expected. Just a placeholder for now. Healpix in next PR.

src/weathergen/datasets/masking.py
shmh40: Removed, my misunderstanding

shmh40:
Note, this PR does introduce a new config parameter, and hence other developers will be notified.

# include a masking strategy here, currently only supporting "random" and "block"
masking_strategy: "random"

Issue #382, opened 2025-06-24
A special branch to have a default workable configuration for training on ERA5


ankitpatnala:
There have been a lot of configuration changes committed to the code. In some systems, there is an issue with memory bottlenecks.

To address this issue, we will create a branch where we will fix the configuration. It will include files:

- [ ]  default_config.yml
- [ ] config/streams/era5.yml

clessig:
Can you make a suggestion for the config and open a draft PR. That's best to collaborate on this.

PR #387, opened 2025-06-24
added default forecasting config for finetuning

ankitpatnala:
## Description

This is a dummy PR to keep track of the workable configuration for training on ERA5.

- [ ]  `config/default_config.yml`
- [ ]  `forecasting_finetuning_config.yml` ( overwrite over default config for finetuning forecast experiments)
- [ ]  `config/streams/streams_anemoi/era5.yml`

## Issue Number #382 

The motivation behind this PR to keep track on the config that is working for all the collaborators across different HPCs

### DIFF
--- a/config/forecating_finetuning_config.yml
+++ b/config/forecating_finetuning_config.yml
@@ -1,0 +1,4 @@
+forecast_steps: 1
+fe_num_blocks: 4
+forecast_policy: "fixed"
+training_mode: "forecast"


clessig:
@sophie-xhonneux @Jubeku : do we use this config? If yes, we should keep it open. Otherwise, how to we track the default config to use?

Jubeku:
> Somehow this was meant for config. Would be nice if it is left open and latest config is documented here. I do not know if the config in the develop is the desired config or still in testing phase. @clessig ??

I have never used it. I think it's from @ankitpatnala.

clessig:
We need to converge on one config for the scientific experiments. I think a draft PR is a good way to track it but we can use something else if you prefer. But we do need to converge on a common config to have comparable results.

sophie-xhonneux:
Then I will send the large scale run!

PR #395, opened 2025-06-26
Continue training through slurm script

### LINKED ISSUE #370: submit slurm script does not allow to continue an experiment

The private submit.py script currently expects the passed slurm script to contain all the logic: train, evaluate, etc.

What is better?
- a single slurm script with many options
- one slurm script for each of the train, finetune, continue, evaluate/sample

It seems to me each of these stages has different hardware requirements, so I would be enclined to have one script for each stage, but I am not an HPC expert here.

_No response_

_No response_

### Additional context

_No response_

### Organisation

_No response_

Jubeku:
## Description

Allow to continue previous experiments via `WeatherGenerator-private/hpc/launch-slurm.py` by specifying `--from-run-id`.

## Type of Change

-   [ ] Documentation update

## Issue Number

Fixes #370 

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -118,6 +118,7 @@ val_initial: False
 
 loader_num_workers: 8
 log_validation: 0
+analysis_streams_output: ["ERA5"]
 
 istep: 0
 run_history: []
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,6 +97,7 @@ ignore = [
   "SIM118",
   "SIM102",
   "SIM401",
+  "UP040", # TODO: enable later
   # To ignore, not relevant for us
   "SIM108" # in case additional norm layer supports are added in future  
 ]
--- a/scripts/check_gh_issue.py
+++ b/scripts/check_gh_issue.py
@@ -13,9 +13,10 @@ Source:
 https://stackoverflow.com/questions/60717142/getting-linked-issues-and-projects-associated-with-a-pull-request-form-github-ap
 """
 
+import re
+
 import requests
 from bs4 import BeautifulSoup
-import re
 
 repo = "ecmwf/WeatherGenerator"
 
--- a/src/weathergen/datasets/data_reader_anemoi.py
+++ b/src/weathergen/datasets/data_reader_anemoi.py
@@ -202,7 +202,7 @@ class DataReaderAnemoi(DataReaderTimestep):
 
         return rd
 
-    def select_channels(self, ds0: anemoi_datasets, ch_type: str) -> NDArray[np.int32]:
+    def select_channels(self, ds0: anemoi_datasets, ch_type: str) -> NDArray[np.int64]:
         """
         Select source or target channels
 
--- a/src/weathergen/datasets/data_reader_base.py
+++ b/src/weathergen/datasets/data_reader_base.py
@@ -31,7 +31,7 @@ DType: TypeAlias = np.float32  # The type for the data in the datasets.
 """
 The type for indexing into datasets. It is a multiple of hours.
 """
-TIndex: TypeAlias = np.int32
+TIndex: TypeAlias = np.int64
 
 
 _DT_ZERO = np.datetime64("1850-01-01T00:00")
@@ -136,8 +136,8 @@ class TimeWindowHandler:
             start and end of temporal window
         """
 
-        idx_start: TIndex = np.int32(0)
-        idx_end = np.int32((self.t_end - self.t_start) // self.t_window_step)
+        idx_start: TIndex = np.int64(0)
+        idx_end = np.int64((self.t_end - self.t_start) // self.t_window_step)
         assert idx_start <= idx_end, f"time window idxs invalid: {idx_start} <= {idx_end}"
 
         return TimeIndexRange(idx_start, idx_end)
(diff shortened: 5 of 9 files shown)


src/weathergen/run_train.py
```diff
@@ -165,5 +168,10 @@ def train_with_args(argl: list[str], stream_dir: str | None):
 
 
 if __name__ == "__main__":
-    train()
-    # train_continue()
+    # Entry point for slurm script.
+    # Check whether --from_run_id is "None".
+    if sys.argv[-1] == "None":
```
tjhunter: this is expecting a certain order of the variables, which is going to break more rapidly than you think. We need something a bit more robust. What you could do is use argparse here:

```
parser = cli.get_train_parser()
(args, _) = parser.parse_known_args()
if args.from_run_id == args.run_id and args.from_run_id is not None: # (see comment in the other PR)
... continue mode
else:
... training mode

```

src/weathergen/train/trainer.py
```diff
@@ -176,7 +176,11 @@ def run(self, cf, run_id_contd=None, epoch_contd=None):
         # load model if specified
         if run_id_contd is not None:
             _logger.info(f"Continuing run with id={run_id_contd} at epoch {epoch_contd}.")
+            param_sum = sum([p.cpu().data.numpy().sum() for p in self.model.parameters()])
```
tjhunter: should we keep that?

clessig: no

src/weathergen/run_train.py
Jubeku: Totally agree.

Jubeku: Only that `get_train_parser()` doesn't have `from_run_id`. I'd need to modify _cli.py_. 
Wondering if this should rather be a quick fix or we need to restructure the whole code as we discussed on Wednesday?

src/weathergen/train/trainer.py
Jubeku: I removed this.

src/weathergen/run_train.py
clessig: We need to do it properly. There's `from_run_id` in the cli so it shouldn't be hard to adapt. @grassesi can help if necessary.

Jubeku: I'm afraid I will break people's workflow when I change the cli, so instead I introduced a specific check (`if "--from_run_id" in sys.argv`), related to my latest changes in the private repo ([PR26](https://gitlab.jsc.fz-juelich.de/esde/WeatherGenerator-private/-/merge_requests/26#note_316610)).
Probably that's not the most proper solution yet..

src/weathergen/run_train.py
```diff
@@ -165,5 +168,9 @@ def train_with_args(argl: list[str], stream_dir: str | None):
 
 
 if __name__ == "__main__":
-    train()
-    # train_continue()
+    # Entry point for slurm script.
+    # Check whether --from_run_id passed as argument.
+    if "--from_run_id" in sys.argv:
```
tjhunter, line 170: as discussed with @Jubeku , there is good solution without more refactoring . I am ok with this - provided it works

Jubeku: I'm testing if it works, also including the changes in the private repo introduced by [2a3f0c91](https://gitlab.jsc.fz-juelich.de/esde/WeatherGenerator-private/-/commit/2a3f0c916f813af9327a2d09b3d6e5760a28e50b)

tjhunter (approved):
Approving so that it can be merged without additional review (once tests pass)

src/weathergen/run_train.py
clessig: Not pretty but ok

src/weathergen/train/trainer.py
```diff
@@ -17,8 +17,13 @@
 from torch import Tensor
 from torch.distributed.fsdp import FullStateDictConfig, StateDictType
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
-from torch.distributed.fsdp.fully_sharded_data_parallel import MixedPrecision, ShardingStrategy
-from torch.distributed.fsdp.wrap import size_based_auto_wrap_policy  # default_auto_wrap_policy,
+from torch.distributed.fsdp.fully_sharded_data_parallel import (
```
clessig, line 20: Still puzzled by these reformatting issues.

clessig (approved):

src/weathergen/run_train.py
```diff
@@ -69,6 +69,9 @@ def inference_from_args(argl: list[str]):
 def train_continue() -> None:
     parser = cli.get_continue_parser()
     args = parser.parse_args()
+    if args.run_id:
+        if args.run_id == args.from_run_id:
+            args.reuse_run_id = True
```
grassesi: This is redundant code from a functional perspective, I would prefer if you remove this code. The only difference it makes is in the logging: it will be logged as `reusing run_id from previous run: {config.run_id}` instead of `using assigned run_id: {config.run_id}.  If you manually selected this run_id, this is an error.`. If this is important for you, `config.set_run_id` would be the appropriate place to implement it. However if you do that, the message should also reflect that the run id was assigned instead of using the standard `--reuse_run_id` mechanism

Jubeku: Hi @grassesi, thanks for looking into this!
You are right, I thought `--reuse_run_id` is important for model loading in the training, but it's not true, it is `run_id_contd` instead. 
Will change that as you suggested.

PR #407, opened 2025-06-27
Implementation of healpix cell masking

### LINKED ISSUE #397: Masking strategies based on healpix cells

Current masking is random over a 1D array of cells.

Implement methods to mask based on healpix level - e.g. masking whole healpix cells at arbitrary levels.

_No response_

### Additional context

_No response_

### Organisation

_No response_

shmh40:
## Description

This draft PR implements masking of cell based on healpix cells and the healpix level. Masking can be done at arbitrary healpix levels, and makes use of the nested indexing of the healpix cells.

One question: is the given implementation ok for passing args for specific masking strategies? I am not sure what style we want. For example, here for healpix masking, we want to pass the healpix level of the data, and the healpix level that we want the masking to occur on e.g. our data is healpix level 5, and we want to do very large scale masking (e.g. level 0 or 1). Of course these args are only relevant when we are doing healpix masking, so it is implemented here just as a dictionary of "strategy_kwargs" that can be passed in the config, with args specific to the masking strategy (hl_data, hl_mask), otherwise it is ignored. Hope that is ok.

## Type of Change

-   [ ] Documentation update

## Issue Number

Resolves #397.

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -72,13 +72,13 @@ batch_size_validation: 1
 # for "masking" to train with auto-encoder mode, forecast_offset should be 0
 training_mode: "masking"
 # masking rate when training mode is "masking"; ignored in foreacast mode
-masking_rate: 0.6
+masking_rate: 0.5
 # sample the masking rate (with normal distribution centered at masking_rate)
-masking_rate_sampling: True
+masking_rate_sampling: False
 # sample a subset of all target points, useful e.g. to reduce memory requirements
 sampling_rate_target: 1.0
 # include a masking strategy here, currently only supporting "random" and "block"
-masking_strategy: "block"
+masking_strategy: "healpix"
 
 num_epochs: 24
 samples_per_epoch: 4096
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -81,6 +81,11 @@ sampling_rate_target: 1.0
 # include a masking strategy here, currently only supporting "random" and "block"
 masking_strategy: "healpix"
 
+# this strategy_kwargs is a dictionary that can hold any additional parameters
+# for the masking strategy, e.g. for healpix strategy, we might need hl_data
+# and hl_mask parameters, or for different strategies, we might need different parameters
+strategy_kwargs: {"hl_data": 5, "hl_mask": 0}
+
 num_epochs: 24
 samples_per_epoch: 4096
 samples_per_validation: 512
--- a/src/weathergen/datasets/masking.py
+++ b/src/weathergen/datasets/masking.py
@@ -25,11 +25,18 @@ class Masker:
         masking_strategy: str,
         # masking_combination: str,
         masking_rate_sampling: bool,
+        # NOTE: adding strategy_kwargs to allow for strategy-specific configurations
+        # e.g., for healpix strategy, we might need hl_data and hl_mask parameters
+        strategy_kwargs: dict = None, # ### IMPROVEMENT: Added for strategy-specific config ###
+
     ):
         self.masking_rate = masking_rate
         self.masking_strategy = masking_strategy
         # self.masking_combination = masking_combination
         self.masking_rate_sampling = masking_rate_sampling
+
+        # NOTE: strategy_kwargs is a dictionary that can hold any additional parameters
+        self.strategy_kwargs = strategy_kwargs or {}
         
         # Initialize the random number generator.
         worker_info = torch.utils.data.get_worker_info()
@@ -57,6 +64,9 @@ class Masker:
         """
         token_lens = [len(t) for t in tokenized_data]
         num_tokens = sum(token_lens)
+
+        #print("Length of each token t in tokenized_data:", token_lens)
+        print("Number of tokens in the batch:", num_tokens)
 
         # If there are no tokens, return empty lists.
         if num_tokens == 0:
@@ -94,6 +104,10 @@ class Masker:
             if block_size > 0 and num_tokens > 0:
                 start_index = self.rng.integers(0, max(1, num_tokens - block_size + 1))
                 flat_mask[start_index : start_index + block_size] = True
+
+        elif self.masking_strategy == "healpix":
+            flat_mask = self._generate_healpix_mask(token_lens, rate)          
+
         else:
             assert False, f"Unknown masking strategy: {self.masking_strategy}"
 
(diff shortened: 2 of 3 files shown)


src/weathergen/datasets/masking.py
```diff
@@ -22,11 +22,18 @@ def __init__(
         masking_rate: float,
         masking_strategy: str,
         masking_rate_sampling: bool,
+        # NOTE: adding strategy_kwargs to allow for strategy-specific configurations
+        # e.g., for healpix strategy, we might need hl_data and hl_mask parameters
+        # or for different strategies, we might need different parameters?
+        strategy_kwargs: dict,
```
clessig: Thanks. This looks good

src/weathergen/datasets/masking.py
```diff
@@ -22,11 +22,18 @@ def __init__(
...
     ):
         self.masking_rate = masking_rate
         self.masking_strategy = masking_strategy
         self.masking_rate_sampling = masking_rate_sampling
 
+        # NOTE: strategy_kwargs is a dictionary that can hold any additional parameters
+        self.strategy_kwargs = strategy_kwargs or {}
```
clessig: It seems better to make sure that {} is passed to the function when strategy_kwargs is not set. This should do it:

`cf.get( "strategy_kwargs", {})`

Also please don't use NOTE in a comment ... that's the whole point of a code comment :)

src/weathergen/datasets/masking.py
```diff
@@ -54,6 +61,9 @@ def mask_source(
         token_lens = [len(t) for t in tokenized_data]
         num_tokens = sum(token_lens)
 
+        # print("Length of each token t in tokenized_data:", token_lens)
+        print("Number of tokens in the batch:", num_tokens)
```
clessig: Remove before we merge the PR

src/weathergen/datasets/masking.py
```diff
@@ -89,6 +99,10 @@ def mask_source(
             if block_size > 0 and num_tokens > 0:
                 start_index = self.rng.integers(0, max(1, num_tokens - block_size + 1))
                 flat_mask[start_index : start_index + block_size] = True
+
+        elif self.masking_strategy == "healpix":
+            flat_mask = self._generate_healpix_mask(token_lens, rate)
```
clessig, line 117: Then we should have a separate function for each of the masking strategies. My feeling is, we might want to implement in small classes at some point for generality but a separate function for every strategy seems like a good starting point.

src/weathergen/datasets/masking.py
```diff
@@ -146,3 +160,88 @@ def mask_target(
...
+
+        Args:
+            token_lens (list[int]): A list containing the number of tokens in each cell.
+            rate (float): The desired masking rate, applied to the parent cells.
+
+        Returns:
+            np.ndarray: A flat boolean array (the token-level mask).
+        """
+
+        print("Generating HEALPix mask...")
```
clessig: Remove before merging

src/weathergen/datasets/masking.py
```diff
@@ -146,3 +160,88 @@ def mask_target(
...
+
+        # NOTE: hl_data and hl_mask are expected to be provided in strategy_kwargs?
```
clessig: No NOTE

src/weathergen/datasets/masking.py
```diff
@@ -146,3 +160,88 @@ def mask_target(
...
+        hl_data = self.strategy_kwargs.get("hl_data")
+        hl_mask = self.strategy_kwargs.get("hl_mask")
```
clessig: We should fail as early as possible when hl_data and hl_mask are not set. Or can we fall back to default values in this case?

src/weathergen/datasets/masking.py
```diff
@@ -146,3 +160,88 @@ def mask_target(
...
+
+        print(
```
clessig: Remove

src/weathergen/datasets/masking.py
```diff
@@ -146,3 +160,88 @@ def mask_target(
...
+        num_parent_cells = 12 * (4**hl_mask)
+        level_diff = hl_data - hl_mask
+        num_children_per_parent = 4**level_diff
+
+        # print(f"[HEALPix Setup] Data Level (hl_data): {hl_data} ({num_data_cells} cells)")
+        # print(f"[HEALPix Setup] Mask Level (hl_mask): {hl_mask} ({num_parent_cells} parent cells)")
+        # print(f"[HEALPix Setup] Each parent cell at L{hl_mask} contains {num_children_per_parent} child cells at L{hl_data}.")
+
+        # Choose parent cells to mask based on the specified rate.
+        num_parents_to_mask = int(np.round(rate * num_parent_cells))
```
clessig, line 218: This should be sampled when masking_rate_sampling = True

src/weathergen/datasets/masking.py
```diff
@@ -146,3 +160,88 @@ def mask_target(
...
+            # print("[HEALPix Masking] Masking rate is too low. No parent cells were selected to be masked.")
+            return np.zeros(sum(token_lens), dtype=bool)
+
+        # Mask, and print about what we are doing.
+        parent_ids_to_mask = self.rng.choice(num_parent_cells, num_parents_to_mask, replace=False)
+        # print(f"[HEALPix Masking] Based on rate {rate:.2f}, selected {num_parents_to_mask}/{num_parent_cells} parent cells to mask.")
+        # print(f"[HEALPix Masking] Parent IDs selected: {parent_ids_to_mask}")
+
+        # Now determine which child cells (and their tokens) are masked.
+        # This is cells.
```
clessig: Unclear

src/weathergen/datasets/masking.py
```diff
@@ -146,3 +160,88 @@ def mask_target(
...
+        cell_mask = np.zeros(num_data_cells, dtype=bool)
+        # print("[HEALPix Masking] Mapping parent cells to child cell indices:")
+        for parent_id in parent_ids_to_mask:
```
clessig: Can this loop be vectorized?

src/weathergen/datasets/masking.py
shmh40: Using NOTE for myself! Will remove before merging

src/weathergen/datasets/masking.py
shmh40: @tjhunter is it ok to fall back to default values for a case like this?

src/weathergen/datasets/masking.py
shmh40: Yes, good point, done

src/weathergen/datasets/masking.py
shmh40: Removed

src/weathergen/datasets/masking.py
shmh40: Will raise another issue once this is merged

src/weathergen/datasets/masking.py
shmh40: Implemented this, I think it works neatly

clessig: True, didn't read carefully enough.

src/weathergen/datasets/masking.py
tjhunter: yes! This is a good place for default values.

src/weathergen/datasets/masking.py
```diff
@@ -156,3 +165,76 @@ def mask_target(
...
+
+        Returns:
+            np.ndarray: A flat boolean array (the token-level mask).
+        """
+
+        # hl_data and hl_mask should be provided in strategy_kwargs
+        hl_data = self.strategy_kwargs.get("hl_data")
+        hl_mask = self.strategy_kwargs.get("hl_mask")
+
+        if hl_data is None or hl_mask is None:
```
tjhunter: style: it is better to say: `assert hl_data is not None and hl_mask is not None, "If ..."`

src/weathergen/datasets/masking.py
```diff
@@ -156,3 +165,76 @@ def mask_target(
...
+            assert False, (
+                "If masking with HEALPix, hl_data and hl_mask must be provided in strategy_kwargs."
+            )
+
+        if hl_mask >= hl_data:
```
tjhunter: same here.

src/weathergen/datasets/masking.py
tjhunter: testing systems such as `pytest` can then do clever things to present to you the offending values if it fails.

src/weathergen/datasets/masking.py
```diff
@@ -156,3 +165,76 @@ def mask_target(
...
+            assert False, "hl_mask must be less than hl_data for HEALPix masking."
+
+        num_data_cells = 12 * (4**hl_data)
+        if len(token_lens) != num_data_cells:
```
tjhunter: same

src/weathergen/datasets/masking.py
```diff
@@ -156,3 +165,76 @@ def mask_target(
...
+            )
+
+        # Calculate the number of parent cells at the mask level (hl_mask)
+        num_parent_cells = 12 * (4**hl_mask)
+        level_diff = hl_data - hl_mask
+        num_children_per_parent = 4**level_diff
+
+        # if masking_rate_sampling is enabled, sample the rate from a normal distribution.
+        if self.masking_rate_sampling:
+            rate = np.clip(
```
tjhunter, line 211: this chunk is copied over. can you factorize them into a single class method? I am sure we will need to adjust this formula eventually, better to adjust it once.

config/default_config.yml
```diff
@@ -73,14 +73,15 @@ batch_size_validation: 1
...
-masking_rate: 0.6
+masking_rate: 0.5
 # sample the masking rate (with normal distribution centered at masking_rate)
 masking_rate_sampling: True
 # sample a subset of all target points, useful e.g. to reduce memory requirements
 sampling_rate_target: 1.0
 # include a masking strategy here, currently only supporting "random" and "block"
 masking_strategy: "random"
 
+
```
tjhunter: can you put a commented out example of the healpix strategy? Right now, we do not have documentation for what is expected in hl_data and hl_mask

src/weathergen/datasets/masking.py
tjhunter: personally, in this case, I would say `self.strategy_kwargs.hl_data` etc. rather than looking (yet) for a sensible default.

src/weathergen/datasets/masking.py
```diff
@@ -156,3 +165,76 @@ def mask_target(
...
+        This method identifies parent cells at a lower resolution (hl_mask) and
+        masks all the child cells (and their corresponding tokens) at the data
+        resolution (hl_data).
+
+        Args:
+            token_lens (list[int]): A list containing the number of tokens in each cell.
+            rate (float): The desired masking rate, applied to the parent cells.
+
+        Returns:
+            np.ndarray: A flat boolean array (the token-level mask).
```
tjhunter, line 192: what is the len() of this array? the indexing is unclear to me

src/weathergen/datasets/masking.py
```diff
@@ -25,11 +25,15 @@ def __init__(
...
         masking_strategy: str,
         masking_rate_sampling: bool,
+        strategy_kwargs: dict,
     ):
         self.masking_rate = masking_rate
         self.masking_strategy = masking_strategy
         self.masking_rate_sampling = masking_rate_sampling
 
+        # strategy_kwargs is a dictionary that can hold any additional parameters
+        self.strategy_kwargs = strategy_kwargs
```
tjhunter: I would call it strategy_config. `kwargs` is an implementation choice, and it will be later replaced by a class (likely)

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -182,7 +182,9 @@ def __init__(
         if cf.training_mode == "forecast":
             self.tokenizer = TokenizerForecast(cf.healpix_level, cf.data_loader_rng_seed)
         elif cf.training_mode == "masking":
-            masker = Masker(cf.masking_rate, cf.masking_strategy, cf.masking_rate_sampling)
+            masker = Masker(
+                cf.masking_rate, cf.masking_strategy, cf.masking_rate_sampling, cf.get("strategy_kwargs", {})
```
tjhunter: `masking_strategy_extra` instead of `strategy_kwargs` ? strategy is vague.

tjhunter (approved):
@shmh40  thanks! I have some style comments. 2 high level comments:

- it is missing how you would use it. when writing doc / updating the config, think how a newcomer would want to use this feature
- (not you): the `rng` in masking is depending on time. This makes the code non-deterministic. This issue to me is larger than biased training. I would love for us to just have a single seed for everything set in config

src/weathergen/datasets/masking.py
clessig: number of tokens

src/weathergen/datasets/masking.py
clessig: But can we test this in the constructor. This will be a much cleaner stack trace and it will fail much earlier

tjhunter:
> > * (not you): the `rng` in masking is depending on time. This makes the code non-deterministic. This issue to me is larger than biased training. I would love for us to just have a single seed for everything set in config
> 
> In my review I already made a suggestion how to fix it ;)

Yes! Somehow it did not appear in the history.

src/weathergen/datasets/masking.py
shmh40: This has been moved to be tested in the constructor, thanks.

I have not included default values at the moment, which I think is fine for now, we can include them later/in another PR if we like. I am not sure what our plan is overall with defaults, or @tjhunter the reasons for not having defaults. Did you write up a brief doc on your discussion with Sophie? Sorry if I missed it.

src/weathergen/datasets/masking.py
shmh40: Thank you for pointing these out

src/weathergen/datasets/masking.py
shmh40: There is an existing issue closely related to this, will do it at some point in a separate PR

config/default_config.yml
shmh40: Yes, good point, thank you! Added!

src/weathergen/datasets/multi_stream_data_sampler.py
shmh40: Changed name to masking_strategy_config

Issue #409, opened 2025-06-30
Integer overflow when computing indexes for FESOM dataset


kacpnowak:
### What happened?

Currently  in `data_reader_base.py` indexes are stored in `np.int32` which results in integer overlow when too large dataset is used. It's observed at `end_row = (t_idxs[-1] + 1) * self.mesh_size` inside `DataReaderFesom`

The solution is to use `np.int64` instead of `np.int32`

### What are the steps to reproduce the bug?

Run model with config including `streams_directory: "./config/streams/streams_ocean/"`

### Version

0.1

### Platform (OS and architecture)

Levante

### Relevant log output

```shell

```

### Organisation

AWI

Issue #350, opened 2025-06-15
Improve FESOM dataset


clessig:
### Is your feature request related to a problem? Please describe.

Some cleanup is needed in FESOM dataset

### Describe the solution you'd like

Cleanup in FESOM dataset

CC @kacpnowak

clessig:
> I'll take care of it. Are we keeping the support for atmorep format? IMO it could be usefull to have option to be able to load gridded data and I could do it alongside FESOM dataloader,

We have other data loaders for gridded data, so I do not see the use case. Could you give an example?

kacpnowak:
MITGCM and NEMO are using (kinda) regular grids. It might be usefull to keep them in native format, but I don't have any supporting arguments. It just something I can do alongside FESOM dataloader without much effort

PR #417, opened 2025-07-01
Fix FESOM datareader and int overflow 

kacpnowak:
## Description

Fixes issue with loading FESOM data indexing and integer overflow in datareader's base class

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/409 and https://github.com/ecmwf/WeatherGenerator/issues/350

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/datasets/tokenizer_forecast.py
+++ b/src/weathergen/datasets/tokenizer_forecast.py
@@ -263,7 +263,7 @@ class TokenizerForecast:
         target_times_raw = np.split(times_reordered, ll)
         target_times = np.split(times_reordered_enc, ll)
 
-        target_tokens_lens = torch.tensor([len(s) for s in target_tokens], dtype=torch.int64)
+        target_tokens_lens = torch.tensor([len(s) for s in target_tokens], dtype=torch.int32)
 
         # compute encoding of target coordinates used in prediction network
         if target_tokens_lens.sum() > 0:


src/weathergen/datasets/utils.py
```diff
@@ -448,12 +454,14 @@ def get_target_coords_local_ffast(
 
     # target_coords_lens = [len(t) for t in target_coords]
     tcs = [
-        s2tor3(
-            torch.deg2rad(90.0 - t[..., 0]),
-            torch.deg2rad(180.0 + t[..., 1]),
+        (
```
tjhunter, line 457: after all the work on having a formatter that works for everyone, I am still puzzled that it gets different results for you. let's not worry about it.

tjhunter (approved):

src/weathergen/datasets/data_reader_fesom.py
```diff
@@ -94,13 +94,17 @@ def __init__(
...
-        ) * self.mesh_size
-        self.end_idx = (
-            (tw_handler.t_end - start_ds).astype("timedelta64[D]").astype(int) + 1
-        ) * self.mesh_size - 1
+        period = self.time[self.mesh_size][0] - self.time[0][0]
+
+        if tw_handler.t_start > start_ds:
+            self.start_idx = ((tw_handler.t_start - start_ds) // period + 1) * self.mesh_size
+        else:
+            self.start_idx = 0
```
clessig, line 102: When a window before the start of the dataset is selected, does the FESOM data reader correctly return an empty window?

src/weathergen/datasets/stream_data.py
```diff
@@ -106,7 +106,7 @@ def add_empty_source(self) -> None:
         """
 
         self.source_raw += [torch.tensor([])]
-        self.source_tokens_lens += [torch.zeros([self.nhc_source], dtype=torch.int32)]
+        self.source_tokens_lens += [torch.zeros([self.nhc_source], dtype=torch.int64)]
```
clessig: This should remain int32, at least it cannot see that the int64 range could be necessary; it's definitely unrelated to the issue that is addressed in the PR

src/weathergen/datasets/stream_data.py
```diff
@@ -125,11 +125,11 @@ def add_empty_target(self, fstep: int) -> None:
         """
 
         # self.target_tokens[fstep] += [
-        #     [torch.zeros((0,0), dtype=torch.int32) for _ in range(self.nhc_target)]]
-        self.target_tokens[fstep] += [[torch.tensor([], dtype=torch.int32)]]
-        self.target_tokens_lens[fstep] += [torch.zeros([self.nhc_target], dtype=torch.int32)]
+        #     [torch.zeros((0,0), dtype=torch.int64) for _ in range(self.nhc_target)]]
```
clessig: Can we remove the commented out line

src/weathergen/datasets/stream_data.py
```diff
@@ -125,11 +125,11 @@ def add_empty_target(self, fstep: int) -> None:
...
+        self.target_tokens[fstep] += [[torch.tensor([], dtype=torch.int64)]]
```
clessig: Same as above: no int64 needed

src/weathergen/datasets/stream_data.py
```diff
@@ -125,11 +125,11 @@ def add_empty_target(self, fstep: int) -> None:
...
+        self.target_tokens_lens[fstep] += [torch.zeros([self.nhc_target], dtype=torch.int64)]
```
clessig: Same as above

src/weathergen/datasets/stream_data.py
```diff
@@ -125,11 +125,11 @@ def add_empty_target(self, fstep: int) -> None:
...
         self.target_coords[fstep] += [[torch.zeros((0, 106)) for _ in range(self.nhc_target)]]
-        self.target_coords_lens[fstep] += [torch.zeros([self.nhc_target], dtype=torch.int32)]
+        self.target_coords_lens[fstep] += [torch.zeros([self.nhc_target], dtype=torch.int64)]
```
clessig: Same as above

src/weathergen/datasets/tokenizer_forecast.py
```diff
@@ -177,7 +177,7 @@ def batchify_source(
 
         source_tokens_cells = torch.tensor([])
         source_centroids = torch.tensor([])
-        source_tokens_lens = torch.zeros([self.num_healpix_cells_source], dtype=torch.int32)
+        source_tokens_lens = torch.zeros([self.num_healpix_cells_source], dtype=torch.int64)
```
clessig: Same as above: cannot be out of int32 range

src/weathergen/datasets/tokenizer_forecast.py
```diff
@@ -194,7 +194,7 @@ def batchify_source(
         source_tokens_cells = [
             torch.stack(c) if len(c) > 0 else torch.tensor([]) for c in source_tokens_cells
         ]
-        source_tokens_lens = torch.tensor([len(s) for s in source_tokens_cells], dtype=torch.int32)
+        source_tokens_lens = torch.tensor([len(s) for s in source_tokens_cells], dtype=torch.int64)
```
clessig: Same as above: cannot be out of int32 range

src/weathergen/datasets/tokenizer_forecast.py
```diff
@@ -227,9 +227,9 @@ def batchify_target(
         time_win: tuple,
         normalizer,  # dataset
     ):
-        target_tokens = torch.zeros([self.num_healpix_cells_target], dtype=torch.int32)
-        target_coords = torch.zeros([self.num_healpix_cells_target], dtype=torch.int32)
-        target_tokens_lens = torch.zeros([self.num_healpix_cells_target], dtype=torch.int32)
+        target_tokens = torch.zeros([self.num_healpix_cells_target], dtype=torch.int64)
```
clessig: Same as above: cannot be out of int32 range

clessig (commented):
In the tokenizers, utils no change to int64 is necessary (I didn't mark everything).

Let's please not merge until this is fixed.

kacpnowak:
> In the tokenizers, utils no change to int64 is necessary (I didn't mark everything).
> 
> Let's please not merge until this is fixed.

It's not directly related, but mixing types of integers can lead to implicit casting to 32 bit and eventual overflow. This was the cause of the bug. Thus I think it can be a good idea to use int64 everywhere

clessig:
> > In the tokenizers, utils no change to int64 is necessary (I didn't mark everything).
> > Let's please not merge until this is fixed.
> 
> It's not directly related, but mixing types of integers can lead to implicit casting to 32 bit and eventual overflow. This was the cause of the bug. Thus I think it can be a good idea to use int64 everywhere.

Implicit casting can only happen if you have the same data and move it from one rep to another, which isn't the case here. Some of these tensors are used on the GPU, where 64-bit datatypes can have a huge performance penalty.

src/weathergen/datasets/data_reader_fesom.py
kacpnowak: Yes, this case is handled above

src/weathergen/datasets/utils.py
```diff
@@ -587,7 +595,7 @@ def compute_offsets_scatter_embed(batch: StreamData) -> StreamData:
                 )
                 s.source_idxs_embed_pe = torch.cat(
                     [
-                        torch.arange(offset, offset + token_len, dtype=torch.int32)
+                        torch.arange(offset, offset + token_len, dtype=torch.int64)
```
kacpnowak: @clessig Just double checking if here one tensor should be using int32 while the other uses int64

src/weathergen/datasets/tokenizer_forecast.py
```diff
@@ -263,7 +263,7 @@ def batchify_target(
         target_times_raw = np.split(times_reordered, ll)
         target_times = np.split(times_reordered_enc, ll)
 
-        target_tokens_lens = torch.tensor([len(s) for s in target_tokens], dtype=torch.int32)
+        target_tokens_lens = torch.tensor([len(s) for s in target_tokens], dtype=torch.int64)
```
clessig: Can we please also change this back.

src/weathergen/datasets/utils.py
clessig: 100%

clessig (approved):
Just one minor thing for consistency. Then we are ready to merge. Thanks for the work!

kacpnowak:
> Just one minor thing for consistency. Then we are ready to merge. Thanks for the work!

Done

Issue #430, opened 2025-07-02
key error in loss computation when training on obs datasets


iluise:
### What happened?

I get key errors when training on SEVIRI, NPP ATMS and Surface Combined due to a mismatch in the names. 
The code removes all spaces and special characters, but the stream names contain spaces and special characters like "Surface Combined" or "NPP, ATMS". 

I prepared a quick fix for the issue. opening the PR in a second. 

### What are the steps to reproduce the bug?

run with `streams_mixed` 

### Version

develop

### Platform (OS and architecture)

ATOS 

### Relevant log output

```shell

```

PR #431, opened 2025-07-02
fix obs dataset names in stream_mixed

iluise:
## Description

fix streams names for NPP ATMS, SEVIRI and SYNOP in steam_mixed to avoid key error not found crash. 

## Type of Change

-   [ *] Bug fix (non-breaking change which fixes an issue)
-   [ ] Documentation update

## Issue Number

related to this issue: https://github.com/ecmwf/WeatherGenerator/issues/430

## Code Compatibility

-   [* ] I have performed a self-review of my code

-   [ *] I ran the `uv run train` and (if necessary) `uv run evaluate` on a least one GPU node and it works 

### Dependencies

-   [ *] I have not introduced new dependencies in the inference portion of the pipeline

### DIFF
--- a/config/streams/streams_mixed/geos_1.yml
+++ b/config/streams/streams_mixed/geos_1.yml
@@ -7,7 +7,7 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-METEOSAT-X, SEVIRI :
+SEVIRI :
   type : obs
   filenames : ['observations-od-ai-0001-2018-2023-meteosat-11-seviri-v1.zarr']
   loss_weight : 1.0
--- a/config/streams/streams_mixed/npp_atms.yml
+++ b/config/streams/streams_mixed/npp_atms.yml
@@ -7,7 +7,7 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-NPP, ATMS :
+NPPATMS :
   type : obs
   filenames : ['observations-ea-ofb-0001-2012-2023-npp-atms-radiances-v2.zarr']
   loss_weight : 1.0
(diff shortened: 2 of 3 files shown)


config/streams/streams_mixed/geos_1.yml
```diff
@@ -7,7 +7,7 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-METEOSAT-X, SEVIRI :
+SEVIRI :
```
tjhunter, line 10: this change makes logging easier, but I thought these keys follow a specific convention (satellite, instrument). 
@clessig , what are your thoughts?

config/streams/streams_mixed/npp_atms.yml
```diff
@@ -7,7 +7,7 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-NPP, ATMS :
+NPPATMS :
```
tjhunter, line 10: If we do that, I would suggest either `NppAtms` (Java) or `NPP_ATMS` (python)

config/streams/streams_mixed/geos_1.yml
iluise: We should follow the STAC convention for naming, right?

config/streams/streams_mixed/npp_atms.yml
iluise: sorry, this should be NPP-ATMS according to the STAC convention. I will change it when I get the thumb up from both :)

config/streams/streams_mixed/geos_1.yml
tjhunter: yes, that would be best. I will follow the opinion of @clessig  on that naming matter though.

clessig:
The problem comes from #283.

clessig:
@kacpnowak : when we re-merge the channel PR we should also fix this.

PR #438, opened 2025-07-03
introducing pereceiver-io style module for embedding layer 

### LINKED ISSUE #225: embedding networks

Currently two modes for the embedding are implemented. More should be explored.

A thorough exploration of the design space for the embedding networks.

_No response_

### Additional context

_No response_

### Organisation

_No response_

csjfwang:
## Description
For channel-wise attention on ERA5, we currently use only 96 channels. As more channels are incorporated in the future, the computational burden will increase significantly. To address this, we introduce a Perceiver IO module before the channel-wise attention stage. 

## Type of Change

-   [ ] Documentation update

## Issue Number

Refs #225 

## Code Compatibility

### Dependencies

### Documentation

## Additional Notes
Fixes #225


src/weathergen/model/embeddings.py
```diff
@@ -52,6 +77,29 @@ def __init__(
         self.dim_embed = dim_embed
         self.dim_out = dim_out
         self.num_blocks = num_blocks
+        if cross_attn_params is not None:
```
sophie-xhonneux: For a couple reasons it would be better to have an explicit flag, use Perceiver in EmbeddingNetwork, than trying to check if the config exists. Looking back at logs it is easier to always check the same flag than to verify something is not present

src/weathergen/model/embeddings.py
```diff
@@ -52,6 +77,29 @@ def __init__(
...
+            assert mode == "channels", "cross-attention is only supported in channels mode"
+            
+            self.num_queries = cross_attn_params.get("num_queries", 0) if cross_attn_params else 0
+            self.selu = torch.nn.SELU(inplace=True)
+            if self.num_queries > 0:
+            # Multi-Cross Attention Head
+                num_channels = self.num_queries
+                self.cross_attn_num_blocks = cross_attn_params.get("num_blocks", 1) if cross_attn_params else 1
+                self.cross_attn_num_heads = cross_attn_params.get("num_heads", 1) if cross_attn_params else 1
+                self.queries = torch.nn.Parameter(
```
sophie-xhonneux: Pretty sure this code wasn't formatted

src/weathergen/model/embeddings.py
```diff
@@ -52,6 +77,29 @@ def __init__(
...
+                torch.randn(1, self.num_queries, self.dim_embed)
```
sophie-xhonneux: these parameters should be initialised to me much smaller than randn, please use an explicit init strategy! So for gaussian the standard deviation should be something like 0.01 to 0.0005, but really there should be an init_weights function that explicitly initialises all the weights of the function.

src/weathergen/model/embeddings.py
```diff
@@ -52,6 +77,29 @@ def __init__(
...
+                self.cross_attn_num_blocks = cross_attn_params.get("num_blocks", 1) if cross_attn_params else 1
+                self.cross_attn_num_heads = cross_attn_params.get("num_heads", 1) if cross_attn_params else 1
+                self.queries = torch.nn.Parameter(
+                torch.randn(1, self.num_queries, self.dim_embed) 
+            )
+                with torch.no_grad():
+                    self.queries.normal_(mean=0.0, std=1.0 / np.sqrt(self.dim_embed))
+                self.perceiver_io = PerceiverBlock(self.dim_embed, self.cross_attn_num_heads)
+            else:
+                self.perceiver_io = torch.nn.Identity()
```
sophie-xhonneux: shouldn't we also set `self.selu = torch.nn.Identity()` here? else there is a difference between num_queries = 0 and no cross attention parameters which is odd

src/weathergen/model/embeddings.py
```diff
@@ -145,7 +193,17 @@ def forward_channels(self, x_in, centroids):
         peh = positional_encoding_harmonic
 
         # embed provided input data
-        x = peh(checkpoint(self.embed, x_in.transpose(-2, -1), use_reentrant=False))
+        x = peh(self.selu(checkpoint(self.embed, x_in.transpose(-2, -1), use_reentrant=False)))
+
+
+        if self.num_queries > 0:
```
sophie-xhonneux: Please annotate with some batch dimension etc stuff, right now this is very hard to review/follow

Issue #271, opened 2025-05-27
Separate log files for model output streams


clessig:
### Is your feature request related to a problem? Please describe.

Currently, we only use stdout and stderr. This means that the model info/warning/error message are buried in messages from torch, NCCL, ...

### Describe the solution you'd like

The info, warning, err from the model should be written to separate files that are separate from stdout, stderr.

PR #439, opened 2025-07-03
Remove print statements for logging (#421)

### LINKED ISSUE #421: Change the print() statements to logger statements (except in the train loop)
The `print` statement does not propagate in the logging system. We should not have logging statement in the code except for `src/weathergen/model/model.py` (where it is used to track progress)

The goal of this task is to ask a linter rule and to audit the few places where we use a print statement. It should either be removed (old debugging) or turned into a _logger.info() statement.

The linter has a built-in rule called `T20`, which should be added in `pyproject.toml`:

```
...
    "TID",
    "T20"
]
```

See instructions in the wiki to try the linter locally, and here is how to disable the linter rule for specific usages:
https://github.com/astral-sh/ruff/discussions/3442

enssow:
## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #421 

Removes print statements from files except for `src/weathergen/model/model.py` and edits other print statements to be logged

## Code Compatibility

### Dependencies

### Documentation

## Additional Notes

- Failed for integration pytest:
- `uv run pytest ./integration_tests/small1_test.py --verbose ` which returned `E   ImportError: cannot import name 'inference_from_args' from 'weathergen' (unknown location)`


enssow:
@kacpnowak Is that in the `train/trainer.py`? Maybe @tjhunter could give a bit of insight on this one - I'm happy to change it back to print statements in that file if it messes with the current workflow for model training etc.

clessig:
> @kacpnowak Is that in the `train/trainer.py`? Maybe @tjhunter could give a bit of insight on this one - I'm happy to change it back to print statements in that file if it messes with the current workflow for model training etc.

Yes, we should keep the infos separate from warning, errors etc. There is also a PR to have a proper redirection of the different streams to different files to make sure things are visible: #271 . This would be the ultimate solution.

To which stream does __logger.info go?

scripts/check_gh_issue.py
```diff
@@ -13,9 +13,11 @@
 https://stackoverflow.com/questions/60717142/getting-linked-issues-and-projects-associated-with-a-pull-request-form-github-ap
 """
 
+import logging
```
tjhunter: sorry I was unclear. the scripts/ directory is special and has one-off scripts for automation. In general, it does not need to follow the stricter rules of the rest of the code. For example here, you did not initialize the logger so it would print nothing. Let's revert this change in `/scripts`

src/weathergen/train/trainer.py
```diff
@@ -682,14 +682,13 @@ def validate(self, epoch):
                     self.train_logger.add_val(samples, losses_all, stddev_all)
 
                 if self.cf.rank == 0:
-                    print(
```
tjhunter: let's keep this one too, it is relevant for the people working on the models.

src/weathergen/utils/run_id.py
```diff
@@ -6,8 +6,11 @@
...
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
+import logging
 
 from weathergen.train.utils import get_run_id
 
+_logger = logging.getLogger(__name__)
+
 if __name__ == "__main__":
-    print(get_run_id())
```
tjhunter, line 13: actually we can fully remove this main block. This is meant for people who want to create their own run ids.This is also implemented in the private repo, and it has no dependencies.

tjhunter (commented):
Thank you Sorcha for the change. Regarding the failing integration test: indeed, a fix had been reverted, let's do it separately.

Have you been able make a limited run with real data? I am a bit surprised by the comment of @kacpnowak  since before this PR, the print and logging statement are both redirected to the output_xxx.txt:

```
0: 023 : 01020/01024 : 024572 : loss = 2.9625E-02 (lr=9.77E-10, s/sec=0.746)
0: 	ERA5 : 2.9625E-02
0:
1: 2025-07-05 05:01:09,709 183795 trainer.py:321 : INFO     : Epoch 23 of 24: validate.
3: 2025-07-05 05:01:09,709 183797 trainer.py:321 : INFO     : Epoch 23 of 24: validate.
```

src/weathergen/utils/run_id.py
enssow: I can't seem to find any dependencies on this file so can I go ahead and delete the whole file?

scripts/check_gh_issue.py
```diff
@@ -1,3 +1,5 @@
+# ruff: noqa: T201
```
tjhunter: this breaks the script

Issue #364, opened 2025-06-18
Stream names are not friendly with logging systems (mlflow, wandb, etc.)


tjhunter:
### Is your feature request related to a problem? Please describe.

Currently, we allow descriptive stream names such as `NPP, ATMS` following a known convention in the satellite world. See [this example](https://github.com/ecmwf/WeatherGenerator/pull/328/files#diff-f893baba4f83facdfc557997e6bf4b389b927123e41035d4450fbb06e957b5f7). However, logging systems only want simple names for metrics such as `a.b.c`. where `a` is `a-zA-Z0-9-_` . It could be more but I would not push because it would break pandas column selection.

As a result, we cannot really have per-stream metrics, which would be useful for example to track how many samples are being ingested per stream. In this case it would be `streams.NPP, ATMS.count_samples` which would cause issues with WandB or MLFlow.

I see 3 ways forward:

- magic: conversion of `NPP, ATMS` to camelCase `NppAtms` for example
- convention: only allow names that follow the convention above and return error instead
- extra info: when defining a stream name, require a user to set an `stream_id` as a unique machine-ready identifier (for example `NPP_ATMS`)

In any case, I believe this name should be `a-zA-Z0-9_` (no _ to prevent weird issues with databases or pandas)

https://stackoverflow.com/questions/47964380/pandas-dataframe-column-naming-conventions

Do you have any thoughts on that?

clessig:
Can't we use "NPP, ATMS" as the user friendly identified and converted automagically to the machine friendly form (like NppAtms) that is typically hidden from the user?

tjhunter:
That sounds good to me. Explicit is always better than Implicit, but I do not have strong opinions for this naming convention.

clessig:
> That sounds good to me. Explicit is always better than Implicit, but I do not have strong opinions for this naming convention.

For the user format: for obs, "[PLATFORM, INSTRUMENT]" is standard. For the internal format, camel case with all offending chars parsed away sounds good.

clessig:
We should retain the human readable form (like "NPP, ATMS") if possible and additionally store the machine friendly version.

For the function documentation, can we use the agreed on style.

Sindhu-Vasireddy:
assign me

PR #445, opened 2025-07-04
Updated to camel case.

Sindhu-Vasireddy:
## Description

Made _clean_name() method in `train_logger.py` public and corrected the naming to camel case(as recommended by Julius).
 
## Type of Change

-   [ ] Documentation update

## Issue Number
 
#364 

## Code Compatibility

### Dependencies

### DIFF
--- a/src/weathergen/utils/train_logger.py
+++ b/src/weathergen/utils/train_logger.py
@@ -19,7 +19,7 @@ from typing import Literal
 
 import numpy as np
 import polars as pl
-
+import re
 import weathergen.utils.config as config
 from weathergen.utils.metrics import read_metrics_file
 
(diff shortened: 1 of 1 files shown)


Sindhu-Vasireddy:
Fixes #364

Sindhu-Vasireddy:
@tjhunter Could you please review the changes.

jpolz:
Thank you for the implementation. Is it sufficient to use the new name for logging or should there also be an entry in the config (or somewhere else) that gets updated?

The default regex may also be changed to allow `.`, `-` or `_`.

clessig:
> > There was so far no agreement that we do implement these changes, see the discussion in the issue (see here). Let's not merge it before we agreed on what should be done.
> 
> Where is "here"? I saw that there were multiple PRs or issues related to this. Would be good to link them.

https://github.com/ecmwf/WeatherGenerator/issues/364#issuecomment-2994030736

tjhunter:
@tjhunter  to test that the camelcase name only appears in metrics and does not impact logging etc.

clessig (changes requested):
Just flagging this as requiring a final review/discussion.

PR #490, opened 2025-07-10
Fix to the script running the integration test

tjhunter:
## Description

Found out when testing #439


clessig (approved):
I approve although it's not clear what "_test" refers to? Is it a test for small1.py? Or a renaming (but small1.py is still there).

tjhunter:
@enssow  @clessig thank you both for the review. 

@clessig  you must be looking at an old version of the code? 
https://github.com/ecmwf/WeatherGenerator/tree/develop/integration_tests

Issue #464, opened 2025-07-07
Inference always expect the target


ankitpatnala:
### Is your feature request related to a problem? Please describe.

For long roll-out to the decadal scales or to the end of century, there are no targets. Currently, the feature available uses inference which expects target always to compute loss. 
Maybe it is nice to simply have roll-out without targets. 

### Describe the solution you'd like

Either use `inference` function with an argument (`is_target=True`) or write a new function 'rollout`

### Organisation

JSC

ankitpatnala:
assign me

PR #495, opened 2025-07-10
Mlangguth/develop/issue 251

### LINKED ISSUE #251: Adapt and rework validation_io module to enable more efficient evaluation workflows

mlangguth89:
## Description

Introduce score-class and script for fast evaluation in evaluate-package. 
Furthermore, the a fix to `io.py` was made and the common dependencies have been updated.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #251

Addition to PR #365 

## Code Compatibility

I've ran `fast_evaluation.py` which uses a self-contained venv with `uv`. 

### Dependencies

For running `fast_evaluation`, dependencies in the `common`-package needed to be updated. Noteworthy is that the constraint `numcodecs<0.16.0` must be introduced to avoid spurious import errors with `zarr 2.18.4`.

### Documentation


packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+    Returns
+    -------
+    metric_stream : xr.DataArray
+        An xarray DataArray containing the computed metrics for the specified stream.
+    """
+    # Get stream-specific information
+    forecast_steps = zio.forecast_steps
+    nmetrics, nforecast_steps = len(metrics), len(forecast_steps)
+    samples = [int(sample) for sample in zio.samples]
+    channels_stream = peek_tar_channels(zio, stream)
```
clessig, line 83: Why are these not retrieved from the config? This would be easier and fast

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+_REPO_ROOT = Path(
+    __file__
+).parent.parent.parent.parent.parent.parent  # TODO use importlib for resources
+_DEFAULT_RESULT_PATH = _REPO_ROOT / "results"
+
+
+### Auxiliary functions
+def peek_tar_channels(zio: ZarrIO, stream: str):
+    """
+    Peek the channels of a tar stream in a ZarrIO object.
```
clessig: What is a tar stream?

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+    # filter channels if provided
+    channels_stream = (
+        [ch for ch in channels_stream if ch in to_list(channels)]
+        if channels
+        else channels_stream
+    )
+
+    # initialize the DataArray to store metrics
+    metric_stream = xr.DataArray(
+        np.full((nforecast_steps, len(channels_stream), nmetrics), np.nan),
```
clessig: This is fine for the moment but will break when nforecast_steps is large. The loop over forecast steps needs to be on the outsize and only one forecast step materialized at a time (all standard forecast scores are independent per step).

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+            out = zio.get_data(sample, stream, fstep)
+            targets.append(out.target.as_xarray().squeeze())
+            preds.append(out.prediction.as_xarray().squeeze())
+
+        # Concatenate targets and predictions along the 'ipoint' dimension and verify the data
+        _logger.debug(
+            f"Concatenating targets and predictions for stream {stream}, forecast_step {fstep}..."
+        )
+        targets_all, preds_all = (
+            xr.concat(targets, dim="ipoint"),
```
clessig, line 119: What is the ipoint dimension?

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+                                {
+                                    "metric": metric_idx,
+                                    "forecast_step": step_idx,
+                                    "channel": ch_idx,
+                                }
+                            ).values
+                        )
+                        result[stream][ch][step][metric] = value
+        else:
+            # If no channel dimension, data is per forecast_step and metric only
```
clessig: What is the canonical example here?

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+        f"Finished computing metric scores for all streams. Total streams processed: {len(all_metric_streams)}"
+    )
+
+    # Convert the list of metric streams to a nested dictionary structure
+    _logger.debug("Converting metric streams to dictionary format...")
+    metric_dict = metric_list_to_dict(all_metric_streams, streams)
+
+    # Save the results to a JSON file
+    save_dir.mkdir(parents=True, exist_ok=True)
+    save_path = save_dir / f"metrics_{run_id}_epoch{epoch:05d}_rank{rank:04d}.json"
```
clessig: This should not have a rank dimension. If multiple ranks are used for running inference, this should be combined in here.

clessig: We don't typically do this at the moment. Let's remove the rank part and later adapt the class to process output from multiple ranks.

packages/evaluate/src/weathergen/evaluate/score.py
```diff
@@ -0,0 +1,912 @@
...
+try:
+    import xskillscore
+    from xhistogram.xarray import histogram
+except Exception:
+    _logger.warning(
+        "Could not import xskillscore and xhistogram. Thus, CRPS and "
+        + "rank histogram-calculations are not supported."
+    )
+
+def to_list(obj: Any) -> list:
```
clessig: This function is also in fast_evaluation.py. Please put into a utils.py

packages/evaluate/src/weathergen/evaluate/score.py
```diff
@@ -0,0 +1,912 @@
...
+        Skill score data array
+    """
+
+    skill_score = (score_fcst - score_ref) / (score_perf - score_ref)
+
+    return skill_score
+
+
+@dataclass(frozen=True)
+class VerifiedData:
```
clessig, line 63: When or why could this fail?

clessig (changes requested):
Just some minor things. But would like to test it before it's merged.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+from pathlib import Path
+from typing import Any
+
+import numpy as np
+import xarray as xr
+
+from weathergen.common.io import ZarrIO
+from weathergen.evaluate.score import VerifiedData, get_score
+
+handler = logging.StreamHandler(sys.stdout)
```
tjhunter: never put the handler or setLevel in a file that is meant to be imported. I can go over in more details, but this can trigger all sorts of funny bugs in your logger system. It should always be:

```py

# the main entry point:
if __name__ == "__main__":
   # set up the handler, formatter, etc.
   # but you should look at basicLogging, this is really all you need here
```

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+_logger.propagate = False  # Prevent double logging if root logger also logs
+
+_REPO_ROOT = Path(
+    __file__
+).parent.parent.parent.parent.parent.parent  # TODO use importlib for resources
+_DEFAULT_RESULT_PATH = _REPO_ROOT / "results"
+
+
+### Auxiliary functions
+def peek_tar_channels(zio: ZarrIO, stream: str):
```
tjhunter: `-> list[str]`

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+        raise TypeError("zio must be an instance of ZarrIO")
+
+    dummy_out = zio.get_data(0, stream, 0)
+    channels = dummy_out.target.channels
+
+    return channels
+
+
+def calc_scores_per_stream(
+    zio: ZarrIO, stream: str, metrics: list[str], channels: str | list[str] = None
```
tjhunter: note type annotation is wrong, it should be `channels: str | list[str] | None = None`

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+    ----------
+    zio : ZarrIO
+        The ZarrIO object containing the data.
+    stream : str
+        The name of the stream to process.
+    metrics : list[str]
+        A list of metric names to calculate.
+    Returns
+    -------
+    metric_stream : xr.DataArray
```
tjhunter, line 75: return type in the function signature

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+def calc_scores_per_stream(
+    zio: ZarrIO, stream: str, metrics: list[str], channels: str | list[str] = None
+):
+    """
+    Calculate the provided score metrics for a specific.
+    Parameters
+    ----------
+    zio : ZarrIO
```
tjhunter, line 67: you don't need to put the types in the doc. python's doc system will retrieve them from the function signature for you

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+        A list of metric names to calculate.
+    Returns
+    -------
+    metric_stream : xr.DataArray
+        An xarray DataArray containing the computed metrics for the specified stream.
+    """
+    # Get stream-specific information
+    forecast_steps = zio.forecast_steps
+    nmetrics, nforecast_steps = len(metrics), len(forecast_steps)
+    samples = [int(sample) for sample in zio.samples]
```
tjhunter, line 82: a bit surprised we need to make a conversation, but ok

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
tjhunter: I would leave it as a todo for now. the config reading part is still in the main code, not in the `common` package

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+        A list containing the object, or the object itself if it was already a list.
+    """
+    if isinstance(obj, set | tuple):
+        obj = list(obj)
+    elif not isinstance(obj, list):
+        obj = [obj]
+    return obj
+
+
+def metric_list_to_dict(metric_list: list[xr.DataArray], streams: list):
```
tjhunter: `streams: list[str] ) -> dict[str,dict[str,dict[int,dict[str, float]]]]` ?

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+    """
+    Convert a list of xarray DataArrays containing metrics into a nested dictionary structure.
+
+    Parameters
+    ----------
+    metric_list : list[xr.DataArray]
```
tjhunter, line 150: it would be simpler to combine both arguments in a `dict[str, xr.DataArray]` . python dictionaries keep the order of insertion if you need it.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+
+    return result
+
+
+def fast_evaluation(
+    run_id: str,
+    metrics: list[str],
+    save_dir: Path,
+    results_dir: Path = _DEFAULT_RESULT_PATH,
+    streams: str | list[str] = None,
```
tjhunter: same comment: `| None`

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+    _logger.debug("Converting metric streams to dictionary format...")
+    metric_dict = metric_list_to_dict(all_metric_streams, streams)
+
+    # Save the results to a JSON file
+    save_dir.mkdir(parents=True, exist_ok=True)
+    save_path = save_dir / f"metrics_{run_id}_epoch{epoch:05d}_rank{rank:04d}.json"
+
+    _logger.info(f"Saving results to {save_path}")
+    with open(save_path, "w") as f:
+        json.dump(metric_dict, f, indent=4)
```
tjhunter, line 260: thank you for making it readable by a human!

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+    result = {}
+
+    assert len(metric_list) == len(streams), (
+        "Inconsistent list of metrics and streams passed."
+    )
+
+    for istream, da_metric in enumerate(metric_list):
+        # Get the stream name (e.g., 'ERA5', 'IMERG', etc.)
+        stream = streams[istream]
+        if stream not in result:
```
tjhunter: you can make your life much easier with the collections.defaultdict: 
https://stackoverflow.com/questions/41455065/understanding-the-use-of-defaultdict-in-python

```py
results:  dict[str,dict[str,dict[int,dict[str, float]]]] = defaultdict(defaultdict(defaultdict(defaultdict))))
result[stream][ch][step][metric] = value # No need to initialize the sub-dicts
```

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+            result[stream] = {}
+
+        # Check if 'channel' exists (not all arrays have it)
+        has_channel = "channel" in da_metric.dims
+
+        metrics = da_metric.coords["metric"].values
+        forecast_steps = to_list(da_metric.coords["forecast_step"].values)
```
tjhunter: check the content, you are getting lists of lists right now: `[["0"]]`

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,364 @@
...
+        json.dump(metric_dict, f, indent=4)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Fast evaluation of weather generator runs."
+    )
+
+    parser.add_argument(
+        "-id", "--run_id", type=str, help="The ID of the run to evaluate."
```
tjhunter: We have `--run-id` in the rest of the code. I know it is different from `--from_run_id` but it is better to be consistent. We will later build that automatically.

packages/evaluate/src/weathergen/evaluate/score.py
```diff
@@ -0,0 +1,912 @@
...
+import numpy as np
+import pandas as pd
+import xarray as xr
+
+# from common.io import MockIO
+
+_logger = logging.getLogger(__name__)
+
+try:
+    import xskillscore
```
tjhunter, line 25: why not include it by default in the dependencies for now?

packages/evaluate/src/weathergen/evaluate/score.py
tjhunter: maybe a small piece of documentation would be helpful here. When you encounter a `vd: VerifiedData` , you can always assume the dimensions have been validated to match.

packages/evaluate/src/weathergen/evaluate/score.py
```diff
@@ -0,0 +1,912 @@
...
+        # Call lazy evaluation function
+        result = f(**args)
+
+        if compute:
+            return result.compute()
+        else:
+            # Return lazy evaluation result
+            return result
+
+    def _validate_agg_dims(self, dims):
```
tjhunter: -> list[str]

packages/evaluate/src/weathergen/evaluate/score.py
```diff
@@ -0,0 +1,912 @@
...
+
+# scores class
+class Scores:
+    """
+    Class to calculate scores and skill scores.
+    """
+
+    def __init__(
+        self,
+        agg_dims: str | list[str] = "all",
```
tjhunter, line 146: rather than having a magic value "all", I would say `str | list[str] | None = None` where None means all. No big preference, you decide.

packages/evaluate/src/weathergen/evaluate/score.py
```diff
@@ -0,0 +1,912 @@
...
+        Note: this version is found to be very slow. Use calc_rank_histogram alternatively.
+        """
+        rank_hist = xskillscore.rank_histogram(
+            gt, p, member_dim=self.ens_dim, dim=self._agg_dims
+        )
+
+        return rank_hist
+
+    @staticmethod
+    def calc_geo_spatial_diff(
```
tjhunter, line 836: why static here?

tjhunter (commented):
@mlangguth89 this looks great! Most of my code comments are minor. I have 2 high level issues:

the output json is wrong:

```
 cat test/metrics_bo932pyc_epoch00000_rank0000.json
{
    "ERA5": {
        "q_50": {
            "['0']": {. <--- see here
                "rmse": 4.2673315192587324e-07,
                "mae": 3.703972311086545e-07,
                "bias": -2.627452033721056e-07
            }
        },
        "q_100": {
            "['0']": {
```

- your code does not suppor a different order / subset of target variables. It was the same problem for the io code. I will send you a way to reproduce it.

Otherwise, I think we should be able to merge it realy quickly.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: @clessig I indeed had the same thought, but also put it aside for now.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: Ups, the doc-string is incomplete. It should be target stream.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: I don't think that `nforecast_steps` will ever be this larger to result into serious memory constraints (>O10**3?). 
In case, one could initailze it with a dask-array.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: That's the number of data points per sample, cf. Simon's interface code: https://github.com/ecmwf/WeatherGenerator/blob/f75388bfa97aaca9dc4642c58a90f32d77b30af4/src/weathergen/utils/io.py#L102

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: It used to happen with the IMERG-data (one channel dataset) as a result of a `squeeze`-application. Need to check if this can still happen at all.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: That's how the zarr storage path is currently formatted. So, it's not related to the code of this PR. Please open a separate issue for this.

mlangguth89:
> @mlangguth89 this looks great! Most of my code comments are minor. I have 2 high level issues:
> 
> the output json is wrong:
> 
> ```
>  cat test/metrics_bo932pyc_epoch00000_rank0000.json
> {
>     "ERA5": {
>         "q_50": {
>             "['0']": {. <--- see here
>                 "rmse": 4.2673315192587324e-07,
>                 "mae": 3.703972311086545e-07,
>                 "bias": -2.627452033721056e-07
>             }
>         },
>         "q_100": {
>             "['0']": {
> ```
> 
>     * your code does not suppor a different order / subset of target variables. It was the same problem for the io code. I will send you a way to reproduce it.
> 
> 
> Otherwise, I think we should be able to merge it realy quickly.

I didn#t fully get your note on the JSON-file. Can you further explain?

packages/evaluate/src/weathergen/evaluate/score.py
mlangguth89: No information from the scoring class is required.
It may change in the future (e.g. a grid-identifier, since it only works this way for regular grids).
I've not further worked on this yet, and would like to tackle it in the future.

packages/evaluate/src/weathergen/evaluate/score.py
mlangguth89: Yes, would be helpful I think -> provides functionality out-of-the-box.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: From the current xarray-interface, one receives a string (-> from already merged code).
But yes, this should probably be changed.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
clessig: See #500

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
clessig: For climate roll out it will be thousands or hundreds of thousands of steps. For climate runs some variables are stored at 12h or 24h intervals. For a 100 year pi-control one then has 365 x 100 = 36,500 steps. For various things one also needs 10,000 year runs.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
clessig: We should have 1-dim channel there. That's much cleaner and avoids the special case handling.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
clessig: That's not the load, it's the save path. Yes, it will be in the load path but it shouldn't be for saving: the scoring should consolidate across ranks.

clessig:
> > @mlangguth89 this looks great! Most of my code comments are minor. I have 2 high level issues:
> > the output json is wrong:
> > ```
> >  cat test/metrics_bo932pyc_epoch00000_rank0000.json
> > {
> >     "ERA5": {
> >         "q_50": {
> >             "['0']": {. <--- see here
> >                 "rmse": 4.2673315192587324e-07,
> >                 "mae": 3.703972311086545e-07,
> >                 "bias": -2.627452033721056e-07
> >             }
> >         },
> >         "q_100": {
> >             "['0']": {
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > * your code does not suppor a different order / subset of target variables. It was the same problem for the io code. I will send you a way to reproduce it.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Otherwise, I think we should be able to merge it realy quickly.
> 
> I didn't fully get your note on the JSON-file. Can you please further explain?

There's a "." after "{" that shouldn't be there.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: That's true. Thanks for the foresighted view - I will give it a try with a dask-array.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: I think, that's what I have right now.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: Alright!

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: Implementing this, I came to the conclusion that we should not go ahead with this at the current stage.

Bear in mind that `metric_stream` must be converted to a dict and serialized in a JSON-file later in the current code. This requires currently to have all data to be loaded into memory. Realizing a fully lazy approach to handle the 'extreme' situations with a very, very large large number of forecasts_steps introduces therefore challenges in the downstream processng, and I would therefore like to skip this aside for now. 
a) Because we are not doing this right now and it will not be the case for fast_evaluation in  the next few months
b) We agreed at some point that evaluation of climate runs should be performed with community tools.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: No, this function just supports processing a single stream (as needed in `calc_scores_per_stream`).

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: Personally, I find this more explicit and simple.
Otherwise, one would need to put both into a dictionary when calling the function, and one would need to unpack it inside the function as well.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: @tjhunter So what should be changed here?

mlangguth89:
> > > @mlangguth89 this looks great! Most of my code comments are minor. I have 2 high level issues:
> > > the output json is wrong:
> > > ```
> > >  cat test/metrics_bo932pyc_epoch00000_rank0000.json
> > > {
> > >     "ERA5": {
> > >         "q_50": {
> > >             "['0']": {. <--- see here
> > >                 "rmse": 4.2673315192587324e-07,
> > >                 "mae": 3.703972311086545e-07,
> > >                 "bias": -2.627452033721056e-07
> > >             }
> > >         },
> > >         "q_100": {
> > >             "['0']": {
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ```
> > > * your code does not suppor a different order / subset of target variables. It was the same problem for the io code. I will send you a way to reproduce it.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Otherwise, I think we should be able to merge it realy quickly.
> > 
> > 
> > I didn't fully get your note on the JSON-file. Can you please further explain?
> 
> There's a "." after "{" that shouldn't be there.

And I think it shouldn't be a list here ;-)

packages/evaluate/src/weathergen/evaluate/score.py
mlangguth89: As discussed earlier, I prefer a bit having 'all' ;-)

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
mlangguth89: I've fixed this

packages/evaluate/src/weathergen/evaluate/score.py
mlangguth89: It turns out that importing these packages takes quite a long time due to raised Syntax-warnings.
I'll include it however for now.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,332 @@
...
+
+        # Store the computed metrics in the DataArray and does computation
+        metric_stream.loc[{"forecast_step": fstep}] = combined_metrics.compute()
+        _logger.info(f"Computed metrics for forecast_step {fstep} of stream {stream}.")
+
+    return metric_stream
+
+
+def metric_list_to_dict(
+    metric_list: list[xr.DataArray], streams: list
```
tjhunter: list[str]

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
tjhunter: `--run_id` -> `--run-id`

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -0,0 +1,332 @@
...
+        "-m",
+        "--metrics",
+        type=str,
+        nargs="+",
+        default=["rmse", "mae", "bias"],
+        help="List of metrics to evaluate.",
+    )
+    parser.add_argument(
+        "-sd",
+        "--save_dir",
```
tjhunter: can we have `-` instead of `_` ? 
- python will do the conversion for you for the python names
- `-` is shorter to type
- `-` is the default in unix systems

src/weathergen/utils/validation_io.py
```diff
@@ -46,7 +46,7 @@ def write_output(
 
     # assume: is batch size guarnteed and constant:
     # => calculate global sample indices for this batch by offsetting by sample_start
-    sample_start = batch_idx * cf.batch_size_validation
+    sample_start = batch_idx * cf.batch_size_validation_per_gpu
```
tjhunter, line 49: thanks!

tjhunter (approved):

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
clessig: #464 works on supporting arbitrary long roll-outs right now

Issue #396, opened 2025-06-26
Parameterise mask_rate_sampling


shmh40:
### Is your feature request related to a problem? Please describe.

masking_rate_sampling currently samples masking rates according to a prescribed mean and standard Gaussian. This could be extended to different distributions/parameterised.

We should also clip so we do not have 0 and 1 values.

### Describe the solution you'd like

Parameterise the Gaussian currently used to sample values for masking_rate if masking_rate_sampling is True.

PR #496, opened 2025-07-10
Implement global and per-cell channel masking

### LINKED ISSUE #458: Per-channel masking

Add ability to do per-channel masking as a masking strategy during training.

Implement per-channel masking where different channels can be masked out, using the new masking class.

_No response_

### Additional context

_No response_

### Organisation

_No response_

shmh40:
## Description

This draft PR implements global and per-cell channel masking, allowing the choice of these along with standard masking_rate, masking rate sampling and so on. 

Currently this does not support per channel masking with different numbers, or different choices of source and target channels. The source and target channels for per channel masking must be the same. This will be addressed in a subsequent PR.

The code builds on some of the small changes in shmh40/dev/masking_strategies which implements masking at arbitrary healpix levels.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #458 

## Code Compatibility

### Dependencies

### Documentation


config/default_config.yml
```diff
@@ -23,8 +23,8 @@ ae_adapter_with_qk_lnorm: True
 ae_adapter_with_residual: True
 ae_adapter_dropout_rate: 0.1
 
-ae_global_dim_embed: 2048
-ae_global_num_blocks: 8
+ae_global_dim_embed: 1024
```
sophie-xhonneux: how come this change?

src/weathergen/datasets/masking.py
```diff
@@ -47,6 +47,13 @@ def __init__(
             )
             assert hl_mask < hl_data, "hl_mask must be less than hl_data for HEALPix masking."
 
+        if self.masking_strategy == "channel":
```
sophie-xhonneux, line 66: I would just add the self.masking_strategy == channel into the assert cleaner that way

src/weathergen/datasets/masking.py
```diff
@@ -155,18 +182,38 @@ def mask_target(
...
+
         for cc, pp in zip(target_tokenized_data, self.perm_sel, strict=True):
-            selected_tensors = [c for c, p in zip(cc, pp, strict=True) if p]
+            if self.masking_strategy == "channel":
+                # If the masking strategy is channel,
+                # we need to handle the target tokens differently.
+                # Since we don't have true or false on a per cell basis,
+                # instead per channel for each cell,
+                # we set the not masked channels to NaN so
+                # they are not used in the loss calculation.
```
sophie-xhonneux: reformat the comment to take fewer lines maybe

src/weathergen/datasets/masking.py
```diff
@@ -155,18 +182,38 @@ def mask_target(
...
+
+                selected_tensors = []
+                for c, p in zip(cc, pp, strict=True):
+                    
+                    # assert that c has the same shape at dimension 1 as p
+                    assert c.shape[-1] == p.shape[-1], (
```
sophie-xhonneux: can you just confirm that this does not cause a huge performance hit, because that assrt will get checked very often

src/weathergen/datasets/masking.py
```diff
@@ -155,18 +182,38 @@ def mask_target(
...
+                        f"The source and target channels must be the same."
+                    )
+
+                    # slightly complicated as the first dimension of c varies with data in the cell.
+                    c[:, ~p[0, :]] = torch.nan  # Set the channels that are not masked to NaN
```
sophie-xhonneux: why nan? can't this cause all sorts of issues further than the line?

src/weathergen/datasets/masking.py
```diff
@@ -239,3 +286,93 @@ def _generate_healpix_mask(self, token_lens: list[int], rate: float) -> np.typin
...
+
+        # If all tensors in the batch are empty, return masks matching their shapes.
+        if num_channels is None:
+            return [np.empty(t.shape, dtype=bool) for t in tokenized_data]
+
+        # masking rate sampling, to be refactored as shared between methods
+        if self.masking_rate_sampling:
+            rate = np.clip(
+                np.abs(self.rng.normal(loc=rate, scale=1.0 / (2.5 * np.pi))),
+                0.0,
```
sophie-xhonneux: do we want to allow 0 and 1? can't these cause bugs?

src/weathergen/datasets/masking.py
```diff
@@ -239,3 +286,93 @@ def _generate_healpix_mask(self, token_lens: list[int], rate: float) -> np.typin
...
+        num_fixed_channels = 6 + coords.shape[-1] + geoinfos.shape[-1]
+        num_data_channels = num_channels - num_fixed_channels
+        mask_count = int(num_data_channels * rate)
+
+        # generate masks for each cell...TODO: optimise this
+        full_mask = []
+        is_per_cell = self.masking_strategy_config.get("per_cell", False)
+        base_channel_mask_1d = None
+
+        for cell_data in tokenized_data:
```
sophie-xhonneux: this should be easily parallelisable no?

sophie-xhonneux (commented):
Didn't run the code and some small computation time concerns

src/weathergen/datasets/masking.py
shmh40: Thanks, done

src/weathergen/datasets/masking.py
shmh40: Agreed, removed

src/weathergen/datasets/masking.py
shmh40: Doing it for now, may be changed in future

src/weathergen/datasets/masking.py
shmh40: This should be handled maybe in the constructor, I will open a separate issue, thank you for noting!

src/weathergen/datasets/masking.py
shmh40: It is actually tricky I think because of the variable shape tokens...will think about this more if/when this is a bottleneck

src/weathergen/datasets/masking.py
sophie-xhonneux: as in the constructor has a min and max value?

config/default_config.yml
shmh40: Mistake from an older merge - thank you for spotting!

src/weathergen/datasets/masking.py
shmh40: Like it is now?

src/weathergen/datasets/masking.py
shmh40: Possibly yes, or raise an error straight away if masking_rate == 0 or 1 maybe?

shmh40: assert masking_rate != 0.0 or 1.0 "You need some masking but not too much!"

sophie-xhonneux: not sure need to think about it with a clear mind tomorrow

src/weathergen/datasets/masking.py
sophie-xhonneux: yes I think, would need to check the boolean logic is still the same

src/weathergen/datasets/masking.py
sophie-xhonneux: I think the problem is that our code cannot handle 0 or 1, so we should just prevent it from ever happening by clipping with some epsilon

shmh40: Ok, with a warning so the user knows they are not actually getting the masking rate they asked for? Might it be better to fail with a clear error so the user knows the code doesn't support this, and then they can change the masking rate themselves?

sophie-xhonneux: I would clip at eps and 1-eps, and write in the docs or function and the config that the code cannot handle masking rate 0 and 1 and thus will always be clipped at eps and 1-eps

sophie-xhonneux: Too many warnings people will just ignore them

shmh40: Thank you, this was needed! I have clipped at 0.01 and 0.99, still not perfect but there is another issue where this will be addressed: #396

src/weathergen/datasets/masking.py
shmh40: Left it for readability

clessig: Directly having asserts helps automatic testing. You can do two different asserts if it helps readability.

src/weathergen/datasets/masking.py
```diff
@@ -116,9 +127,25 @@ def mask_source(
         elif self.masking_strategy == "healpix":
             flat_mask = self._generate_healpix_mask(token_lens, rate)
 
+        elif self.masking_strategy == "channel":
+            mask = self._generate_channel_mask(tokenized_data, rate, coords, geoinfos)
+
         else:
             assert False, f"Unknown masking strategy: {self.masking_strategy}"
 
+        # if masking_strategy is channel,
```
clessig: Reduce comment to half the lines

src/weathergen/datasets/masking.py
```diff
@@ -239,3 +278,93 @@ def _generate_healpix_mask(self, token_lens: list[int], rate: float) -> np.typin
...
+            list[np.ndarray]: A list of boolean masks. Each mask corresponds to a tensor
+                            in tokenized_data.
+        """
+
+        if not tokenized_data:
+            return []
+
+        # determine the number of channels
+        num_channels = None
+        for t in tokenized_data:
```
clessig: Leave a comment why the loop is needed

src/weathergen/datasets/masking.py
```diff
@@ -239,3 +278,93 @@ def _generate_healpix_mask(self, token_lens: list[int], rate: float) -> np.typin
...
+            if t.numel() > 0:
+                num_channels = t.shape[2]
```
clessig: Should be t.shape[-1] (we might have ensemble dimension eventually)

src/weathergen/datasets/masking.py
```diff
@@ -239,3 +278,93 @@ def _generate_healpix_mask(self, token_lens: list[int], rate: float) -> np.typin
...
+                break
+
+        # If all tensors in the batch are empty, return masks matching their shapes.
+        if num_channels is None:
```
clessig: This should be handled before the function is called.

src/weathergen/datasets/masking.py
```diff
@@ -239,3 +278,93 @@ def _generate_healpix_mask(self, token_lens: list[int], rate: float) -> np.typin
...
+        num_fixed_channels = 6 + coords.shape[-1] + geoinfos.shape[-1]
+        num_data_channels = num_channels - num_fixed_channels
+        mask_count = int(num_data_channels * rate)
+
+        # generate masks for each cell...TODO: optimise this
+        full_mask = []
+        is_per_cell = self.masking_strategy_config.get("per_cell", False)
+        base_channel_mask_1d = None
+
+        for cell_data in tokenized_data:
```
clessig: How slow is this?

src/weathergen/model/model.py
```diff
@@ -448,7 +448,9 @@ def embed_cells(self, model_params, streams_data):
                     # create full scatter index
                     # (there's no broadcasting which is likely highly inefficient)
                     idxs = idxs.unsqueeze(1).repeat((1, self.cf.ae_local_dim_embed))
+
```
clessig: Can we remove this before we merge.

config/default_config.yml
```diff
@@ -82,8 +82,11 @@ masking_rate: 0.6
 masking_rate_sampling: True
 # sample a subset of all target points, useful e.g. to reduce memory requirements
 sampling_rate_target: 1.0
-# include a masking strategy here, currently only supporting "random" and "block"
+# include a masking strategy here, currently only supporting "random", "block", "healpix" and "channel"
 masking_strategy: "random"
+# masking_strategy_config is a dictionary of additional parameters for the masking strategy, here an example for "channel" masking
+# "healpix": e.g. requires hl_data and hl_mask to be specified, with hl_data the hl level of the data and hl_mask for the mask
```
clessig: We need well documented examples here for the valid combinations.

config/default_config.yml
```diff
@@ -130,4 +133,4 @@ run_id: ???
 # Parameters for logging/printing in the training loop
 train_log:
   # The period to log metrics (in number of batch steps)
-  log_interval: 20
+  log_interval: 20
```
clessig, line 139: Who can spot the difference?

src/weathergen/datasets/masking.py
```diff
@@ -13,26 +13,47 @@ class Masker:
     Attributes:
         masking_rate (float): The base rate at which tokens are masked.
         masking_strategy (str): The strategy used for masking (e.g., "random",
-        "block", MORE TO BE IMPLEMENTED...).
+        "block", "healpix", "channel").
```
clessig, line 18: Let's document the supported once, and add to the list if we include new ones.

shmh40: Documented now in the docstring

config/default_config.yml
shmh40: Done

Issue #498, opened 2025-07-10
Extend string format in str_to_timedelta to days to allow for daily frequency in streaminfo


jpolz:
### Is your feature request related to a problem? Please describe.

The [current function for converting frequency information from the streaminfo](https://github.com/ecmwf/WeatherGenerator/blob/15c8c9208d6b33c3f499b51fce66ed1b33660d40/src/weathergen/datasets/data_reader_base.py#L77C1-L84C17) allows the format `"%H:%M:%S"` which is insufficient for a daily frequency needed for e.g. 24h or 48h forecasting

### Describe the solution you'd like

I would like to extend this to `"%d-%H:%M:%S"` or similar to be able to pass a frequency of one or more days. The [timedelta is converted to microseconds](https://github.com/ecmwf/WeatherGenerator/blob/15c8c9208d6b33c3f499b51fce66ed1b33660d40/src/weathergen/datasets/data_reader_anemoi.py#L78) anyways so it does not make a large difference.

jpolz:
assign me

clessig:
Why is 48:00:00 an invalid format? Can’t we fix it by parsing differently?

________________________________

jpolz:
> Why is 48:00:00 an invalid format? Can’t we fix it by parsing differently?

strptime does not accept it.
```
1:   File ".../lib/python3.12/_strptime.py", line 434, in _strptime
1:     raise ValueError("time data %r does not match format %r" %
1: ValueError: time data '24:00:00' does not match format '%H:%M:%S'
```
I have no overview of potential issues with adding days. Maybe you have some cases in mind that would cause issues. It works for 24 hours though. I already tested it.

PR #499, opened 2025-07-10
extend format string and timedelta to days

jpolz:
## Description

extend format string and timedelta parsing in str_to_timedelta to days

## Type of Change

## Issue Number
Closes https://github.com/ecmwf/WeatherGenerator/issues/498

## Code Compatibility

### Dependencies

no new dependencies

### Documentation


src/weathergen/datasets/data_reader_base.py
```diff
@@ -77,15 +77,15 @@ def str_to_datetime64(s: str | int | NPDT64) -> NPDT64:
 def str_to_timedelta(s: str | datetime.timedelta) -> datetime.timedelta:
     """
     Convert a string to a timedelta object.
-    The format is expected to be "HH:MM:SS".
+    The format is expected to be "D-HH:MM:SS".
```
clessig: Why is 48:00:00 invalid?

jpolz: strptime: `ValueError: time data '24:00:00' does not match format '%H:%M:%S'`

src/weathergen/datasets/data_reader_base.py
```diff
@@ -77,15 +77,15 @@ def str_to_datetime64(s: str | int | NPDT64) -> NPDT64:
...
     """
 
     if isinstance(s, datetime.timedelta):
         return s
-    format_str = "%H:%M:%S"
+    format_str = "%d-%H:%M:%S"
     assert isinstance(s, str), type(s)
     t = datetime.datetime.strptime(s, format_str)
-    return datetime.timedelta(hours=t.hour, minutes=t.minute, seconds=t.second)
+    return datetime.timedelta(days=t.day, hours=t.hour, minutes=t.minute, seconds=t.second)
```
tjhunter: it seems to be a standard format for strptime. sounds good to me.

tjhunter (approved):

src/weathergen/datasets/data_reader_base.py
clessig: It's standard in python world. But even for 10 day forecasts we have hours as unit on our x-axis. So it's awkward from the domain science utility perspective.

clessig: Can we just use pandas for the dt conversion:

```
port pandas as pd

print(pd.to_timedelta("48:00:00"))
2 days 00:00:00
```

(thanks @mlangguth89 )

jpolz: Yes, we can. It is also convenient because you can go beyond 99 hours or use days.

src/weathergen/datasets/data_reader_base.py
```diff
@@ -74,18 +75,15 @@ def str_to_datetime64(s: str | int | NPDT64) -> NPDT64:
     return np.datetime64(datetime.datetime.strptime(str(s), format_str))
 
 
-def str_to_timedelta(s: str | datetime.timedelta) -> datetime.timedelta:
+def str_to_timedelta(s : str | datetime.timedelta) -> pd.Timedelta:
     """
-    Convert a string to a timedelta object.
-    The format is expected to be "HH:MM:SS".
+    Convert a string or datetime.timedelta object to a pd.Timedelta object.
+    The format is expected to be "HH:MM:SS" or "dd days HH:MM:SS".
```
clessig: Thanks for the changes--much better :) Did you test that one can specify "dd days HH:MM:SS" and it works. I would prefer if we use a unique format for delta times in the config and we agreed on "HH:MM:SS"

jpolz: I tested the following:
```
Converted '1 days 02:03:04' to Timedelta: 1 days 02:03:04
Converted '26:03:04' to Timedelta: 1 days 02:03:04
Converted timedelta 1 day, 2:03:04 to Timedelta: 1 days 02:03:04
Testing with invalid input: 12345
Error: Input must be a string or a datetime.timedelta object
```
I can do two things: 
enforce "HH:MM:SS" with by writing some checks
or
check all formats that `pd.to_timedelta` accepts and see if those are ok for us

clessig: My preference would be to enforce "HH:MM:SS"

jpolz: Done. Note that the following is allowed: `Converted '1440:00:00' to Timedelta: 60 days 00:00:00`

PR #501, opened 2025-07-11
Added naming convention checks to lint

### LINKED ISSUE #432: Adopt python naming conventions
We deviate from the python naming conventions in a few spots for no specific reason. The goal of this task is to rectify this:

Goal: In pyproject.toml, add the "N" rule in `[tool.ruff.lint]` .

I suggest that we ignore the following rules:
- "N806" -> we use heavy acronyms
- "N817", "N806" -> the tokenizer is being worked on currently, no need to duplicate efforts. This rule can be activated separately.

From a quick look, it leads to 9 offending places

MatKbauer:
## Type of Change

-   [ ] Documentation update

## Issue Number

closes #432


enssow (approved):
All looks good to me - checked with a small run

enssow (approved):
Approved post syncing w/ current develop

PR #510, opened 2025-07-13
Fixed randomization problem with masking 

### LINKED ISSUE #492: Loss oscilation during ERA5 training
### What happened?

Training loss is oscilating with high frequency as shown on the plot below

Image

Train for couple of epochs 

### Version

0.0.1

Linux

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

clessig:
## Description

Fixed randomization problem with masking

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #492

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/datasets/multi_stream_data_sampler.py
+++ b/src/weathergen/datasets/multi_stream_data_sampler.py
@@ -174,12 +174,13 @@ class MultiStreamDataSampler(torch.utils.data.IterableDataset):
 
         self.batch_size = batch_size
 
+        # ensure data_loader_rng_seed is not smaller than loader_num_workers to avoid
+        # issues in per loader rng seed computation
         self.data_loader_rng_seed = (
             cf.data_loader_rng_seed
             if cf.data_loader_rng_seed > cf.loader_num_workers
             else cf.data_loader_rng_seed * 13
         )
-        self.rng = np.random.default_rng(self.data_loader_rng_seed)
 
         self.healpix_level_source: int = cf.healpix_level
         self.healpix_level_target: int = cf.healpix_level
@@ -429,7 +430,7 @@ class MultiStreamDataSampler(torch.utils.data.IterableDataset):
             dist = torch.distributed
             self.data_loader_rng_seed *= (
                 (((dist.get_rank() + 1) * 73) if dist.is_initialized() else 1)
-                * ((worker_info.id + 1) * 39)
+                * ((worker_info.id + 1) * 37)
                 * (self.epoch + 13)
                 * 7
             )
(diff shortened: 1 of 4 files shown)


clessig:
This is with 2.5 epochs for 2 nodes = 8 DDP workers:

s03mn46x_k4hiy385_q8ow3kvj_i14tiowq_t039q2te_wkbn2v9t_train_val_ERA5

Also no repetitive pattern, so as far as I can tell the problem is fixed.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -413,6 +420,19 @@ def worker_workset(self):
...
         else:
+            # ensure the rng seed is fully unique across workers and epochs
+            # the worker processes are generated as bit-wise copy of the "template" (the actual
+            # instance of the present class that is created) whenever __iter__ is started. This
+            # happens for each epoch, for train and validation, and independently for each DDP
+            # worker. After the bit-wise copy, the rng seed needs to be made unique for
+            # DDP workers, loader process, epoch.
+            dist = torch.distributed
+            self.data_loader_rng_seed *= (
+                (((dist.get_rank() + 1) * 73) if dist.is_initialized() else 1)
```
sophie-xhonneux, line 432: is there a particular logic why we picked these numbers? they aren't all prime, which would make more sense to me

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -173,18 +173,24 @@ def __init__(
...
 
         self.batch_size = batch_size
-        self.rng = np.random.default_rng(cf.data_loader_rng_seed)
+
+        self.data_loader_rng_seed = (
+            cf.data_loader_rng_seed
+            if cf.data_loader_rng_seed > cf.loader_num_workers
+            else cf.data_loader_rng_seed * 13
+        )
+        self.rng = np.random.default_rng(self.data_loader_rng_seed)
```
sophie-xhonneux: Sorry if this is a stupid question, but shouldn't the seed also depend on the epoch? the way I understand the code right now we reset the self.rng to the same np.random with the same seed every epoch no?

src/weathergen/datasets/multi_stream_data_sampler.py
clessig: Yes, should all be primes. Fixed.

src/weathergen/datasets/multi_stream_data_sampler.py
clessig: The self.rng initalization here has been removed now, since the rng for and epoch is always created in `reset()`.

The epoch dependence is introduced in `worker_workset()`. The seed here is only used when one runs with `num_loader_workers = 0`, which is only for debug purposes and the randomization is not relevant.

PR #522, opened 2025-07-15
[DRAFT] CLI for scoring and plotting 

### LINKED ISSUE #173: Add performant plotting and evaluation script for ERA5 output
To evaluate results from #126, a simple Jupyter Notebook/script will be provided.
The script is limited to WeatherGenerator output on a fixed grid with a shared time stamp across data points per sample, such outputting data on ERA5's O96 grid.
Data reading from the zarr storage is done with dask, leveraging its lazy and memory efficient data processing capabilities.

The scripts/Jupyter Notebooks are added to JSC's WeatherGenerator-private folder.

### LINKED ISSUE #331: WeatherGenerator quick evaluation

Comparing models to guide development requires an easy to use, fast pipeline evaluation pipeline. This is in contrast to rigorous statistical evaluation frameworks to compare against NWP and other ML baselines. Currently this is done using a scattered collection of scripts. The goal is to unify this into an efficient pipeline used across teams.

This initiative can be roughly divided into:
- [ ] Making output from WeatherGenerator inference runs efficiently available in an convenient format (#365 )
- [ ] Implementing a flexible library of scoring tools (#332 )
- [ ] Developing a infrastructure to plot scores and share results within the WeatherGenerator collaboration (#333 )
- [ ] Providing an CLI interface to trigger scoring and plotting (#334 )

### Additional context

These ideas were developed during the June 2025 hackathon

### Organisation

_No response_

### LINKED ISSUE #333: plotting for quick evaluation

TBD

_No response_

_No response_

### Additional context

_No response_

### Organisation

_No response_

### LINKED ISSUE #479: Replace channel={number} with the name of the channel in the title

Image

Replace channel number with the channel name

https://gitlab.jsc.fz-juelich.de/esde/WeatherGenerator-private/-/blame/main/teams/ecmwf/plot.py?ref_type=heads#L75-L79

_No response_

### Additional context

_No response_

### Organisation

_No response_

iluise:
## Description

Add a unique user interface for scoring and plotting. 
Supersedes `plot.py`

## Type of Change

-   [ x] New feature (non-breaking change which adds functionality)
-   [x ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update

## Issue Number
Solves the following issues: 

Closes #479
Closes #333
Closes #331
Closes #173

## Code Compatibility

-   [x ] I have performed a self-review of my code

-   [ x] I ran the `uv run train` and (if necessary) `uv run evaluate` on a least one GPU node and it works 
-   [x ] If the new feature introduces modifications at the config level, I have made sure to have notified the other software developers through Mattermost and updated the paths in the `$WEATHER_GENERATOR_PRIVATE` directory

### Dependencies

-   [ x] I have ensured that the code is still pip-installable after the changes and runs
-   [ x] I have tested that new dependencies themselves are pip-installable.
-   [x ] I have not introduced new dependencies in the inference portion of the pipeline

### Documentation


packages/evaluate/src/weathergen/evaluate/plot_config.yaml
```diff
@@ -0,0 +1,74 @@
+verbose: true
```
tjhunter: let's put the config along with all the other configs?

tjhunter: also, we should separate the config parts related to plotting  and the pieces for generating the metrics:

```yaml
evaluation:
  metrics: ...
  
plotting:
  ...
  models:
    ...
```

packages/evaluate/src/weathergen/evaluate/plot_config.yaml
```diff
@@ -0,0 +1,74 @@
...
+overwrite : true
+metrics  : ["rmse", "l1", "mse"]
+avg_dims : ["ipoint", "sample"] #dimensions to average over when computing scores
+input_dir : "/home/ecm9336/eval_io/WeatherGenerator/results/"
```
tjhunter: `input` and `output` do not mean much. it should be `metrics_dir`

packages/evaluate/src/weathergen/evaluate/plot_config.yaml
```diff
@@ -0,0 +1,74 @@
...
+output_dir: "/ec/perm/ecm9336/evaluation/test/"
```
tjhunter: output_plotting_dir

packages/evaluate/src/weathergen/evaluate/plot_config.yaml
tjhunter: also, it should be a reasonable default: `results`

packages/evaluate/src/weathergen/evaluate/plot_config.yaml
```diff
@@ -0,0 +1,74 @@
...
+jsons_dir : "/ec/perm/ecm9336/evaluation/test/jsons/"
```
tjhunter: you should not put the format into the name of your config: `metrics_dir` ?

packages/evaluate/src/weathergen/evaluate/plot_config.yaml
```diff
@@ -0,0 +1,74 @@
...
+save_mode : "png" #options: "png", "pdf", "svg", "eps", "jpg" ..
```
tjhunter: not a mode: `save_format`

packages/evaluate/src/weathergen/evaluate/plot_config.yaml
```diff
@@ -0,0 +1,74 @@
...
+dpi_val : 300
+create_html : true
+fstep_summary_plots : true
```
tjhunter: what does that mean?

packages/evaluate/src/weathergen/evaluate/plot_config.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+#MTM models
+model_ids :
```
tjhunter: if you want to allow epoch, then the run_id should not be a key, it should be a list:

```
plotting:
  models:
     - run_id: wz...
        streams:...
        epoch: 0
     - run_id: wz...
        streams:...
        epoch: 30

File: packages/evaluate/src/weathergen/evaluate/plot_inference.py [hunk 8]
```diff
@@ -0,0 +1,216 @@
...
+import matplotlib.pyplot as plt
+import logging
+from plotter import Plotter, LinePlots
+from omegaconf import OmegaConf
+from weathergen.common.io import ZarrIO
+from pathlib import Path
+from tqdm import tqdm
+from collections import defaultdict
+
+handler = logging.StreamHandler(sys.stdout)
```
tjhunter: don't do it. I had a comment on #495  about it

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,216 @@
...
+_logger = logging.getLogger(__name__)
+_logger.setLevel(logging.INFO)
+_logger.addHandler(handler)
+_logger.propagate = False  # Prevent double logging if root logger also logs
+
+select = OmegaConf.select
+load = OmegaConf.load
+
+
+def retrieve_score(dir, model_id, stream, metric, epoch, rank):
```
tjhunter: I know it is quick scripts, but please put the types. It really helps understanding how it works.

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,216 @@
...
+    #TODO adapt this to JSON output
+    json_path = Path(f"{dir}/{metric}_{model_id}_{stream}_epoch{epoch:05d}_rank{rank:04d}.json")
```
tjhunter: I think the ranks/epoch is not included anymore?

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,216 @@
...
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(
+        description="Fast evaluation of weather generator runs."
+    )
+    parser.add_argument(
+        "--config",
+        type=str,
+        default="plot_config.yaml",
```
tjhunter: There should be no default, because it will not work in any case (the run_ids will be missing).

But in the help, you can include an example  or put the URL to the wiki page.

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def retrieve_score(dir, model_id, stream, metric, epoch, rank):
+    #TODO adapt this to JSON output
+    json_path = Path(f"{dir}/{metric}_{model_id}_{stream}_epoch{epoch:05d}_rank{rank:04d}.json")
+    _logger.info(f"Looking for: {json_path}")
+    if json_path.exists():
+        with open(json_path, "r") as f:
+            data_dict = json.load(f)
+            return xr.DataArray.from_dict(data_dict)
```
tjhunter: it is an xarray? I will have to run the code to understand it

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,216 @@
...
+
+                for metric in metrics:
+                    _logger.info(f"Computing {metric} for {model_id} - {stream}")
+                    try:
+                        #TODO: this needs more thinking as only part of the variables might be already pre-computed. 
+                        #with a loop over variables we would loose the advantage of parallel computing. 
+                        #possibility: load already computed scores before initialising the data and use e.g. "new_variable" index for the letfovers.
+                        #problem: different metrics can be pre-computed for different variables, so not easy to define a "new_variables" list.   
+                        scores_dict[metric][stream][model_id] = retrieve_score(jsons_dir, model_id, stream, metric, model.epoch, model.rank)
+                    except Exception as e:
```
tjhunter: let's think a bit about it. You do not want to do it in practice. Here, if you made a typo in your config yaml, you would recompute all the scores.

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,216 @@
...
+if cfg.fstep_summary_plots:
+
+    _logger.info(f"Started creating summary plots..")
+
+    plotter = LinePlots(cfg)
+    
+    for metric in metrics:
+        #get total list of streams
+        #TODO: improve this
+        streams_set = set.union(*[set(model["streams"].keys()) for model in models.values()])
```
tjhunter: make it deterministic: `list(sorted(set(...)))`

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,216 @@
...
+        
+        #get total list of variables
+        #TODO: improve this
+        variables_set = set.union(*[
```
tjhunter: same thing

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -0,0 +1,242 @@
...
+            #set labels and title
+            plt.xlabel(f"Variable: {var}")
+            plt.ylabel('Frequency')
+            plt.title(f"Histogram of Target and Prediction: {var}")
+            plt.legend(frameon=False)
+
+            #TODO: make this nicer
+            parts = ["histogram", self.model_id, tag, str(self.sample), self.stream, str(self.fstep), var]
+            name = "_".join(filter(None,parts))
+            plt.savefig(f"{self.out_dir.joinpath(name)}.{self.out_format}")
```
tjhunter: you should return the path to all the plots

packages/evaluate/src/weathergen/evaluate/plot_config.yaml
clessig: `figure_format` or `image_format`

packages/evaluate/src/weathergen/evaluate/plot_config.yaml
clessig: `forecast_summary_plots`

iluise:
These are the instructions to run the code atm:

```
git checkout --track origin/iluise/plotting
uv sync
cd ./packages/evaluate/src/weathergen/evaluate/
uv run plot_inference.py --config plot_config.yaml
```

I will restructure it based on your comments later today, but this won't change anyway

tjhunter:
Here is a shorter example from the package root. We should eventually put symlinks to all these scripts into `./scripts`

```
./packages/evaluate/src/weathergen/evaluate/plot_inference.py --config plot_config.yaml
```

with the config

```yaml
verbose: true
overwrite : true
avg_dims : ["ipoint", "sample"] #dimensions to average over when computing scores
input_dir : "{CLUSTER_FIXME_SHARED_RESULTS}/results/"
output_dir: "./test/"
jsons_dir : "./jsons/"
save_mode : "png" #options: "png", "pdf", "svg", "eps", "jpg" ..
dpi_val : 300
create_html : true
fstep_summary_plots : true

#MTM models
model_ids :
  h6swvykc:
    streams:
      ERA5:
        variables: ["2t", "10u"]
        fsteps: [0]
        samples: [0, 1]
    label: "MTM ERA5"
    epoch: 0
    rank: 0
    plot_maps: true
    plot_histograms: true
```

packages/evaluate/src/weathergen/evaluate/plot_inference.py
tjhunter: you are loading and saving using the default xarray format. I don't have a strong preference but it should align:
- @mlangguth89  writes in the DOP format
- we were talking about a different record-based format

Either is fine with me, but we should align on a first iteration. I was not aware that xarray had its own format. Maybe we should go with that for the time being, it already exists.

@mlangguth89  what do you think?

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,217 @@
...
+def retrieve_score(dir, model_id, stream, metric, epoch, rank):
+    #TODO adapt this to JSON output
+    json_path = Path(f"{dir}/{metric}_{model_id}_{stream}_epoch{epoch:05d}_rank{rank:04d}.json")
+    _logger.info(f"Looking for: {json_path}")
+    if json_path.exists():
+        with open(json_path, "r") as f:
+            data_dict = json.load(f)
+            return xr.DataArray.from_dict(data_dict)
+    else:
+        raise FileNotFoundError
```
tjhunter: put the filename. Make the error descriptive.

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,217 @@
...
+import cartopy.crs as ccrs
+import numpy as np
+from score import VerifiedData, get_score
+import itertools
+import os
+
+from typing import List
+import matplotlib.pyplot as plt
+import logging
+from plotter import Plotter, LinePlots
```
tjhunter: absolute path: `weathergen.evaluate.plotter`

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -0,0 +1,242 @@
...
+        :param cfg: config from the yaml file
+        :param model_id: if a model_id is given, the output will be saved in a folder called as the model_id
+        """
+        
+        self.cfg = cfg
+        
+        out_dir = Path(cfg.output_dir)
+        self.out_format = cfg.save_mode
+        self.create_html = cfg.create_html
+        self.dpi_val = OmegaConf.select(cfg, "dpi_val") or None
```
tjhunter: you could also do `cgf.get("dpi_val", None)`

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -0,0 +1,242 @@
...
+        self.fig_size = OmegaConf.select(cfg, "fig_size") or (8, 10)
```
tjhunter: `cgf.get("...", (8,10))`

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -0,0 +1,242 @@
...
+
+        self.out_dir = out_dir.joinpath(self.out_format).joinpath(model_id) 
+        overwrite = OmegaConf.select(cfg, "overwrite") or False
+        print(self.out_dir)
+        if overwrite:
+            if os.path.exists(self.out_dir):
+                shutil.rmtree(self.out_dir)  # Remove the existing directory
```
tjhunter: can we not do it for now? I nearly deleted my home directory by mistake

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -0,0 +1,242 @@
...
+        return ''.join(c if c.isalnum() else ' ' for c in string)   
+
+
+    def plot(self, data: xr.DataArray | list, labels: str | list, tag: str = "", x = "forecast_step", y = "value"):
+        """
+        Plot a line graph comparing multiple datasets.
+        :param data: DataArray or list of DataArrays to be plotted
+        :param labels: Label or list of labels for each dataset
+        :param tag: Tag to be added to the plot title and filename
+        :param x: Dimension to be used for the x-axis. The code will average over all other dimensions. (default is "forecast_step")
```
tjhunter: x_dim

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -0,0 +1,242 @@
...
+        :param y: Name of the dimension to be used for the y-axis (default is "value")
```
tjhunter: y_dim

tjhunter (commented):
A few more comments now that I tried your PR.

On a more general level: could you put types for all inputs and oututs of functions as well  as variables of classes? This really helps.

packages/evaluate/src/weathergen/evaluate/plot_inference.py
iluise: uhm, I still see it in the develop branch https://github.com/ecmwf/WeatherGenerator/blob/develop/src/weathergen/utils/config.py#L346 but I agree that it would be nice to remove them from the name, as they are anyway tracked into the config/json file of the run. Two inference runs of the same model at two different epochs would anyway have different `run_ids`, so I see no scenarios in which they will end up in the same directory and we need to distinguish them by epoch and rank.
What do you think? 

cc @clessig @grassesi

packages/evaluate/src/weathergen/evaluate/plot_config.yaml
iluise: oks for the evaluation separation. In addition, with Micha yesterday we were thinking about a structure like this:

```
model_ids :
  wz9rbl3w:
    streams:
      ERA5:
        forecast_step: [0]
        variables: ["2t", "10u"]
        evaluation: true
        plotting: 
          sample: [1, 2, 3]
          plot_maps: true
          plot_histograms: false
    label: "MTM ERA5"
    epoch: 0
    rank: 0
```
So the evaluation (aka scoring) will run over all samples, while the plotting only over a few samples

config/mixed.yml
```diff
@@ -1,2 +1,3 @@
 # The default configuration file for multi streams training.
-streams_directory: "./config/streams/streams_mixed/"
\ No newline at end of file
+streams_directory: "./config/streams/streams_mixed/"
```
clessig: Do we need to have this file in the PR?

config/plot_config.yaml
```diff
@@ -0,0 +1,51 @@
+verbose: true
```
clessig: File ending should be yml.

config/plot_config.yaml
```diff
@@ -0,0 +1,51 @@
...
+model_dir : "./results/"
```
clessig: Since we do have a model dir, which is different, this should be called `result_dir`

config/plot_config.yaml
```diff
@@ -0,0 +1,51 @@
...
+evaluation:
+  metrics  : ["rmse", "l1", "mse"]
+  avg_dims : ["ipoint", "sample"] #dimensions to average over when computing scores
+
+#MTM models
+model_ids :
+  wz9rbl3w:
+     streams:
+       ERA5:
+         forecast_step: [0]
```
clessig: Does one need to set this? It should, by default, use all forecast steps that are available.

config/plot_config.yaml
```diff
@@ -0,0 +1,51 @@
...
+output_plotting_dir: "./plots/"
+output_scores_dir : "./jsons/"
+image_format : "png" #options: "png", "pdf", "svg", "eps", "jpg" ..
+dpi_val : 300
+create_html : true
+summary_plots : true
+
+evaluation:
+  metrics  : ["rmse", "l1", "mse"]
+  avg_dims : ["ipoint", "sample"] #dimensions to average over when computing scores
```
clessig: I think `ipoint` is very unintuitive.

config/plot_config.yaml
```diff
@@ -0,0 +1,51 @@
...
+       ERA5:
+         forecast_step: [0]
+         channels: ["2t", "10u"]
+         evaluation: true
+         plotting: 
+           sample: [1, 2, 3]
+           plot_maps: false
+           plot_histograms: false
+     label: "MTM ERA5"
+     epoch: 0
```
clessig: This should also have defaults and it should not be necessary to specify it.

config/plot_config.yaml
```diff
@@ -0,0 +1,51 @@
...
+  avg_dims : ["ipoint", "sample"] #dimensions to average over when computing scores
+
+#MTM models
+model_ids :
+  wz9rbl3w:
+     streams:
+       ERA5:
+         forecast_step: [0]
+         channels: ["2t", "10u"]
+         evaluation: true
```
clessig: What does this mean?

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,103 @@
...
+
+from collections import defaultdict
+from utils import plot_data, plot_summary, calc_scores_per_stream, metric_list_to_json, retrieve_metric_from_json
+
+_logger = logging.getLogger(__name__)
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(
+        description="Fast evaluation of weather generator runs."
```
clessig: WeatherGenerator

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,103 @@
...
+    # configure logging
+    logging.basicConfig(level=logging.INFO)
+
+    # load configuration
+    cfg = OmegaConf.load(args.config)
+    models = cfg.model_ids
+
+    _logger.info(f"Detected {len(models)} models")
+    
+    assert cfg.get("output_scores_dir"), "Please provide a path to the directory where the score files are stored or will be saved."
```
clessig: This should default to ./results/{RUN_ID}

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,103 @@
...
+    out_scores_dir = Path(cfg.output_scores_dir)
+    out_scores_dir.mkdir(parents=True, exist_ok=True)
+
+    model_dir = Path(cfg.model_dir)
```
clessig: I think it should also be results_dir

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+
+    for model_id, model  in models.items():
+
+        plotter = Plotter(cfg, model_id)
+        _logger.info(f"MODEL {model_id}: Getting data...")
+        
+        streams = model["streams"].keys()
+
+        for stream in streams:
```
clessig: `for stream, stream_dict in model['streams'].items()`

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,103 @@
...
+                    dict
+            )
+        )
+
+    for model_id, model  in models.items():
+
+        plotter = Plotter(cfg, model_id)
+        _logger.info(f"MODEL {model_id}: Getting data...")
+        
+        streams = model["streams"].keys()
```
clessig: See below.

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,103 @@
...
+
+        for stream in streams: 
+            _logger.info(f"MODEL {model_id}: Processing stream {stream}...")
+            
+            stream_dict = model["streams"][stream]
```
clessig: Remove. See above.

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,103 @@
...
+    )
+
+    args = parser.parse_args()
+
+    # configure logging
+    logging.basicConfig(level=logging.INFO)
+
+    # load configuration
+    cfg = OmegaConf.load(args.config)
+    models = cfg.model_ids
```
clessig: This are the configs, right? I don't think `models` is a good name since (for me) it's more the weights. Being explicit and calling it `model_configs` would be better.

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,103 @@
...
+        _logger.info(f"MODEL {model_id}: Getting data...")
+        
+        streams = model["streams"].keys()
+
+        for stream in streams: 
+            _logger.info(f"MODEL {model_id}: Processing stream {stream}...")
+            
+            stream_dict = model["streams"][stream]
+
+            _logger.info(f"MODEL {model_id}: Plotting stream {stream}...")
```
clessig: RUN and run_id

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,103 @@
...
+    metrics = cfg.evaluation.metrics
+    
+    # to get a structure like: scores_dict[metric][stream][model_id] = plot
+    scores_dict = defaultdict(           
+                lambda: defaultdict(
+                    dict
+            )
+        )
+
+    for model_id, model  in models.items():
```
clessig: `run_id, run`

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,103 @@
...
+        
+        streams = model["streams"].keys()
+
+        for stream in streams: 
+            _logger.info(f"MODEL {model_id}: Processing stream {stream}...")
+            
+            stream_dict = model["streams"][stream]
+
+            _logger.info(f"MODEL {model_id}: Plotting stream {stream}...")
+            plots = plot_data(cfg, model_id, stream, stream_dict)
```
clessig: `run_id` 

(same below)

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,103 @@
...
+                _logger.info(f"Retrieve or compute scores for {model_id} - {stream}...")
+
+                metrics_to_compute = []
+                for metric in metrics:
+                    try: 
+                        metric_data = retrieve_metric_from_json(out_scores_dir, model_id, stream, metric, model.epoch, model.rank)
+                        scores_dict[metric][stream][model_id] = metric_data
+                    except (FileNotFoundError, KeyError, ValueError) as e:
+                        metrics_to_compute.append(metric)
+                if metrics_to_compute:
```
clessig: Does this pass ruff?

`len(metrics_to_compute) > 0`

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,103 @@
...
+                        scores_dict[metric][stream][model_id] = metric_data
+                    except (FileNotFoundError, KeyError, ValueError) as e:
+                        metrics_to_compute.append(metric)
+                if metrics_to_compute:                
+                    all_metrics, points_per_sample = calc_scores_per_stream(cfg, model_id, stream, metrics_to_compute)
+
+                    metric_list_to_json([all_metrics], [points_per_sample], [stream], out_scores_dir, model_id, model.epoch, model.rank)
+
+                    all_metrics = all_metrics.compute()
+                    all_metrics = all_metrics.mean(dim="sample")
```
clessig: Why do we always take the mean over `samples` here?

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -0,0 +1,284 @@
...
+
+
+class Plotter(object):
+    """
+    Contains all basic plotting functions.
+    """
+    def __init__ (self, cfg: dict, model_id: str = ""):
+        """
+        Initialize the Plotter class.
+        :param cfg: config from the yaml file
```
clessig, line 26: That's not the agreed on style for comments.

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -0,0 +1,284 @@
...
+import shutil
+import logging
+logging.getLogger("matplotlib.category").setLevel(logging.ERROR)
+
+_logger = logging.getLogger(__name__)
+_logger.setLevel(logging.INFO)
+
+
+import cartopy
+import cartopy.crs as ccrs  #https://scitools.org.uk/cartopy/docs/latest/installing.html
```
clessig: Why is the comment relevant.

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -0,0 +1,284 @@
...
+        self.sample   = None
+        self.stream   = None
+        self.fstep    = None
+        self.model_id = model_id
+        self.select   = {}
+
+
+    def update_selection(self, sample: str, stream: str, fstep:str):
+        """
+        Set the selection for the plots. This will be used to filter the data for plotting.
```
clessig: selection of what? scores? types of plots?

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -0,0 +1,284 @@
...
+        self.stream = stream 
+        self.fstep  = fstep
+        self.select = { "sample" : self.sample, 
+                        "stream" : self.stream, 
+                        "forecast_step"  : self.fstep }
+        return self
+    
+    def update_selection(self, select: dict):
+        """
+        Set the selection for the plots. This will be used to filter the data for plotting.
```
clessig, line 51: Same as above

config/plot_config.yaml
```diff
@@ -0,0 +1,51 @@
+verbose: true
+model_dir : "./results/"
+output_plotting_dir: "./plots/"
```
clessig: Output dir: For hygiene and traceability reasons, the run-specific plots should go to `results/{RUN_ID}/plots/`. We can create a symlink to `results/{RUN_ID}/plots/` from plots which helps with vscode. I also cannot think of a better solution than having the plots that combine info from different runs in `plots/` (but maybe someone else can).

config/plot_config.yaml
```diff
@@ -0,0 +1,51 @@
...
+#MTM models
+model_ids :
+  wz9rbl3w:
+     streams:
+       ERA5:
+         forecast_step: [0]
+         channels: ["2t", "10u"]
+         evaluation: true
+         plotting: 
+           sample: [1, 2, 3]
```
clessig: This should default to all samples in the inference run.

config/plot_config.yaml
iluise: for plotting one won't plot all forecast steps so we need a place to specify it. if forecast step is not specified, then it automatically gets all forecast steps anyway. One option could be to move it under plotting (like what happens for sample) and under evaluation in case one wants to subset the forecast steps in the metrics. 
cc @mlangguth89

config/plot_config.yaml
iluise: this is just for plotting. if sample is not specifies it will plot all samples, but one will never plot maps and hists for all samples. the evaluation is done over all samples anyway.

config/plot_config.yaml
tjhunter: @grassesi  FYI. Maybe a name we can improve eventually?

config/plot_config.yaml
tjhunter: to me:
- all plots for a specific run id should go into `results/{run_id}/plots`
- all plots for runid1, runid2, ... should go into `results/{runid1}-{runid2}-{runid3}/plots`

I am rather neutral because these plots are cheap to produce and they are really for story telling.

packages/evaluate/src/weathergen/evaluate/fast_evaluation.py
```diff
@@ -15,7 +15,7 @@
 import numpy as np
 import xarray as xr
 from score import VerifiedData, get_score
-from utils import to_list
+from score_utils import to_list
```
tjhunter, line 18: weathergen.evaluate.score_utils

It should not pass the linter right now, I will check that

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env -S uv run
+# /// script
+# dependencies = [
+#   "weathergen-evaluate",
+#   "weathergen-common",
+#   "panel",
```
tjhunter: panel and omegaconf should be dependencies of `weathergen-evaluate`

packages/evaluate/src/weathergen/evaluate/plotter.py
tjhunter: I think there are linter rules for that. So far it has been the wild west, but feel free to open an issue.

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -0,0 +1,284 @@
...
+        
+        out_plot_dir = Path(cfg.output_plotting_dir)
+        self.image_format = cfg.image_format
+        self.create_html = cfg.create_html
+        self.dpi_val = cfg.get("dpi_val", None)
+        self.fig_size = cfg.get("fig_size", (8, 10)) 
+
+        self.out_plot_dir = out_plot_dir.joinpath(self.image_format).joinpath(model_id) 
+
+        os.makedirs(self.out_plot_dir, exist_ok=True)
```
tjhunter: for such things, it is always good to put an info message, because you could create/remove directories by mistake in surprising places. 

```py
if not os.exists(self.out_plot_dir):
   _logger.info(creating dir ...)
   os.makedirs
```

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -0,0 +1,284 @@
...
+            plt.legend(frameon=False)
+
+            #TODO: make this nicer
+            parts = ["histogram", self.model_id, tag, str(self.sample), self.stream, str(self.fstep), var]
+            name = "_".join(filter(None,parts))
+            plt.savefig(f"{self.out_plot_dir.joinpath(name)}.{self.image_format}")
+            plt.close()
+            plot_names.append(name)
+
+        self.clean_selection()
```
tjhunter: I would not clean. this is surprsing for the user

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -0,0 +1,284 @@
...
+        # convert to lists
+        
+        data_list = [data] if type(data) == xr.DataArray else data
+        label_list = [labels] if type(labels) == str else labels
+
+        assert len(data_list) == len(label_list), "Compare::plot - Data and Labels do not match"
+        
+        return data_list, label_list
+
+    def clean_string(self, string: str):
```
tjhunter: nit: it does not need to be a method

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -8,23 +8,397 @@
...
+
+        da_tars, da_preds = [], []
+
+        if return_counts:
+                points_per_sample = xr.DataArray(
+                    np.full((len(fsteps), len(samples)), np.nan),
+                    coords={"forecast_step": fsteps, "sample": samples},
+                    dims=("forecast_step", "sample"),
+                    name=f"points_per_sample_{stream}",
+                )
```
tjhunter:
```
else:
   points_per_sample = None
```

(you will see why below)

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -8,23 +8,397 @@
...
+            da_preds.append(da_preds_fs)
+            if return_counts:
+                points_per_sample.loc[{"forecast_step": fstep}] = np.array(pps)
+        _logger.debug(
+                f"Concatenating targets and predictions along the forecast_step dimension..."
+            )
+        da_tars  = xr.concat(da_tars , dim="forecast_step")
+        da_preds = xr.concat(da_preds, dim="forecast_step")
+
+        if return_counts:
```
tjhunter: From experience, this is something that really causes confusion, because people do not know what to unpack on the other end. You are really better off with a dataclass (you can ask @Jubeku  and @MatKbauer for all the reasons). In this case:

```py
@dataclass 
ZarrData:
   target: DataArray
   prediction: DataArray
   points_per_sample: DataArray | None       # <- now this is clear that it may or may not be present

def get_data(...) -> ZarrData:
  ...
  return ZarrData(da_tars, da_preds, points_per_sample)
```

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -8,23 +8,397 @@
...
+            return da_tars, da_preds, points_per_sample
+        else:
+            return da_tars, da_preds
+
+
+def calc_scores_per_stream(cfg: dict, model_id: str, stream: str, metrics: list[str]) -> tuple:
```
tjhunter: `tuple[DataArray, DataArray]`

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -8,23 +8,397 @@
...
+    _logger.debug(f"Running computation of metrics for stream {stream}...")
+    combined_metrics = combined_metrics.compute()
+    combined_metrics = scalar_coord_to_dim(combined_metrics, "channel")
+
+    _logger.info(f"Scores for model {model_id} - {stream} calculated successfully.")
+
+    return combined_metrics, points_per_sample
+
+
+def plot_data(cfg: str, model_id: str, stream: str, stream_dict: dict):
```
tjhunter: `-> list[str]` ?

tjhunter (commented):
@iluise  thanks! I have yet to try it. A few suggestions on my side.

config/plot_config.yaml
grassesi: Its basically an index of all points that fall within the original time window. So maybe `datapoint`? but that is also not very descriptive :shrug: . Anyway if that should be changed, it needs to be implemented in the io.py

packages/evaluate/src/weathergen/evaluate/plotter.py
iluise: the logic behind it is a bit like plt.close(), clean and reset the selection and be 100% sure that the previous selection is not there while plotting the next sample

config/plot_config.yaml
iluise: we need these plots in a single location for the html interface. though right? and the initial idea was to upload the analysis to the S3 bucket to be shared across partners who don't anyway work on the same HPC. Can we wait for the final decision on the uploading/sharing/html thing before changing it?

iluise:
Hi, 
just to say that I finished implementing all the comments apart for changing the output paths (see comment above). Can you please test it again?
(the config changed as well so remember to update it wrt what you were using)

cc  @mlangguth89

config/mixed.yml
mlangguth89: We should not touch parts of the code not related to this PR.

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -122,31 +140,72 @@ def calc_scores_per_stream(cfg: dict, model_id: str, stream: str, metrics: list[
...
+        samples = None
+    
+    if fsteps == "all":
+        fsteps = None
+
+    da_preds, da_tars, points_per_sample = get_data(cfg, model_id, stream, return_counts=True)
+
+    assert da_preds.keys() == da_tars.keys(), "Forcast steps in prediction and target do not match. Investigate."
+
+    first_da = next(iter(da_tars.values()))
```
mlangguth89: @iluise 
Caution: Don't apply `.values` on a dask-array! This loads all the data into memory.

Why do we need do this here at all? All the information below can be deduced from zarrIO-object itself.

COORECTION:
I misunderstood the code. `da_tars.values()` does not load the data into memory, but fetches all data arrays for all forecast steps. 
We can, however, circumvent the need to get all the coordinate/dimension-information from the data by just collecting all metric-arrays in a list, and then concat this list of xarrays after looping over forecast_steps.
Symbolic code:
```
metrics_list = []
for fstep in forecast_steps:
   metrics_fstep = [get_score(data, metric, ...) for metric in metrics]
   metrics_all.append(metric_fstep.compute())
 
 metric_stream = xr.concat(metrics_list, dim="forecast_step")
 metric_stream = metric_stream.assign_coords({"forecast_step": fsteps}) 
 ```

File: packages/evaluate/src/weathergen/evaluate/utils.py [hunk 59]
```diff
@@ -122,31 +140,72 @@ def calc_scores_per_stream(cfg: dict, model_id: str, stream: str, metrics: list[
...
+            f"Build computation graphs for metrics for stream {stream}..."
+        )
+
+        combined_metrics = [
+            get_score(score_data, metric, agg_dims="ipoint", group_by_coord="sample") for metric in metrics
+        ]
 
-    _logger.debug(f"Running computation of metrics for stream {stream}...")
-    combined_metrics = combined_metrics.compute()
-    combined_metrics = scalar_coord_to_dim(combined_metrics, "channel")
```
mlangguth89: This was required to handle the case where the user wants to compute the scores for a single channel. 
I will check this edge case.

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -122,31 +140,72 @@ def calc_scores_per_stream(cfg: dict, model_id: str, stream: str, metrics: list[
...
+        combined_metrics = xr.concat(combined_metrics, dim="metric")
+        combined_metrics["metric"] = metrics
 
-    _logger.info(f"Scores for model {model_id} - {stream} calculated successfully.")
+        _logger.debug(f"Running computation of metrics for stream {stream}...")
+        combined_metrics = combined_metrics.compute()
 
-    return combined_metrics, points_per_sample
+        metric_stream.loc[{"forecast_step": int(fstep)}] = combined_metrics.compute()
```
mlangguth89: No need to call `.compute()` twice (already done in l.195).

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -99,19 +105,31 @@ def get_data(cfg: dict, model_id: str, stream: str, samples: list[int] = None,
...
-                f"Concatenating targets and predictions along the forecast_step dimension..."
-            )
-        da_tars  = xr.concat(da_tars , dim="forecast_step")
-        da_preds = xr.concat(da_preds, dim="forecast_step")
+        
+        #Safer than a list
+        da_tars  = {fstep: da for fstep, da in zip(fsteps, da_tars)}
+        da_preds = {fstep: da for fstep, da in zip(fsteps, da_preds)}
+
+        assert da_preds.keys() == da_tars.keys(), "Forcast steps in prediction and target do not match. Investigate."
```
mlangguth89: I think this is not a useful sanity check:
As far as I can see, they cannot be different since the same `fsteps` is used when creating both dictionaries.

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -99,19 +105,31 @@ def get_data(cfg: dict, model_id: str, stream: str, samples: list[int] = None,
...
+
+        #sanity checks
+        for fstep in da_tars.keys():
+            assert np.array_equal(da_tars[fstep].sample.values, da_preds[fstep].sample.values), f"Samples in targets and predictions do not match for model {model_id} and stream"
```
mlangguth89: I don't think we need this sanity check here.
If the number of samples is different between the target and prediction data, we would encounter errors in the score-computation. An operation such as (tar - pred) cannot be applied then.

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -83,13 +83,13 @@
...
                         metrics_to_compute.append(metric)
-                if metrics_to_compute:                
+                if metrics_to_compute: 
+                        
                     all_metrics, points_per_sample = calc_scores_per_stream(cfg, model_id, stream, metrics_to_compute)
 
                     metric_list_to_json([all_metrics], [points_per_sample], [stream], out_scores_dir, model_id, model.epoch, model.rank)
 
                     all_metrics = all_metrics.compute()
-                    all_metrics = all_metrics.mean(dim="sample")
```
mlangguth89: We track the per-sample scores in `scores_dict` now?

mlangguth89: Note: We also don't need to run `.compute()` on `all_metrics`. This is already done in `calc_scores_per_stream`.

mlangguth89:
> Hi, just to say that I finished implementing all the comments apart for changing the output paths (see comment above). Can you please test it again? (the config changed as well so remember to update it wrt what you were using)
> 
> cc @mlangguth89

I've spotted a few issues in the recent code (see review comments below). I'll take care of them tomorrow.

config/plot_config.yaml
mlangguth89: No strong feelings about the naming of this dimension from my side.
Let's keep it for this PR since it is part of `io.py`.
There is anyway work left for the IO and specifically the `as_xarray`-method (e.g. redundancy with singleton dimensions).

config/plot_config.yaml
mlangguth89: If the user does not specify `forecast_steps`, all `forecast_steps` in the inference data will be evaluated (= scores will be computed). It should just be an optional parameter to reduce the computation if the user desires to do so (for whatever reasons).

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -7,24 +7,532 @@
...
+from dataclasses import dataclass
+from weathergen.common.io import ZarrIO
 
-def to_list(obj: Any) -> list:
+_logger = logging.getLogger(__name__)
+_logger.setLevel(logging.INFO)
+
+
+@dataclass 
+class ZarrData:
```
mlangguth89: This name is not very descriptive and rather misleading.
We here have a data class for the inference data of the WeatherGenerator. In principle, this does not need to be read from a zarr-storage, but could even be read from netCDF-files (if we would already an interface for this in place).

mlangguth89: I suggest to rename it to `WeatherGeneratorOutput`

packages/evaluate/src/weathergen/evaluate/plot_inference.py
mlangguth89: Resolved

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -7,24 +7,514 @@
...
+_logger.setLevel(logging.INFO)
+
+
+@dataclass
+class WeatherGeneratorOutput:
+    target: dict
+    prediction: dict
+    points_per_sample: (
+        xr.DataArray | None
+    )  # <- now this is clear that it may or may not be present
```
tjhunter, line 34: I think you can remove my comment here :-)

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -7,24 +7,514 @@
...
+        _logger.info(f"RUN {run_id}: Processing stream {stream}...")
+
+        fsteps = zio_forecast_steps if fsteps is None else fsteps
+        # TODO: Avoid conversion of fsteps and sample to integers (as obtained from the ZarrIO)
+        fsteps = sorted([int(fstep) for fstep in fsteps])
+        samples = sorted(
+            [int(sample) for sample in zio.samples] if samples is None else samples
+        )
+        channels = (
+            channels
```
tjhunter, line 84: FYI, you can also say: `channels = channels or stream_dict.get(...)`

Issue #504, opened 2025-07-11
Support loading FESOM data from split zarr archives


kacpnowak:
### Is your feature request related to a problem? Please describe.

With century-scale datasets we are running into issues when converting data to zarr archives. It's more convinient to split such dataset into multiple zarr's 

### Describe the solution you'd like

In the FESOM dataloader add option to load multiple files based regex and concatenate them 

### Organisation

AWI

PR #524, opened 2025-07-15
Implement splitting zarr and regex filenames

kacpnowak:
## Description

Implements support for loading data from multiple zarr files, provided regular expresion to find their paths 

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/350
Closes https://github.com/ecmwf/WeatherGenerator/issues/504

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/datasets/data_reader_fesom.py
+++ b/src/weathergen/datasets/data_reader_fesom.py
@@ -7,10 +7,12 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+import glob
 import logging
 from pathlib import Path
 from typing import override
 
+import dask.array as da
 import numpy as np
 import zarr
 
@@ -77,15 +79,32 @@ class DataReaderFesom(DataReaderTimestep):
         filename: Path,
         stream_info: dict,
     ) -> None:
-        self.filename = filename
-        self.ds = zarr.open(filename, mode="r")
-        self.mesh_size = self.ds.data.attrs["nod2"]
+        self.filenames = sorted(glob.glob(str(filename)))
 
-        self.time = self.ds["dates"]
+        if len(self.filenames) == 0:
+            name = stream_info["name"]
+            _logger.warning(
+                f"{name} couldn't find any files matching {filename}. Stream is skipped."
+            )
+            super().__init__(tw_handler, stream_info)
+            self.init_empty()
+            return
+
+        groups: list[zarr.Group] = [zarr.open_group(name, mode="r") for name in self.filenames]
+        times: list[zarr.Array] = [group["dates"] for group in groups]
+        data: list[zarr.Array] = [group["data"] for group in groups]
+
+        self.time = da.concatenate(times, axis=0)
+        self.data = da.concatenate(data, axis=0)
+
+        if "nod2" in groups[0].data.attrs:
+            self.mesh_size = groups[0].data.attrs["nod2"]
+        else:
+            self.mesh_size = groups[0].data.attrs["n_points"]
 
         # TODO: time conversion to datetime64 should happen here.
-        start_ds = self.time[0][0]
-        end_ds = self.time[-1][0]
+        start_ds = self.time[0][0].compute()
+        end_ds = self.time[-1][0].compute()
 
         if start_ds > tw_handler.t_end or end_ds < tw_handler.t_start:
             name = stream_info["name"]
@@ -94,7 +113,7 @@ class DataReaderFesom(DataReaderTimestep):
             self.init_empty()
             return
 
-        period = self.time[self.mesh_size][0] - self.time[0][0]
+        period = (self.time[self.mesh_size][0] - self.time[0][0]).compute()
 
         if tw_handler.t_start > start_ds:
             self.start_idx = ((tw_handler.t_start - start_ds) // period + 1) * self.mesh_size
@@ -121,7 +140,7 @@ class DataReaderFesom(DataReaderTimestep):
             period,
         )
 
-        self.colnames: list[str] = list(self.ds.data.attrs["colnames"])
+        self.colnames: list[str] = list(groups[0].data.attrs["colnames"])
         self.cols_idx = list(np.arange(len(self.colnames)))
         self.lat_index = list(self.colnames).index("lat")
         self.lon_index = list(self.colnames).index("lon")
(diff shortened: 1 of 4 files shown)


src/weathergen/datasets/data_reader_fesom.py
```diff
@@ -77,102 +44,145 @@ def __init__(
         filename: Path,
         stream_info: dict,
     ) -> None:
-        self.filename = filename
-        self.ds = zarr.open(filename, mode="r")
-        self.mesh_size = self.ds.data.attrs["nod2"]
+        # Store configuration but DO NOT open files here
+        self.filenames = sorted(glob.glob(str(filename)))
```
tjhunter: thanks for sorting, it helps with reproducibility

src/weathergen/datasets/data_reader_fesom.py
```diff
@@ -77,102 +44,145 @@ def __init__(
...
+        if len(self.filenames) == 0:
+            self.init_empty()
+            self._initialized = True
+            return
 
-        # TODO: time conversion to datetime64 should happen here.
-        start_ds = self.time[0][0]
-        end_ds = self.time[-1][0]
+        # Initialize data-dependent attributes to None. They will be set by _lazy_init.
+        self.time: da.Array | None = None
```
tjhunter, line 58: personally, I prefer having a small dataclass with all the lazily initialized values. I find it easier to reason about. Your call.

config/streams/streams_ocean/fesom.yml
```diff
@@ -7,12 +7,12 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-FESOM :
+FESOM_NODE :
   type : fesom
-  filenames : ['fesom_ifs_awi']
+  filenames : ['ocean_node/*']
```
tjhunter: all the sources are essentially globs. can we just have `ocean_node` here in the config, in line with the other sources? then we do the globbing in the data reader.

Also, why `_NODE`?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -129,7 +129,8 @@ def __init__(
                 else:
                     filename = pathlib.Path(datapath) / fname
 
-                    if not filename.exists():  # see above
+                    # DataReaderFesom can take regular expressions so this check is bypassed
+                    if not filename.exists() and stream_info["type"] != "fesom":  # see above
```
tjhunter: with my suggestion about moving the `/*` into the data reader, you do not need that

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -139,7 +140,7 @@ def __init__(
                 ds_type = stream_info["type"]
                 logger.info(
                     f"Opening dataset with type: {ds_type}"
-                    + f"from stream config {stream_info['name']}.",
+                    + f" from stream config {stream_info['name']}.",
```
tjhunter, line 135: good catch

tjhunter (changes requested):
@kacpnowak  I did a light review on the data reader since I do not want to block you. I just have one comment about the location of the globbing.

src/weathergen/datasets/multi_stream_data_sampler.py
kacpnowak: Good point. Will fix that

src/weathergen/datasets/data_reader_fesom.py
kacpnowak: The reason I did it this way was issue between dask and torch. When creating workers each worker process gets a copy of the memory of the main process. This leads to deadlocks as multiple processes try to control the dask scheduler. Now when each worker starts running, it uses lazy_init independently allowing them to operate independently

config/streams/streams_ocean/fesom.yml
kacpnowak: It's due to the way the fesom works. Some variables (temperature, salinity) are computed on mesh nodes, while others are computed on elements (velocity). This results in having diffrent output variables having different number of "measurements" of variables placed on nodes and on elements. I think it's just easier to have two different streams for them.

src/weathergen/datasets/data_reader_fesom.py
```diff
@@ -77,102 +44,178 @@ def __init__(
...
+            # If the order is already correct, no need to do anything.
+            if local_colnames == self.colnames:
+                reordered_data_arrays.append(da.from_zarr(group["data"]))
+            else:
+                # Create the list of indices to re-shuffle the columns.
+                reorder_indices = [local_colnames.index(name) for name in self.colnames]
+
+                # Lazily re-index the dask array. This operation is not executed immediately.
+                dask_array = da.from_zarr(group["data"])
+                reordered_array = dask_array[:, reorder_indices]
```
clessig, line 158: Does this provide any benefit? The dataset size might be multiple TBs. How is the reordering supposed to happen?

src/weathergen/datasets/data_reader_fesom.py
```diff
@@ -77,102 +44,178 @@ def __init__(
...
-            self.source_idx = self.cols_idx
+        self.data = da.concatenate(reordered_data_arrays, axis=0)
 
-        target_channels = stream_info.get("target")
-        if target_channels:
-            self.target_channels, self.target_idx = self.select(target_channels)
-        else:
-            self.target_channels = self.colnames
-            self.target_idx = self.cols_idx
+        source_channels = self._stream_info.get("source")
```
clessig, line 181: This should be implemented in a separate function `select_channels`, see the data_reader_anemoi class.

src/weathergen/datasets/data_reader_fesom.py
kacpnowak: That's just a necessary thing to do because we've made mistake while putting dataset into zarr. Each year has different order of columns, making it necessary to reorder them so they can be aligned for stacking. 

TBH the best solution would be just to regenerate files but that's very time consuming. At the same time, this doesn't seem to impact IO performance in measurable way

src/weathergen/datasets/data_reader_fesom.py
kacpnowak: Done

src/weathergen/datasets/data_reader_fesom.py
```diff
@@ -7,19 +7,24 @@
...
+    DTRange,
+    NDArray,
     ReaderData,
     TimeWindowHandler,
     TIndex,
-    check_reader_data,
+    t_epsilon,
 )
 
 _logger = logging.getLogger(__name__)
```
clessig, line 30: Can you remove this logger statement.

tjhunter (approved):

PR #440, opened 2025-07-03
Implement per channel logginig again

kacpnowak:
## Description

Implements logging losses per channel, now taking into account that ranks can have different number of samples. 

## Type of Change

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update

## Issue Number

Closes #282 

## Code Compatibility

-   [x] I have performed a self-review of my code

### Code Performance and Testing

-   [x] I ran the `uv run train` and (if necessary) `uv run evaluate` on a least one GPU node and it works 
-   [ ] If the new feature introduces modifications at the config level, I have made sure to have notified the other software developers through Mattermost and updated the paths in the `$WEATHER_GENERATOR_PRIVATE` directory

### Dependencies

-   [ ] I have ensured that the code is still pip-installable after the changes and runs
-   [ ] I have tested that new dependencies themselves are pip-installable.
-   [ ] I have not introduced new dependencies in the inference portion of the pipeline

### Documentation

-   [x] My code follows the style guidelines of this project
-   [ ] I have updated the documentation and docstrings to reflect the changes
-   [ ] I have added comments to my code, particularly in hard-to-understand areas

Issue #414, opened 2025-07-01
backward compatiblitly for newly added configuration options


grassesi:
### Is your feature request related to a problem? Please describe.

Adding new configuration options currently results in incompatibilities on inference/evaluation with older models that dont have the new option as part of their `model.json` configuration file. Since it is foreseen that new parameters regularly are added, a basic level of backward compatibility should be introduced to allow for people having updated code but avoid breaking their models. This can be implemented as a mechanism that applies defaults when the loaded `model.json` file is missing a parameter.

### Describe the solution you'd like

A possible solution would be reusing the existing mechanisms for merging and overwriting configuration from different sources: The `config.load_config` method is responsible for this task.
The current logic considers two different sources of `base_config` (the configuration containing all needed options and lowest overwrite priority) depending on if the current run was started based on a previous run (`from_run_id`).
```python
if from_run_id is None:
    base_config = _load_default_conf()
else:
    base_config = load_model_config(from_run_id, epoch, private_config["model_path"])

return OmegaConf.merge(base_config, private_config, *overwrite_configs)
```
Instead configurations loaded from previous runs could be interpreted as overwrites to the default config. The default config should always contain defaults for all options.
```python
default_config = _load_default_conf()
base_config = load_model_config(from_run_id, epoch, private_config["model_path"]) if from_run_id is None else {}

return OmegaConf.merge(default_config, base_config, private_config, *overwrite_configs)
```
If `config.load_model_config` is changed to return `{}` if `from_run_id == 0` this can be simplified to:
```python
default_config = _load_default_conf()
base_config = load_model_config(from_run_id, epoch, private_config["model_path"])

return OmegaConf.merge(default_config, base_config, private_config, *overwrite_configs)
```

### Describe alternatives you've considered

Alternatively the configuration could be augmented with python `dataclasses`. Default configuration and backward compatibility could be handled by those. Additionally it would help documenting the code since inspecting the configuration object at runtime is currently the only possibility to know its contents. `dataclasses` integrates also well with `OmegaConf` via its [Structured Configs](https://omegaconf.readthedocs.io/en/2.3_branch/structured_config.html#structured-configs) mechanism. Implementing this will require significantly more effort then the solution presented above.

### Additional context

[...]

PR #533, opened 2025-07-16
Loss class refactoring

### LINKED ISSUE #568: Loss function refactoring

Currently, the lsos is computed in a lengthly function within the trainer. 

Refactor the current loss function into a standalone class. 
Design document: [link](https://ecmwf-weathergen.onlyoffice.com/doceditor?fileId=1919226&share=UklrRklBUjlENVc2alpYdkFMQ2NSVWtLTEcrTnZWTFU3ZC9vSk9yUjZUYz0_IjFlYTRiMjQzLWZhMjUtNDBiMy04OTg1LWQyNTYyMzY0NDZjYyI)

_No response_

### Additional context

_No response_

### Organisation

ECMWF

Jubeku:
## Description

In this PR we refactor the `trainer.compute_loss()` function into a standalone `LossModule` class. 
Note: This PR works off of Kacper's branch [kacpnowak:kacpnowak/develop/per-channel-logginig](https://github.com/kacpnowak/WeatherGenerator2/tree/kacpnowak/develop/per-channel-logginig)

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #568 

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/train/trainer.py
+++ b/src/weathergen/train/trainer.py
@@ -27,7 +27,6 @@ from torch.distributed.fsdp.wrap import (
     size_based_auto_wrap_policy,  # default_auto_wrap_policy,
 )
 
-import weathergen.train.loss as losses
 import weathergen.utils.config as config
 from weathergen.datasets.multi_stream_data_sampler import MultiStreamDataSampler
 from weathergen.model.model import Model, ModelParams
@@ -119,9 +118,7 @@ class Trainer(TrainerBase):
         self.model_params = ModelParams().create(cf).to(self.devices[0])
         _logger.info(f"Loaded model id={run_id_trained} at epoch={epoch}.")
 
-        self.loss_fcts_val = []
-        for name, w in cf.loss_fcts_val:
-            self.loss_fcts_val += [[getattr(losses, name), w]]
+        self.loss_calculator_val = LossCalculator(cf=cf, stage=VAL, device=self.devices[0])
 
         if is_root():
             config.save(self.cf, epoch=0)


src/weathergen/train/loss_module.py
```diff
@@ -91,11 +98,12 @@ def compute_loss(
...
 
-        assert len(targets) == self.cf.forecast_steps + 1, "Length of targets does not match number of forecast_steps."
-        for fstep in range(len(targets)):
-            for i_strm, (target, strm) in enumerate(
-                zip(targets[fstep], self.cf.streams, strict=False)
-            ):
+        i_batch = 0 # TODO: Iterate over batch dimension here in future
+        for i_strm, strm in enumerate(self.cf.streams):
+            targets = streams_data[i_batch][i_strm].target_tokens[self.cf.forecast_offset:]
+            #assert len(targets) == self.cf.forecast_steps + 1, "Length of targets does not match number of forecast_steps."
```
clessig: Why is this commented out?

src/weathergen/train/loss_module.py
```diff
@@ -91,11 +98,12 @@ def compute_loss(
...
+            
+            for fstep, target in enumerate(targets):
```
clessig: Add more comments to the individual lines, e.g. why 108.

clessig: l114-122: overwriting in VAL case is not very clean: better to have 

```
if TRAIN :
  ...
elif VAL:
```

clessig: l. 125: we should document what the shape of pred at this point is: (ensemble, target_points, target_channels)

src/weathergen/train/loss_module.py
```diff
@@ -123,49 +131,48 @@ def compute_loss(
...
+                                    loss_fct=loss_fct,
+                                    ens=ens
                                 )
-                                # TODO: Implement separate loss weighting functionality over space, channels, time, etc.
-                                val = val + channel_loss_weight[i] * temp
-                                losses_all[strm.name][i, j] += temp.item()
+                                val = val + channel_loss_weight[i_ch] * temp
+                                losses_all[strm.name][i_ch, i_lfct] += temp.item()
                                 ctr_chs += 1
+                        else:
```
clessig: I think the code would be simpler if mask is just identity when tok_spacetime = False and this could be handled completely in `_construct_masks`.

tests/test_loss_module.py
```diff
@@ -110,14 +90,54 @@ def test_loss(cf):
...
+    with torch.autocast(
+        device_type="cuda", dtype=trainer.mixed_precision_dtype, enabled=cf.with_mixed_precision
+    ):
+        preds = trainer.model(trainer.model_params, batch, cf.forecast_offset, cf.forecast_steps)
+
+    preds2 = [[preds[0][0].clone()]]
+
+    ### Original logging preparation
+    # TODO: remove dependency from old trainer.compute_loss() function
+    _, _, _, logging_items = trainer.compute_loss(
```
clessig: This would be high priority since we cannot merge with the old dependency.

clessig (commented):
Great progress, thanks! Two more points:
- Please add doc strings to all functions
- loss_module is not a great name; "module" is an overloaded term in CS and the class is not a module in most definitions. `LossComputer` is not great but descriptive. Open for other suggestions.

src/weathergen/train/loss_module.py
MatKbauer: While this assertion works well for different configurations under `training_mode: "forecast"`, it crashes with
```
training_mode: "masking"
forecast_offset : 0
forecast_steps: 0
```
since `len(targets)=1` and `forecast_offset+forecast_steps=0`. This seems incompatible with autoencoder training. I have put the assertion into an `if training_mode == "masking"`.

src/weathergen/train/loss_calculator.py
```diff
@@ -0,0 +1,309 @@
...
+        This helper function applies a chosen loss function to the masked target and prediction
+        data for a single channel, handling ensemble predictions by calculating mean and standard
+        deviation over the ensemble dimension.
+
+        Args:
+            target: The ground truth target tensor.
+            pred: The prediction tensor, potentially with an ensemble dimension.
+            mask: A boolean mask tensor, indicating which elements to consider for loss computation.
+            i_ch: The index of the channel for which to compute the loss.
+            loss_fct: The specific loss function to apply. It is expected to accept
```
sophie-xhonneux, line 131: so the loss function definition is very rigid here, for instance any latent loss cannot work in this way, indeed many regularisation losses might not work with this

src/weathergen/train/loss_calculator.py
```diff
@@ -0,0 +1,309 @@
...
+                target[mask, i_ch],
+                pred[:, mask, i_ch],
+                pred[:, mask, i_ch].mean(0),
+                (pred[:, mask, i_ch].std(0) if ens else torch.zeros(1, device=pred.device)),
+            )
+        else:
+            # If no valid data under the mask, return 0 to avoid errors and not contribute to loss
+            return 0
+
+    def compute_loss(
```
sophie-xhonneux, line 208: my suggestion would be to rename this to `compute_input_space_loss` and people to register custom functions that all get executed in a wrapper `compute_loss` function.

each register function can come with a list of arguments to recover out of an argument dict `kwargs`? or we extend the loss function through inheritance for instance

config/default_config.yml
```diff
@@ -40,7 +40,7 @@ pred_mlp_adaln: True
 
 # number of steps offset applied to first target window; if set to zero and forecast_steps=0 then
 # one is training an auto-encoder
-forecast_offset : 0
```
tjhunter: just to check: are we ok with changing the default?

config/streams/streams_ocean/fesom.yml
```diff
@@ -9,7 +9,7 @@
 
 FESOM :
   type : fesom
-  filenames : ['fesom_ifs_awi']
+  filenames : ['test4.zarr']
```
tjhunter: needed?

src/weathergen/train/loss_calculator.py
Jubeku: Isn't `src/weathergen/train/loss.py` such a register of custom functions?

src/weathergen/train/loss.py
```diff
@@ -11,6 +11,8 @@
 import numpy as np
 import torch
 
+stat_loss_fcts = ["stats", "kernel_crps"]  # Names of loss functions that need std computed
```
tjhunter: what do you mean by `std computed`?

src/weathergen/train/loss_calculator.py
```diff
@@ -0,0 +1,309 @@
...
+    This provides a structured way to return the primary loss used for optimization,
+    along with detailed per-stream/per-channel/per-loss-function losses for logging,
+    and standard deviations for statistical analysis.
+    """
+
+    # The primary scalar loss value for optimization.
+    loss: Tensor
+    # Dictionaries containing detailed loss values and standard deviation statistics for each
+    # stream, channel, and loss function.
+    losses_all: dict
```
tjhunter: `dict[str, Tensor]` ? Be more precise.

src/weathergen/train/loss_calculator.py
Jubeku: Wondering if we introduce a dead end here or if we can work on this generalization in a new PR?

tjhunter: you mean the KL divergence in the ELBO, in case of a VAE? Since it is coming from an intermediate stage in the model, I would say that we can add it later as an extra parameter.

Thinking about it, we have roughly 3 losses depending on the stage in the model:
- initial (stage 0): regularization terms (no need for samples)
- mid-way: variational terms in the ELBO
- end-to-end: expected empirical risk

src/weathergen/train/loss_calculator.py
tjhunter: I am not a fan of registration because it includes state to think about. Could it be described fully in the input of the class constructor?

src/weathergen/utils/distributed.py
```diff
@@ -33,3 +36,79 @@ def is_root(pg: dist.ProcessGroup | None = None) -> bool:
...
+
+
+def get_world_size() -> int:
+    """
+    Get MPI world size
+
+    Returns:
+        int: world size
+    """
+    if not dist.is_available():
```
tjhunter: there is a `_is_distributed_initialized` below.

src/weathergen/utils/distributed.py
```diff
@@ -33,3 +36,79 @@ def is_root(pg: dist.ProcessGroup | None = None) -> bool:
...
+        return 1
+    if not dist.is_initialized():
+        return 1
+    return dist.get_world_size()
+
+
+def get_rank() -> int:
```
tjhunter: why not passing processgroups? see `is_root` below. I still don't fully grasp why we need them.

@sophie-xhonneux  or Sebastian Hoffmann would probably know better.

src/weathergen/utils/train_logger.py
```diff
@@ -388,6 +410,13 @@ def _key_loss(st_name: str, lf_name: str) -> str:
     return f"stream.{st_name}.loss_{lf_name}.loss_avg"
 
 
+def _key_loss_chn(st_name: str, lf_name: str, ch_name: str) -> str:
+    st_name = _clean_name(st_name)
+    lf_name = _clean_name(lf_name)
```
tjhunter: do not clean the loss or channel names, they should be standard enough.

config/default_config.yml
Jubeku: No, we have to revert to the default configs before merging.

src/weathergen/train/loss_calculator.py
sophie-xhonneux: regularisation terms can very much need samples frequently, e.g. z-loss as needed for stability in multi-modal transformers as here https://arxiv.org/pdf/2405.09818, or representation regularisation like a dispersion loss here https://arxiv.org/abs/2506.09027

src/weathergen/utils/distributed.py
sophie-xhonneux: get_rank is afaik often used for things you only want to do on one gpu, e.g. logging

src/weathergen/train/loss_calculator.py
sophie-xhonneux: I think there should be an easy to compute a loss on the latents given we are doing representation learning and in particular SSL, this will come quite soon I suspect, we can leave it for a future PR I guess, but this is quite restrictice is my point

config/default_config.yml
clessig: This comes from #440. We need to clean up before it's merged. Same below.

src/weathergen/utils/distributed.py
clessig: This is again #440. Wrong place to discuss.

We need this function because if `distributed` is not initialized when we run interactively, all the calls to distributed will fail.

src/weathergen/train/loss_module.py
MatKbauer: Edit: After testing more configurations, this assertion seems not valid in general. It only seems valid for `training_mode: "forecast"` and `forecast_policy: "fixed"`. Furthermore, it should then be
```
if self.cf.training_mode == "forecast" and self.cf.forecast_policy == "fixed":
    assert len(targets) == self.forecast_steps
```
But that seems too constraining.

config/default_config.yml
```diff
@@ -40,12 +40,12 @@ pred_mlp_adaln: True
 
 # number of steps offset applied to first target window; if set to zero and forecast_steps=0 then
 # one is training an auto-encoder
-forecast_offset : 1
+forecast_offset: 1
```
clessig: Can we remove any changes in default_config before merging

src/weathergen/train/loss_calculator.py
clessig: Let's do the changes for latent space loss when we implement it. Otherwise it's too abstract and we will need to change things based on concrete needs.

src/weathergen/train/loss_calculator.py
```diff
@@ -146,7 +146,7 @@ def _compute_loss_with_mask(
             )
         else:
             # If no valid data under the mask, return 0 to avoid errors and not contribute to loss
-            return 0
+            return torch.zeros(1, device=pred.device)
```
clessig: Zeros are dangerous; NaNs are usual what you want to have. Are we sure we average appropriately? E.g.

(3 + 0 + 2) / 3 \neq (3 + 2) /2 

We want the latter when a stream is empty in terms of predictions.

src/weathergen/train/loss_calculator.py
```diff
@@ -0,0 +1,306 @@
...
+
+import weathergen.train.loss as losses
+from weathergen.train.loss import stat_loss_fcts
+from weathergen.utils.train_logger import TRAIN, VAL, Stage
+
+_logger = logging.getLogger(__name__)
+
+
+@dataclasses.dataclass
+class ModelLoss:
```
clessig: Wouldn't `LossValues` be a more accurate and descriptive name?

src/weathergen/train/loss_calculator.py
```diff
@@ -0,0 +1,306 @@
...
+    """
+    A dataclass to encapsulate the various loss components computed by the LossCalculator.
+
+    This provides a structured way to return the primary loss used for optimization,
+    along with detailed per-stream/per-channel/per-loss-function losses for logging,
+    and standard deviations for statistical analysis.
```
clessig: The standard dev's are not for statistical analysis. This is the stdev of ensemble predictions, e.g. when one trains with CRPS and its variants.

src/weathergen/train/loss_calculator.py
```diff
@@ -0,0 +1,306 @@
...
+    This class handles the initialization and application of various loss functions,
+    applies channel-specific weights, constructs masks for missing data, and
+    aggregates losses across different data streams, channels, and forecast steps.
+    It provides both the main loss for backpropagation and detailed loss metrics for logging.
+    """
+
+    def __init__(
+        self,
+        cf: DictConfig,
+        stage: Stage = TRAIN,
```
clessig: Do we need the default args here? It's better to be explicit.

src/weathergen/train/loss_calculator.py
```diff
@@ -0,0 +1,306 @@
...
+
+    @staticmethod
+    def _construct_masks(
+        target_times_raw: np.array, mask_nan: Tensor, tok_spacetime: bool
+    ) -> list[Tensor]:
+        """
+        Constructs a list of boolean masks for target data.
+
+        If 'tok_spacetime' is enabled, masks are generated for unique intermediate time steps
+        within a single forecast step, combined with a NaN mask. Otherwise, a single mask
```
clessig: , and combined

src/weathergen/train/loss_calculator.py
```diff
@@ -0,0 +1,306 @@
...
+        for non-NaN values is returned. This is particularly useful for datasets where targets
```
clessig: This is useful ...

src/weathergen/train/loss_calculator.py
clessig: Let's address it in a subsequent PR. This is *a refactoring PR* and we agreed all new features come based on it.

src/weathergen/train/loss_calculator.py
```diff
@@ -0,0 +1,306 @@
...
+            preds: A nested list of prediction tensors. The outer list represents forecast steps,
+                   the inner list represents streams. Each tensor contains predictions for that
+                   step and stream.
+            streams_data: A nested list representing the input batch data. The outer list is for
+                          batch items, the inner list for streams. Each element provides an object
+                          (e.g., dataclass instance) containing target data and metadata.
+
+        Returns:
+            A ModelLoss dataclass instance containing:
+            - loss: The final aggregated and normalized scalar loss for backpropagation.
```
clessig: "The loss for back-propagation"

src/weathergen/train/loss_calculator.py
```diff
@@ -0,0 +1,306 @@
...
+            - losses_all: A dictionary mapping stream names to a tensor of per-channel and
+                          per-loss-function losses, normalized by non-empty targets/forecast steps.
+            - stddev_all: A dictionary mapping stream names to a tensor of mean standard deviations
+                          of predictions for channels with statistical loss functions, normalized.
+        """
+        ctr_ftarget = 0  # Counter for non-empty forecast steps/targets contributing to overall loss
```
clessig: could we use a better variable name

src/weathergen/train/loss_calculator.py
```diff
@@ -0,0 +1,306 @@
...
+            st.name: torch.zeros(
+                (len(st[str(self.stage) + "_target_channels"]), len(self.loss_fcts)),
+                device=self.device,
+            )
+            for st in self.cf.streams  # No nan here as it's divided so any remaining 0 become nan
+        }  # Create tensor for each stream
+        stddev_all: dict[str, Tensor] = {
+            st.name: torch.zeros(len(stat_loss_fcts), device=self.device) for st in self.cf.streams
+        }  # Create tensor for each stream
+        # assert len(targets) == len(preds) and len(preds) == len(self.cf.streams)
```
clessig: Either remove or uncomment

src/weathergen/train/loss_calculator.py
```diff
@@ -0,0 +1,306 @@
...
+            targets = streams_data[i_batch][i_strm].target_tokens[self.cf.forecast_offset :]
+            for fstep, target in enumerate(targets):
+                pred = preds[fstep][i_strm]
+                # Skip if either target or prediction has no data points
+                if not (target.shape[0] > 0 and pred.shape[0] > 0):
+                    continue
+
+                num_channels = len(strm[str(self.stage) + "_target_channels"])
+
+                # Determine observation and channel loss weights based on the current stage
```
clessig: What's the difference between observation and channel loss weights? Do you mean stream and channel loss weights?

config/default_config.yml
MatKbauer: Gonna revert. How about `forecast_att_dense_rate: 0.25`, which produces a CUDA error #588 when not set to `1.0`?

src/weathergen/train/loss_calculator.py
MatKbauer: Agree -- renamed to LossValues

src/weathergen/train/loss_calculator.py
MatKbauer: Right, added `dict[str, Tensor]`.

src/weathergen/train/loss_calculator.py
MatKbauer: Agreed -> removed the defaults.

config/default_config.yml
clessig: We leave `default_config.yml` as it is. If we want to change it, we open a PR. Parasitic changes in a PR that has nothing to do with the subject just invite trouble.

config/default_config.yml
MatKbauer: Reverted to `default_config.yml` in #414, which is the state of the repository when @Jubeku opened this branch on July 11, with SHA `007ccccc`

src/weathergen/train/loss_calculator.py
MatKbauer: Renamed to `n_trgts`

src/weathergen/train/loss_calculator.py
MatKbauer: Good point. Changed observation to stream.

src/weathergen/train/loss_calculator.py
MatKbauer: Checked this and modified the `val += temp` accumulation to only occur when `temp` is not `NaN`. Normalization was modified accordingly.

PR #534, opened 2025-07-16
Fixed stream name factoring

### LINKED ISSUE #364: Stream names are not friendly with logging systems (mlflow, wandb, etc.)

Sindhu-Vasireddy:
## Description

Removed regex and fixed stream name factoring to include alphanumeric and underscores alone. Run Id ll5g806o on Juwels was tested with stream name `ERA5,_ .AIFS o-96` which was formatted to `ERA5_AIFSo96` in the metrics.json

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #364 

## Code Compatibility

### Dependencies


Sindhu-Vasireddy:
@tjhunter Could you please review this

src/weathergen/utils/train_logger.py
```diff
@@ -377,17 +375,26 @@ def clean_df(df, columns: list[str] | None):
...
+             in the same order and capitalization as they appeared in the input.
+    """
+    return "".join(c for c in s if c.isalnum() or c == "_")
 
 
 def _key_loss(st_name: str, lf_name: str) -> str:
-    st_name = _clean_name(st_name)
-    lf_name = _clean_name(lf_name)
+    st_name = clean_name(st_name)
+    lf_name = clean_name(lf_name)
```
jpolz: I think the loss function (`lf_name`) should remain untouched, right?

tjhunter: yes indeed, this was the issue that led to the revert

src/weathergen/utils/train_logger.py
```diff
@@ -377,17 +375,26 @@ def clean_df(df, columns: list[str] | None):
...
+    Remove all characters from a string except letters, digits, and underscores.
+
+    Args:
+        s (str): The input string.
+
+    Returns:
+        str: A new string containing only alphanumeric characters and underscores,
+             in the same order and capitalization as they appeared in the input.
+    """
+    return "".join(c for c in s if c.isalnum() or c == "_")
```
tjhunter, line 442: to be clear to others:
- ERA5 -> ERA5
- NPP, X-AT_MS -> NPPAT_MS
- a_8 -> a_8
- b-7 -> b7

The reason for deviating from the agreed upon camel case:
- the current names (ERA5) will stay the same : backward compatibility
- it is still readable and is accepted by the logging backends

tjhunter (changes requested):
@Sindhu-Vasireddy thank you for opening again this request. There is still the name to change.

src/weathergen/utils/train_logger.py
Sindhu-Vasireddy: The current clean_name wrapper keeps the alphanumeric and underscore characters so wrapping the lf_name shouldn't have any negative impact. Hence, left the original implementation that way.

src/weathergen/utils/train_logger.py
grassesi: can this be documented somewhere, ideally in a test?

tjhunter (approved):
@Sindhu-Vasireddy I just checked the output metrics and it seems to be working as expected. I am approving this PR, but note it still needs to resolve conflicts against the latest develop. That should be quick to update.

Thanks again for working on this subtle but important feature!

Sindhu-Vasireddy:
@tjhunter I have now pulled in all the new changes and merged. Could you please have a look at your convenience?

src/weathergen/utils/train_logger.py
```diff
@@ -120,12 +120,11 @@ def add_train(
         log_vals += [lr]
 
         for st in self.cf.streams:
-            st_name = st["name"]
+            st_name = clean_name(st["name"])
```
clessig: @tjhunter : This has just been changed in a PR by @kacpnowak . We either would have to use `clean_name()` also in trainer or be careful where we expect which version. It's perhaps a more general question where human-readable form should be used and where the machine readable one.

Sindhu-Vasireddy: @clessig @tjhunter For the scope of this issue, I have kept the human-readable form only in metrics.json and removed it from the key name of the internal dict where I left it to be a machine-readable one, assuming that the broader scope can be addressed in a different issue. 

PS: This PR has all the latest changes fetched and merged.

src/weathergen/utils/train_logger.py
```diff
@@ -429,24 +428,31 @@ def clean_df(df, columns: list[str] | None):
...
+    st_name = clean_name(st_name)
     return f"stream.{st_name}.loss_{lf_name}.loss_avg"
 
 
 def _key_loss_chn(st_name: str, lf_name: str, ch_name: str) -> str:
-    st_name = _clean_name(st_name)
-    lf_name = _clean_name(lf_name)
-    ch_name = _clean_name(ch_name)
+    st_name = clean_name(st_name)
+    ch_name = clean_name(ch_name)
```
tjhunter: to my knowledge, the channels already have "good" names. We do not need to clean the channel names.

src/weathergen/utils/train_logger.py
```diff
@@ -429,24 +428,31 @@ def clean_df(df, columns: list[str] | None):
     return df
 
 
-def _clean_name(n: str) -> str:
-    """Cleans the stream name to only retain alphanumeric characters"""
-    return "".join([c for c in n if c.isalnum()])
+def clean_name(s):
```
tjhunter: do not forget the type annotations: `(s: str) -> str:`

src/weathergen/utils/train_logger.py
tjhunter: I just tried the change and it works as expecte on a mixed stream:
https://dbc-080c3210-c159.cloud.databricks.com/ml/experiments/384213844828345/runs/df95a762f9ab47399814bbab51d78673?o=3342786197435636

kacpnowak: @tjhunter the issue is when there's non-alphanumeric character in the stream name. For example, IFS-FESOM dataset is split into FESOM_NODES, FESOM_ELEM and IFS_ATMO

clessig: > @tjhunter the issue is when there's non-alphanumeric character in the stream name. For example, IFS-FESOM dataset is split into FESOM_NODES, FESOM_ELEM and IFS_ATMO

What do you mean here? FESOM_NODES should never be a dataset name?

kacpnowak: It's a stream name

PR #589, opened 2025-07-24
[DRAFT] Quick fix for #553 NaT from encode_times_target, move offset to before trigs

### LINKED ISSUE #553: Wrong `target_times_raw` in masking mode
### What happened?

When masking mode is on, the `target_times_raw` attribute of the streams data class is corrupted and doesn't return the expected timestamps.

Load a batch item in masking mode, assuming the ERA5 stream.
We should get `int(len_hrs / step_hrs)` unique timestamps which we can get as follows:
```
stream_data = batch[0]
i_batch = 0
i_strm = 0
fstep = 0
t_unique = np.unique(streams_data[i_batch][i_strm].target_times_raw[fstep])
```
See `t_unique` outputs for several example cases below, compared with the desired output in forecast mode.

### Version

develop

Linux

### Relevant log output

```shell
## len_hrs = 6, step_hrs = 6, mode = "masking"
t_unique = array(['NaT'], dtype='datetime64[us]') 
  
## len_hrs = 6, step_hrs = 6, mode = "forecasting"
t_unique  = array(['2021-10-03T00:00:00'], dtype='datetime64[s]') 

## len_hrs = 24, step_hrs = 6, mode = "masking"
t_unique  = array(['2021-11-21T04:00:00.000000', 'NaT'], dtype='datetime64[us]')

## len_hrs = 24, step_hrs = 6, mode = "forecasting"
t_unique  = array(['2021-07-01T12:00:00', '2021-07-01T18:00:00', '2021-07-02T00:00:00', '2021-07-02T06:00:00'], dtype='datetime64[s]')
```

### Accompanying data

_No response_

### Organisation

ECMWF

shmh40:
## Description

Fix to avoid NaT in mode "masking", which originates from adding a 0.5 offset in encode_target_times and leads to NaNs when passed through arc_alpha as arccos is nan for values greater than 1 or less than -1.

It is not clear to me why the offset was 0.5 before, and what the offset should be now...

We could simply add the offset here, e.g. target_times = target_times + 0.X: https://github.com/ecmwf/WeatherGenerator/blob/96e72e9e0693d8263b21abe0109787303a7f7223/src/weathergen/datasets/tokenizer_masking.py#L324

so as not to lead to incorrect target_times_raw (by whatever offset we choose in encode_times_target).

## Type of Change

-   [ ] Documentation update

## Issue Number

Fixes #553 

## Code Compatibility

### Dependencies

### Documentation


src/weathergen/datasets/tokenizer_utils.py
```diff
@@ -81,7 +86,7 @@ def encode_times_target(times, time_win) -> torch.tensor:
 
     # We add + 0.5 as in ERA5 very often we otherwise get 0 as the first time and to prevent too
     # many zeros in the input, where we cannot learn anything we add an offset
-    return time_tensor + 0.5
+    return time_tensor  # + 0.5 remove this offset to avoid breaking trigonometric ranges
```
clessig: See https://github.com/ecmwf/WeatherGenerator/issues/553#issuecomment-3122012904.

shmh40: Ok, updated the PR with this change.

shmh40: tokenizer_forecast seems to have a different structure, with no arc_alpha and not computing the offset as far as I can see - so nothing to do there? There is still encode_times_target for the target times (which will still have the 0.5 offset...) but the target_times_raw are takes directly from times (which looks good, more what we want for masking?).

Jubeku (approved):

PR #594, opened 2025-07-25
Sophiex/dev/latent noise

### LINKED ISSUE #593: Latent noise interpolation

We want to do masked token modelling pre-training with interpolated latent noise following https://arxiv.org/abs/2507.15856

_No response_

_No response_

### Additional context

_No response_

### Organisation

_No response_

sophie-xhonneux:
## Description

Draft for how to implement interpolated noise during masked token modelling pre-training.

Closes #593 

## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Dependencies

### Documentation


config/default_config.yml
```diff
@@ -136,4 +142,4 @@ run_id: ???
 # Parameters for logging/printing in the training loop
 train_log:
   # The period to log metrics (in number of batch steps)
-  log_interval: 20
```
shmh40: I never understand this one

config/default_config.yml
```diff
@@ -17,13 +18,13 @@ ae_local_with_qk_lnorm: True
 
 ae_local_num_queries: 1
 ae_local_queries_per_cell: False
-ae_adapter_num_heads: 16
+ae_adapter_num_heads: 8
```
shmh40: Usual comment for default_config

config/default_config.yml
```diff
@@ -61,6 +62,11 @@ attention_dtype: bf16
 mlp_norm_eps: 1e-5
 norm_eps: 1e-4
 
+kl_weight: 0.0 # 1e-5
+noise_gamma: 2.0
+use_additive_noise: False
+deterministic_latents: False
+
```
shmh40, line 69: Looks good

config/default_config_atos.yml
```diff
@@ -0,0 +1,129 @@
...
+embed_local_coords: True
+embed_centroids_local_coords: False
+embed_size_centroids: 0
+embed_unembed_mode: "block"
+
+target_cell_local_prediction: True
+
+ae_local_dim_embed: 1024
+ae_local_num_blocks: 2
+ae_local_num_heads: 16
```
shmh40, line 16: Remove

config/default_config_leonardo.yml
```diff
@@ -0,0 +1,129 @@
+streams_directory: "./config/streams/streams_anemoi/"
```
shmh40: "

config/default_config_atos.yml
```diff
@@ -0,0 +1,129 @@
+streams_directory: "./config/streams/streams_anemoi/"
```
shmh40: Usual, remove before merge

config/streams/streams_anemoi/era5.yml
```diff
@@ -11,7 +11,7 @@ ERA5 :
   type : anemoi
```
shmh40: "

src/weathergen/model/attention.py
```diff
@@ -153,7 +154,8 @@ def sparsity_mask(score, b, h, q_idx, kv_idx):
 
             return flex_attention(qs, ks, vs, score_mod=sparsity_mask)
 
-        self.compiled_flex_attention = torch.compile(att, dynamic=False)
+        self.compiled_flex_attention = torch.compile(att, dynamic=False, mode="max-autotune")
```
shmh40: Should these changes be in a different PR? What is happening here - was this to make the flex attention work on Leonardo or whatever?

src/weathergen/model/model.py
```diff
@@ -202,6 +203,14 @@ def create(self) -> "Model":
         # local assimilation engine
         self.ae_local_blocks = LocalAssimilationEngine(cf).create()
 
+        if cf.kl_weight > 0.0:
```
shmh40: So the entry point to use the latent noise is kl_weight > 0.0 in the config. I am happy with this, but do you think it would be clearer to people to have a True/False in the config? I think not necessary for now, but just mentioning for future.

use_latent_noise : true
kl_weight : 1e-5
noise_gamme : whatever

sophie-xhonneux: I wanted to avoid yet another config value, but happy to change that. What do you think/prefer?

src/weathergen/model/attention.py
sophie-xhonneux: Will revert, this was me debugging things

src/weathergen/model/model.py
```diff
@@ -356,7 +365,7 @@ def create(self) -> "Model":
             )
             self.pred_heads.append(
                 EnsPredictionHead(
-                    dims_embed[-1],
+                    dims_embed[0],
```
shmh40: This change makes no difference right?

src/weathergen/model/model.py
```diff
@@ -613,8 +622,15 @@ def assimilate_local(
         )
 
         # # local assimilation model
-        # for block in self.ae_local_blocks :
-        #   tokens = checkpoint( block, tokens, cell_lens, use_reentrant=False)
+        # for block in self.ae_local_blocks:
```
shmh40, line 636: Remove

src/weathergen/model/parametrised_prob_dist.py
```diff
@@ -0,0 +1,108 @@
...
+        self.mean_and_var = nn.Linear(dim, 2 * dim, bias=False)
+
+    def interpolate_with_noise(self, z, sampling=False, noise_level=-1):
+        self.diag_gaussian.reset_parameters(self.mean_and_var(z))
+        z_latents = self.diag_gaussian.sample() if sampling else self.diag_gaussian.mean
+
+        if self.training and self.gamma > 0.0:
+            device = z_latents.device
+            s = z_latents.shape
+            bsz = 1
```
shmh40: Sorry, can you call this batch_size. 

And can we just get the batch_size here straight from the default config to address your TODO below?

shmh40: Otherwise I think this looks good

src/weathergen/train/trainer.py
```diff
@@ -50,7 +51,18 @@ def init(
         self,
         cf: Config,
     ):
-        self.cf = cf
+        self.cf = OmegaConf.merge(
```
shmh40, line 56: Sorry, maybe I am being stupid but why are we doing this? Can't this stuff come directly from the config now anyway?

shmh40: I guess it is left in to make sure it still works with old configs?

src/weathergen/model/attention.py
```diff
@@ -94,7 +94,8 @@ def forward(self, x, x_lens, ada_ln_aux=None):
             dropout_p=self.dropout_rate,
         )
 
-        out = self.proj_out(outs.flatten(-2, -1))
+        out = self.dropout(self.proj_out(outs.flatten(-2, -1)))
```
shmh40: Another PR?

src/weathergen/model/model.py
```diff
@@ -652,6 +669,15 @@ def assimilate_local(
             for block in self.ae_local_blocks:
                 tokens_c = checkpoint(block, tokens_c, cell_lens_c, use_reentrant=False)
 
+            if self.cf.kl_weight > 0.0:
```
shmh40: Despite what I said before, I am actually not sure where this should happen. I have the sense that doing it for the global latents has some neat properties, but it could be local, or both. Do you have strong views on this?

src/weathergen/train/trainer.py
```diff
@@ -466,11 +481,16 @@ def train(self, epoch):
                 dtype=self.mixed_precision_dtype,
                 enabled=cf.with_mixed_precision,
             ):
-                preds = self.ddp_model(self.model_params, batch, cf.forecast_offset, forecast_steps)
+                preds, posteriors = self.ddp_model(
```
shmh40, line 503: It should probably happen in another PR, but can we start to think about what we need to start generating an ensemble with the sampled latents and start training with CRPS? 

We need to sample the differently noised latents, and pass each one through the global and then the decoders, and then store with [sample_size, preds], then in the loss all these samples are used to compute the CRPS against the true target. I don't think is very difficult? Can you think of where this would be tricky? It is slightly different task, as this should encourage us to learn an appropriate std. dev. for the noised latents, rather than focussing on just making the task harder, but I think the two could work in tandem (although perhaps this is not as neat as it should be, perhaps there should be a separation of the tasks).

src/weathergen/model/model.py
sophie-xhonneux: -1 is more correct

src/weathergen/model/model.py
sophie-xhonneux: Yea happy to discuss this, I think more broadly, I want a clearer delineation between encoder and decoder. Conceptually, to me the encoder should not write to the masked tokens

src/weathergen/model/model.py
shmh40: I think it is fine for now, likely just a few users for the next few weeks and you are right about config clutter!

src/weathergen/model/parametrised_prob_dist.py
sophie-xhonneux: I think so much of the code relies on batchsize 1 in a number of places and I don't think this will change in the near future, given we are planning for more datasets not fewer, that it isn't worth the time, especially given that I don't think the rest of the code works with batchsize >1

src/weathergen/train/trainer.py
sophie-xhonneux: yes backwards compatibility

src/weathergen/train/trainer.py
sophie-xhonneux: I think we (in this project) conceptually muddy a few concepts. The pre-training of latents is different from an autoencoder which again differs from forecasting, I think it is worth having a longer discussion about what we want to learn when.

shmh40: Agreed, sounds good

src/weathergen/model/model.py
sophie-xhonneux: I believe this has always been there for documentation purposes

clessig: yes

shmh40 (commented):
Looks good to me if configs are reverted as expected, previously tested on Leonardo. I have not evaluated the performance through to scores though.

config/default_config.yml
```diff
@@ -17,7 +17,7 @@ ae_local_with_qk_lnorm: True
 
 ae_local_num_queries: 1
 ae_local_queries_per_cell: False
-ae_adapter_num_heads: 8
+ae_adapter_num_heads: 16
```
clessig: Is this change to the default config necessary?

config/default_config.yml
```diff
@@ -23,7 +23,7 @@ ae_adapter_with_qk_lnorm: True
 ae_adapter_with_residual: True
 ae_adapter_dropout_rate: 0.1
 
-ae_global_dim_embed: 2048
+ae_global_dim_embed: 2048
```
clessig: Is this ruffed? There's an extra whitespace

config/default_config.yml
```diff
@@ -61,6 +61,11 @@ attention_dtype: bf16
 mlp_norm_eps: 1e-5
 norm_eps: 1e-4
 
+kl_weight: 0.0 # 1e-5
+noise_gamma: 2.0
```
clessig: Could we add a common prefix to these parameters so that it's clear which part of the model/training they affect (so something like `ae_global_?`

config/streams/era5_1deg/era5.yml
```diff
@@ -30,7 +30,7 @@ ERA5 :
   target_readout :
     type : 'obs_value'  # token or obs_value
     num_layers : 2
-    num_heads : 4
+    num_heads : 8
```
clessig: Does this change really fit this PR?

src/weathergen/model/attention.py
```diff
@@ -154,6 +155,7 @@ def sparsity_mask(score, b, h, q_idx, kv_idx):
             return flex_attention(qs, ks, vs, score_mod=sparsity_mask)
 
         self.compiled_flex_attention = torch.compile(att, dynamic=False)
+        # self.compiled_flex_attention = flex_attention
```
clessig: Why do the lines appear here as new? In any case, I think they should be removed.

src/weathergen/model/parametrised_prob_dist.py
```diff
@@ -0,0 +1,108 @@
...
+        return 0.5 * torch.sum(
+            logtwopi + self.logvar + torch.pow(sample - self.mean, 2) / self.var,
+            dim=dims or self.sum_dims,
+        )
+
+    def mode(self):
+        return self.mean
+
+
+class InterpolatedLatents(nn.Module):
```
clessig: The class implements/performs the interpolation so `LatentInterpolator` seems like a better name?

src/weathergen/model/parametrised_prob_dist.py
```diff
@@ -0,0 +1,108 @@
...
+        self.mean_and_var = nn.Linear(dim, 2 * dim, bias=False)
+
+    def interpolate_with_noise(self, z, sampling=False, noise_level=-1):
+        self.diag_gaussian.reset_parameters(self.mean_and_var(z))
+        z_latents = self.diag_gaussian.sample() if sampling else self.diag_gaussian.mean
+
+        if self.training and self.gamma > 0.0:
+            device = z_latents.device
+            s = z_latents.shape
+            batch_size = 1
```
clessig: We can use the code in model.py in l.769 to get the batch size. No one will remember it's there and we should eventually enable batch_size>1 again

src/weathergen/train/trainer.py
clessig: The probabilistic aspect of the autoencoder should learn the data uncertainty (within and across datasets). The forecasting engine should learn the dynamic uncertainty. But let's discuss how to do this best.

config/streams/era5_1deg/era5.yml
sophie-xhonneux: I got flash attention crashes at some point because it expects at least 8 heads, hence why

clessig: I just tested with `num_heads: 4` and it worked. Was the problem on clariden?

sophie-xhonneux: it was in some combination of the prediction heads and  flash attn and other things, not sure which anymore, but will revert this

src/weathergen/model/parametrised_prob_dist.py
sophie-xhonneux: still given how assimilate_local works at the moment dealing with batch_size > 1 is so outrageously complicated that until we need it I wouldn't bother implementing it

src/weathergen/train/trainer.py
sophie-xhonneux: discussed synchronously

src/weathergen/model/parametrised_prob_dist.py
sophie-xhonneux: Sure

Issue #595, opened 2025-07-25
Hackathon script


tjhunter:
Put together a first hackathon demo:
- load metrics
- finetune a model

Assumed to work on juwels-booster only for the time being.

clessig:
I would suggest:

1. Load model, run inference, run evaluatoin and show generated plots
2. Load model, fine-tune for a limited amount of time, run inference on fine-tuned prediction head, run evaluatoin and show generated plots

PR #598, opened 2025-07-25
[595] Changes for running a notebook script 

tjhunter:
## Description

Small changes necessary to run a notebook with the latest version of the code.

Related to #595


src/weathergen/train/trainer.py
```diff
@@ -77,6 +77,24 @@ def init(
         self.init_perf_monitoring()
         self.train_logger = TrainLogger(cf, config.get_path_run(self.cf))
 
+    def init_inference(self, cf: Config, run_id_trained: str, epoch: int, msds: MultiStreamDataSampler):
```
clessig: For inference, we want to write the config since we can overwrite options and we need traceability which options were used to generate output.

tjhunter: fair enough. this function makes sense if the model is purely read only, which I was initially doing. since we are not going that route for the tutorial, I will remove this function.

packages/evaluate/src/weathergen/evaluate/score.py
```diff
@@ -14,7 +14,7 @@
 import numpy as np
 import pandas as pd
 import xarray as xr
-from score_utils import to_list
+from weathergen.evaluate.score_utils import to_list
```
tjhunter, line 18: @iluise  @mlangguth89  : with python, you should always put the full path. it was working before in the script but for the wrong reasons.

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -25,23 +25,8 @@
 
 _logger = logging.getLogger(__name__)
 
-if __name__ == "__main__":
```
tjhunter, line 28: this is just moving the content of the main script into a function, so that it can be called from a notebook

packages/evaluate/src/weathergen/evaluate/score.py
mlangguth89: @iluise @clessig @tjhunter 
I agree that this is a cleaner way to import packages. However, it produces a nightmare of caching issues with uv.
In particular, any changes to the imported scripts remain ineffective since `uv` cashes the local path dependencies by creating a copy in its archive-v0 directory, and then symlinking that copy into the temporary virtual environment. This happens even if one deletes the temporary virtual environment. 
Example:
I created a new virtual environment at 18:08. Here, `uv` creates links to `archive-v0` as seen below
```
[langguth1@jwlogin21 ~]$ ls -lrt /p/project1/hclimrep/langguth1/uv_cache/environments-v2/plot-inference-dedffc9799a552e8/lib/python3.12/site-packages/weathergen/evaluate/
total 4
lrwxrwxrwx 1 langguth1 hclimrep  107 Aug  5 18:08 score_utils.py -> /p/project1/hclimrep/langguth1/uv_cache/archive-v0/KW0rJSOvkCKdZp5Nh746l/weathergen/evaluate/score_utils.py
lrwxrwxrwx 1 langguth1 hclimrep  104 Aug  5 18:08 __init__.py -> /p/project1/hclimrep/langguth1/uv_cache/archive-v0/KW0rJSOvkCKdZp5Nh746l/weathergen/evaluate/__init__.py
lrwxrwxrwx 1 langguth1 hclimrep  101 Aug  5 18:08 utils.py -> /p/project1/hclimrep/langguth1/uv_cache/archive-v0/KW0rJSOvkCKdZp5Nh746l/weathergen/evaluate/utils.py
```
These symlinks may point to outdated code. In my case, code that was 1 1/2 hours older
```
[langguth1@jwlogin21 ~]$ ls -lrt /p/project1/hclimrep/langguth1/uv_cache/archive-v0/KW0rJSOvkCKdZp5Nh746l
total 1
drwxr-sr-x 3 langguth1 hclimrep 4096 Aug  5 16:24 weathergen
drwxr-sr-x 2 langguth1 hclimrep 4096 Aug  5 16:24 weathergen_evaluate-0.1.0.dist-info
```
Any changes to, for instance `utils.py`, remain therefore ineffective.
I have not found a reliable way to circumvent this, because `uv` linked magically to some other archived code when deleting ` /p/project1/hclimrep/langguth1/uv_cache/archive-v0/KW0rJSOvkCKdZp5Nh746l`

PR #625, opened 2025-07-29
Mlangguth/develop/issue 586

### LINKED ISSUE #586: Map plots: point plot size should depend on the resolution

Currently, the point plot size for maps is constant. This can hide resolution dependent effects. For example:

N320: 

O96: 

A lot of the N320 details are hidden through the large point size

The point size for map plots should be configurable or automatically determined.

_No response_

### Additional context

_No response_

### Organisation

_No response_

mlangguth89:
## Description

The option to configure the marker size as well as the marker type has been introduced for the map plots.
Additionally, the marker size can also be scaled based on the latitude of the plotted data points.

Furthermore, the doc-strings in `utils.py` and `plotter.py` of the `evaluate`-package have been updated to follow the code standard.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #586 

## Code Compatibility

The newly introduced option to parse config-parameters to the map-plots is backward compatible.
Thus, both works:
```
[...]
    plot_maps: 
[...]
```
which triggers default behaviour (standard marker size is set), and
```
    plot_maps: true
        marker_size: 0.5
[...]
```
which sets the `marker_size` in the scatter plot to 0.5.

### Dependencies

### Documentation

## Additional Notes

Note that it is also now posible to omit `plot_maps` and `plot_histograms`. 
With the older code version, an error was raised due to these lines: [here](https://github.com/ecmwf/WeatherGenerator/blob/d19f08bdd6ef995451fd659777a9c0e6d59962f1/packages/evaluate/src/weathergen/evaluate/utils.py#L282) and [here](https://github.com/ecmwf/WeatherGenerator/blob/d19f08bdd6ef995451fd659777a9c0e6d59962f1/packages/evaluate/src/weathergen/evaluate/utils.py#L288).

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/plotter.py
+++ b/packages/evaluate/src/weathergen/evaluate/plotter.py
@@ -281,7 +281,7 @@ class Plotter:
             ]
             name = "_".join(filter(None, parts))
             fname = f"{self.out_plot_dir.joinpath(name)}.{self.image_format}"
-            _logger.info(f"Saving map to {fname}")
+            _logger.debug(f"Saving map to {fname}")
             plt.savefig(fname)
             plt.close()
             plot_names.append(name)
--- a/packages/evaluate/src/weathergen/evaluate/plotter.py
+++ b/packages/evaluate/src/weathergen/evaluate/plotter.py
@@ -245,10 +245,9 @@ class Plotter:
             ax.coastlines()
             da = self.select_from_da(data, select_var).compute()
 
+            marker_size = marker_size_base
             if scale_marker_size:
-                marker_size = (marker_size_base + 1.0) * np.cos(np.radians(da["lat"]))
-            else:
-                marker_size = marker_size_base
+                marker_size = (marker_size + 1.0) * np.cos(np.radians(da["lat"]))
 
             scatter_plt = ax.scatter(
                 da["lon"],
@@ -281,7 +280,9 @@ class Plotter:
                 str(self.fstep).zfill(3),
             ]
             name = "_".join(filter(None, parts))
-            plt.savefig(f"{self.out_plot_dir.joinpath(name)}.{self.image_format}")
+            fname = f"{self.out_plot_dir.joinpath(name)}.{self.image_format}"
+            _logger.info(f"Saving map to {fname}")
+            plt.savefig(fname)
             plt.close()
             plot_names.append(name)
 
@@ -375,7 +376,7 @@ class LinePlots:
         x_dim:
             Dimension to be used for the x-axis. The code will average over all other dimensions.
         y_dim:
-            Name of the dimension to be used for the y-axis (default is "value")
+            Name of the dimension to be used for the y-axis.
         print_summary:
             If True, print a summary of the values from the graph.
         """
@@ -426,3 +427,49 @@ class LinePlots:
         name = "_".join(filter(None, parts))
         plt.savefig(f"{self.out_plot_dir.joinpath(name)}.{self.image_format}")
         plt.close()
+
+
+class DefaultMarkerSize:
+    """
+    Utility class for managing default configuration values, such as marker sizes
+    for various data streams.
+    """
+
+    _marker_size_stream = {
+        "era5": 2.5,
+        "imerg": 0.25,
+        "cerra": 0.1,
+    }
+
+    _default_marker_size = 0.5
+
+    @classmethod
+    def get_marker_size(cls, stream_name: str) -> float:
+        """
+        Get the default marker size for a given stream name.
+
+        Parameters
+        ----------
+        stream_name : str
+            The name of the stream.
+
+        Returns
+        -------
+        float
+            The default marker size for the stream.
+        """
+        return cls._marker_size_stream.get(
+            stream_name.lower(), cls._default_marker_size
+        )
+
+    @classmethod
+    def list_streams(cls):
+        """
+        List all streams with defined marker sizes.
+
+        Returns
+        -------
+        list[str]
+            List of stream names.
+        """
+        return list(cls._marker_size_stream.keys())
--- a/packages/evaluate/src/weathergen/evaluate/plotter.py
+++ b/packages/evaluate/src/weathergen/evaluate/plotter.py
@@ -23,8 +23,13 @@ class Plotter:
     def __init__(self, cfg: dict, model_id: str = ""):
         """
         Initialize the Plotter class.
-        :param cfg: config from the yaml file
-        :param model_id: if a model_id is given, the output will be saved in a folder called as the model_id
+
+        Parameters
+        ----------
+        cfg: 
+            Configuration dictionary containing all information for the plotting.
+        model_id: 
+            If a model_id is given, the output will be saved in a folder called as the model_id.
         """
 
         self.cfg = cfg
@@ -49,7 +54,14 @@ class Plotter:
     def update_data_selection(self, select: dict):
         """
         Set the selection for the plots. This will be used to filter the data for plotting.
-        :param select: dictionary containing the selection parameters
+
+        Parameters
+        ----------
+        select:
+            Dictionary containing the selection criteria. Expected keys are:
+                - "sample": Sample identifier
+                - "stream": Stream identifier
+                - "forecast_step": Forecast step identifier 
         """
         self.select = select
 
@@ -78,9 +90,7 @@ class Plotter:
 
     def clean_data_selection(self):
         """
-        :param sample: sample name
-        :param stream: stream name
-        :param fstep: forecasting step
+        Clean the data selection by resetting all selected values.
         """
         self.sample = None
         self.stream = None
@@ -91,9 +101,17 @@ class Plotter:
     def select_from_da(self, da: xr.DataArray, selection: dict) -> xr.DataArray:
         """
         Select data from an xarray DataArray based on given selectors.
-        :param da: xarray DataArray to select data from.
-        :param selection: Dictionary of selectors where keys are coordinate names and values are the values to select.
-        :return: xarray DataArray with selected data.
+
+        Parameters
+        ----------
+        da:
+            xarray DataArray to select data from.
+        selection: 
+            Dictionary of selectors where keys are coordinate names and values are the values to select.
+        
+        Returns
+        -------
+            xarray DataArray with selected data.
         """
         for key, value in selection.items():
             if key in da.coords and key not in da.dims:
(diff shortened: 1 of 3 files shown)


packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -166,16 +195,45 @@ def histogram(
...
+            Selection to be applied to the DataArray
+        tag: str
+            Any tag you want to add to the plot
+        map_kwargs: dict
+            Additional keyword arguments for the map.
+            Known keys are:
+                - marker_size: base size of the marker (default is 1)
+                - scale_marker_size: if True, the marker size will be scaled based on latitude (default is False)
+                - marker: marker style (default is 'o')
+            Unknown keys will be passed to the scatter plot function.
```
iluise, line 226: does it mean that in practice we already support all other plotting adjustments supported by plt? 
Can you clarify it and change known and Unknown keys into custom keys and "default plt keys" or something similar?

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -187,13 +245,21 @@ def map(
             ax.coastlines()
             da = self.select_from_da(data, select_var).compute()
 
+            if scale_marker_size:
+                marker_size = (marker_size_base + 1.0) * np.cos(np.radians(da["lat"]))
+            else:
```
iluise:
```
marker_size = marker size_base
if scale_marker_size:
      marker_size = (marker_size + 1.0) * np.cos(np.radians(da["lat"]))
```

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -289,12 +363,21 @@ def plot(
...
+        Parameters
+        ----------
+        data:
+            DataArray or list of DataArrays to be plotted
+        labels:
+            Label or list of labels for each dataset
+        tag:
+            Tag to be added to the plot title and filename
+        x_dim:
+            Dimension to be used for the x-axis. The code will average over all other dimensions.
```
iluise, line 377: misses the default value here. Which is the convention for the Defaults? In any case I would make it consistent with line 378 and in all other places, either adding or removing the defaults

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -289,12 +363,21 @@ def plot(
...
+        y_dim:
+            Name of the dimension to be used for the y-axis (default is "value")
+        print_summary:
+            If True, print a summary of the values from the graph.
```
iluise, line 381: also here missing the default

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -247,6 +282,36 @@ def plot_data(cfg: str, run_id: str, stream: str, stream_dict: dict) -> list[str
     plot_fsteps = plot_settings.get("forecast_step", None)
     plot_chs = stream_dict.get("channels")
 
+    # Check if maps should be plotted and handle configuration if provided
+    maps_config = {}
+    if hasattr(plot_settings, "plot_maps"):
```
iluise: I tested it and for me the syntax: 

```
plot_maps: true
            marker_size: 0.5
```
doesn't work. Omegaconf complained and I had to use 


```
plot_maps:
            marker_size: 0.5
```

and the following syntax:

```
plot_maps = plot_settings.get("plot_maps", False)
if not (isinstance(plot_maps, oc.dictconfig.DictConfig) or isinstance(plot_maps, bool)):
    raise TypeError( "plot_maps must be a boolean or a dictionary with configuration.")

maps_config = plot_maps if  isinstance(plot_maps, dict) else {}
```

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -247,6 +282,36 @@ def plot_data(cfg: str, run_id: str, stream: str, stream_dict: dict) -> list[str
...
+            plot_maps = True
+            maps_config = plot_settings.plot_maps
+        else:
+            raise TypeError(
+                "plot_maps must be a boolean or a dictionary with configuration."
+            )
+    else:
+        plot_maps = False
+
+    if not hasattr(maps_config, "marker_size"):
```
iluise, line 299: if we want to support other keys maybe we can create a function to fill the maps config with the default values?

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -247,6 +282,36 @@ def plot_data(cfg: str, run_id: str, stream: str, stream_dict: dict) -> list[str
...
+        # Set default marker size if not specified in maps_config
+        maps_config["marker_size"] = get_default_markersize(stream)
+
+    # Check if histograms should be plotted
+    if hasattr(plot_settings, "plot_histograms"):
```
iluise: same as above to simplify the syntax?

```
plot_histograms = plot_settings.get("plot_histograms", False)
``` 

I don't see that the plotting options are passed to the histogram though. So histogram can only be bool, right?

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -461,6 +537,31 @@ def plot_summary(cfg: dict, scores_dict: dict, print_summary: bool):
                     )
 
 
+def get_default_markersize(stream_name: str) -> float:
```
iluise: minor: wondering if these values should actually go into `plotter.py` in a dedicated class e.g. DefaultSettings or so

kctezcan:
I think the marker size is also a parameter of the area that is shown. If we zoom into an area of Europe, say for CERRA, then we will have to readjust the marker size. I was thinking of calculating it on the go based on the relative size of the shown area to the full global area

mlangguth89:
> I think the marker size is also a parameter of the area that is shown. If we zoom into an area of Europe, say for CERRA, then we will have to readjust the marker size. I was thinking of calculating it on the go based on the relative size of the shown area to the full global area

No, the marker size determines its size in terms of pixel/ponts (specifically the squared number of points). So it depends on the size of the plot area (in inches) and the dpi (pixel resolution) of the plot. I spent a few hours figuring out how to do it, but it's fairly complicated since it requires inferring the spatial resoultion from the data which is highly challenging for tabular data. Please have a look on what is documented in the related issue #586.

For now: **Please let's focus on what is implemented in this PR**...and that's just to customize manually the marker size.

packages/evaluate/src/weathergen/evaluate/utils.py
mlangguth89: No strong feelings about this as well. Might be neat to have it this way :+1:

packages/evaluate/src/weathergen/evaluate/utils.py
iluise: it still complains also if I use 

```
plot_maps:
            marker_size: 0.5
```

for me instance is not a dictionary but a `oc.dictconfig.DictConfig`, so I had to use the syntax above :/

packages/evaluate/src/weathergen/evaluate/plotter.py
mlangguth89: Yes, with my PR you could pass through all known parameters to `ax.scatter`. An exception holds for those parameters that we already set, e.g. `s` for the marker size. If you pass such a parameter through `map_kwargs`, Python will throw an error, because you try to set this parameter two times. 
That's why you need `marker_size_base = map_kwargs_save.pop("marker_size", 1)` which filters out the eventually duplicated parameter.

packages/evaluate/src/weathergen/evaluate/utils.py
mlangguth89: Yes, if we have more than a small number of parameters, then we should have it like this (follow-up work).

packages/evaluate/src/weathergen/evaluate/utils.py
mlangguth89: This PR is only for the marker size of the map-plots (see #586). If we want to enable configuring the histograms, we should open a separate issue.

packages/evaluate/src/weathergen/evaluate/utils.py
mlangguth89: It should be more clearly documented somewhere.
My intention wa sto make the code backward compatible, i.e. the old config with
```
plot_maps: true
```
should work as well as the new one
```
plot_maps:
    marker_size: 1.
```
I will adapt the documentation on our private Wiki, when the PR is merged.

packages/evaluate/src/weathergen/evaluate/plotter.py
mlangguth89: see above

packages/evaluate/src/weathergen/evaluate/plotter.py
mlangguth89: The default is visible from the type and default-indication in the signature of the function.
@tjhunter pointed out to me that there is no need to repeat this in the doc-string.

I've made it consistent. Thanks for the hint :-)

packages/evaluate/src/weathergen/evaluate/utils.py
mlangguth89: I'm implementing a class under `plotter.py` in my recent update to this PR.

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -215,7 +280,9 @@ def map(
                 str(self.fstep).zfill(3),
             ]
             name = "_".join(filter(None, parts))
-            plt.savefig(f"{self.out_plot_dir.joinpath(name)}.{self.image_format}")
+            fname = f"{self.out_plot_dir.joinpath(name)}.{self.image_format}"
+            _logger.info(f"Saving map to {fname}")
```
iluise: This can be super verbose. Can we add an option for this or put it to DEBUG level?

packages/evaluate/src/weathergen/evaluate/utils.py
iluise: yes, the histogram can not be a dictionary though, so I think we should do:
```
plot_histograms = plot_settings.get("plot_histograms", False)
if not isinstance(plot_histograms, bool):
        raise TypeError(
            "plot_histograms must be a boolean."
        )
```

packages/evaluate/src/weathergen/evaluate/plotter.py
mlangguth89: Done!

iluise (approved):

Issue #608, opened 2025-07-28
train_logger uses fixed path "./results"


kctezcan:
### What happened?

The line here assumes the results path as ./results instead of taking it from the config:
https://github.com/ecmwf/WeatherGenerator/blob/bb7e26968260d9d3522caa450b5f420432a31598/src/weathergen/utils/train_logger.py#L69

This should come from the config paths.

### What are the steps to reproduce the bug?

run training with some other result_dir folder and don't have a folder results in your project home folder. 

### Version

0.1

### Platform (OS and architecture)

linux

### Relevant log output

```shell

```

PR #631, opened 2025-07-30
using config results path instead of fixed path

kctezcan:
## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/608

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/utils/train_logger.py
+++ b/src/weathergen/utils/train_logger.py
@@ -87,7 +87,9 @@ class TrainLogger:
         # TODO: performance: we repeatedly open the file for each call. Better for multiprocessing
         # but we can probably do better and rely for example on the logging module.
 
-        metrics_path = get_train_metrics_path(base_path=Path(self.cf.run_path), run_id=self.cf.run_id)
+        metrics_path = get_train_metrics_path(
+            base_path=Path(self.cf.run_path), run_id=self.cf.run_id
+        )
         with open(metrics_path, "ab") as f:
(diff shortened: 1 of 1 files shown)


src/weathergen/utils/train_logger.py
```diff
@@ -87,7 +87,7 @@ def log_metrics(self, stage: Stage, metrics: dict[str, float]) -> None:
         # TODO: performance: we repeatedly open the file for each call. Better for multiprocessing
         # but we can probably do better and rely for example on the logging module.
 
-        metrics_path = get_train_metrics_path(base_path=Path("results"), run_id=self.cf.run_id)
+        metrics_path = get_train_metrics_path(base_path=Path(self.cf.run_path), run_id=self.cf.run_id)
```
mlangguth89: One could also use the class attribute self.path_run.parent directly.
I'm not 100% sure if this would be more consistent indeed.

clessig:
@kctezcan : many thanks for picking this up. `Path(self.cf.run_path)` is the result directory, i.e. the file will end up in the same location? Otherwise, we need to adapt `plot_train` and potentially other things.

src/weathergen/utils/train_logger.py
kctezcan: Yeah, this takes it directly from the private config, where it is prescribed, so I thought closer to the source...

sophie-xhonneux (approved):

Issue #627, opened 2025-07-30
Align color scale across forecast steps


clessig:
### Is your feature request related to a problem? Please describe.

Currently, the color scale for each forecasts step is determined independently, e.g. 

<img width="1920" height="1440" alt="Image" src="https://github.com/user-attachments/assets/2286e828-7f3f-4f5b-a150-fb2eb4f13aa3" />

<img width="1920" height="1440" alt="Image" src="https://github.com/user-attachments/assets/9ad2a6f5-08c9-4b3e-8323-da7d59e37d67" />

### Describe the solution you'd like

Ensure a consistent color scale across forecast step using the max values across the forecast sequence.

Related to #623

Issue #623, opened 2025-07-29
Align color scale across multiple streams


clessig:
### Is your feature request related to a problem? Please describe.

It's currently difficult to compare predictions of the same field for different streams since the color scaling is not aligned, e.g.:

<img width="1920" height="1440" alt="Image" src="https://github.com/user-attachments/assets/77979af0-006c-454d-a904-7c4f775b333e" />

<img width="1920" height="1440" alt="Image" src="https://github.com/user-attachments/assets/558b6031-76e6-461e-9378-ed87e3414b85" />

### Describe the solution you'd like

Allow for consistent color scaling across streams. The easiest solution seems to be to allow that one specifies a parent stream for the plotting and domain extent and color scaling are inherited from this stream.

CC @mlangguth89 @iluise

PR #638, opened 2025-07-30
[DRAFT] introduce self.ranges dictionary for consistent plotting ranges across streams and fsteps

iluise:
## Description

Need to merge the other 2 PRs before this, but so you can start reviewing it and comment it. 

self.ranges is a dictionary that stores the vmin and vmax values of all variables. 
It is automatically filled after the first plot of that variable is called and the ranges are kept across all the run_id/streams/sftep combinations of that variable. 

- if variables from 2 different streams have the same name they will have the same range. 
- The order matters as the range is filled by the first plot

Alternative: 
one can use a custom "plotting_opts" dictionary from the config. If no ranges is found, the code falls back to automatically filling from the first plot. 

```
run_ids:
     my_run_id:
           evaluation: 
                blah
           plotting:
                 blah

plotting_opts:
    2t:
      vmin: 0.2
      vmax: 0.3
```

## Type of Change

-   [ ] Documentation update

## Issue Number

closes https://github.com/ecmwf/WeatherGenerator/issues/623
closes https://github.com/ecmwf/WeatherGenerator/issues/627

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/plot_inference.py
+++ b/packages/evaluate/src/weathergen/evaluate/plot_inference.py
@@ -113,6 +113,6 @@ if __name__ == "__main__":
 
 
 # plot summary
-if cfg.summary_plots:
+if scores_dict and cfg.summary_plots:
     _logger.info("Started creating summary plots..")
     plot_summary(cfg, scores_dict, print_summary=cfg.print_summary)
--- a/packages/evaluate/src/weathergen/evaluate/plot_inference.py
+++ b/packages/evaluate/src/weathergen/evaluate/plot_inference.py
@@ -58,8 +58,11 @@ if __name__ == "__main__":
     # to get a structure like: scores_dict[metric][stream][run_id] = plot
     scores_dict = defaultdict(lambda: defaultdict(dict))
 
+    # common plotting options across streams and run_ids. e.g. vertical ranges. 
+    plotting_opts = {}
+
     for run_id, run in runs.items():
-        plotter = Plotter(cfg, run_id)
+
         _logger.info(f"RUN {run_id}: Getting data...")
 
         streams = run["streams"].keys()
@@ -71,7 +74,7 @@ if __name__ == "__main__":
 
             if stream_dict.get("plotting"):
                 _logger.info(f"RUN {run_id}: Plotting stream {stream}...")
-                plots = plot_data(cfg, run_id, stream, stream_dict)
+                plots, plotting_opts = plot_data(cfg, run_id, stream, stream_dict, plotting_opts)
 
             if stream_dict.get("evaluation"):
                 _logger.info(f"Retrieve or compute scores for {run_id} - {stream}...")
--- a/packages/evaluate/src/weathergen/evaluate/plot_inference.py
+++ b/packages/evaluate/src/weathergen/evaluate/plot_inference.py
@@ -59,7 +59,6 @@ if __name__ == "__main__":
     scores_dict = defaultdict(lambda: defaultdict(dict))
 
     for run_id, run in runs.items():
-
         plotter = Plotter(cfg, run_id)
         _logger.info(f"RUN {run_id}: Getting data...")
 
@@ -69,7 +68,7 @@ if __name__ == "__main__":
             _logger.info(f"RUN {run_id}: Processing stream {stream}...")
 
             stream_dict = run["streams"][stream]
-           
+
             if stream_dict.get("plotting"):
                 _logger.info(f"RUN {run_id}: Plotting stream {stream}...")
                 plots = plot_data(cfg, run_id, stream, stream_dict)
--- a/packages/evaluate/src/weathergen/evaluate/plot_inference.py
+++ b/packages/evaluate/src/weathergen/evaluate/plot_inference.py
@@ -14,7 +14,6 @@ from collections import defaultdict
 from pathlib import Path
 
 from omegaconf import OmegaConf
-from plotter import Plotter
 from utils import (
     calc_scores_per_stream,
     metric_list_to_json,
@@ -58,11 +57,10 @@ if __name__ == "__main__":
     # to get a structure like: scores_dict[metric][stream][run_id] = plot
     scores_dict = defaultdict(lambda: defaultdict(dict))
 
-    # common plotting options across streams and run_ids. e.g. vertical ranges. 
+    # common plotting options across streams and run_ids. e.g. vertical ranges.
     plotting_opts = {}
 
     for run_id, run in runs.items():
-
         _logger.info(f"RUN {run_id}: Getting data...")
 
         streams = run["streams"].keys()
@@ -74,7 +72,9 @@ if __name__ == "__main__":
 
             if stream_dict.get("plotting"):
                 _logger.info(f"RUN {run_id}: Plotting stream {stream}...")
-                plots, plotting_opts = plot_data(cfg, run_id, stream, stream_dict, plotting_opts)
+                plots, plotting_opts = plot_data(
+                    cfg, run_id, stream, stream_dict, plotting_opts
+                )
 
             if stream_dict.get("evaluation"):
                 _logger.info(f"Retrieve or compute scores for {run_id} - {stream}...")
--- a/packages/evaluate/src/weathergen/evaluate/plotter.py
+++ b/packages/evaluate/src/weathergen/evaluate/plotter.py
@@ -20,11 +20,12 @@ class Plotter:
     Contains all basic plotting functions.
     """
 
-    def __init__(self, cfg: dict, model_id: str = ""):
+    def __init__(self, cfg: dict, model_id: str = "", ranges: dict = {}):
         """
         Initialize the Plotter class.
         :param cfg: config from the yaml file
         :param model_id: if a model_id is given, the output will be saved in a folder called as the model_id
+        :param ranges: dictionary containing global vmax vmin. If Empty it will be filled by the first plot
         """
 
         self.cfg = cfg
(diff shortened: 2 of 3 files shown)


packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -20,11 +20,12 @@ class Plotter:
     Contains all basic plotting functions.
     """
 
-    def __init__(self, cfg: dict, model_id: str = ""):
+    def __init__(self, cfg: dict, model_id: str = "", ranges: dict = None):
         """
         Initialize the Plotter class.
         :param cfg: config from the yaml file
         :param model_id: if a model_id is given, the output will be saved in a folder called as the model_id
+        :param ranges: dictionary containing global vmax vmin. If Empty it will be filled by the first plot
```
mlangguth89: Two notes here:
1. We should follow the standard format of doc-strings that we have at other places. Example:
```
"""
<Description>

Parameters
------------

cfg: 
    config from YAML file
model_id: 
      <description>
"""
```
2. Specify how the dictionary looks like. I would suggest that it is a dictionary with the channel/variable name as key and a tuple for the value range (tuple to ensure immutabiliy).

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -104,6 +107,29 @@ def select_from_da(self, da: xr.DataArray, selection: dict) -> xr.DataArray:
                 da = da.sel({key: value})
         return da
 
+    def get_range(self, var: str) -> tuple[float, float]:
+        """
+        Get stored values of vmin and vmax per variable.
+        :param var: variable for which one needs to retrieve the range
+        :return: tuple with minimum and maximum values.
+        """
+        vmin = self.ranges.get(var, {}).get("vmin", None)
```
mlangguth89: No need for complicated, nested dictionaries IMO. 
Having the range as variable-specific tuples is fine.

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -104,6 +107,29 @@ def select_from_da(self, da: xr.DataArray, selection: dict) -> xr.DataArray:
...
+
+    def set_range(self, var: str, vmin: float, vmax: float) -> None:
+        """
+        Set new values of vmin and vmax for a given variable.
+        :param var: variable for which one needs to set the range
+        :param vmin
+        para: vmax
+        """
+        self.ranges[var] = {
+            "vmin": vmin,
```
mlangguth89: See above. 
Note that tuples are immutable which is a mean to introduce inconsistencies by (accidentally) changing the values on-the-fly.

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -104,6 +107,29 @@ def select_from_da(self, da: xr.DataArray, selection: dict) -> xr.DataArray:
...
+            "vmax": vmax,
+        }
+        return self
```
mlangguth89: There is no need to return self in general in class-methods.

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -187,23 +213,33 @@ def map(
             ax.coastlines()
             da = self.select_from_da(data, select_var).compute()
 
+            vmin, vmax = self.get_range(var)
```
mlangguth89, line 284: If the variable-specific ranges are tuple, the `get_range` becomes a one-liner as it will be unpacked correctly without the need to assess key-value pairs.

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -239,6 +275,8 @@ def __init__(self, cfg: dict):
             _logger.info(f"creating dir {self.out_plot_dir}")
             os.makedirs(self.out_plot_dir, exist_ok=True)
 
+        _logger.info(f"saving summary plots to: {self.out_plot_dir}")
```
mlangguth89: Capitalize 'saving'

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -232,6 +240,7 @@ def plot_data(cfg: str, run_id: str, stream: str, stream_dict: dict) -> list[str
     :param run_id: Run identifier.
     :param stream: Stream name.
     :param stream_dict: Dictionary containing stream configuration.
+    :plotting_opts: Dictionary containing all common plotting options across streams and variables.
```
mlangguth89: We should enhance usability by indicating how the dictionary should look like, i.e.
{ var : (vmin, vmax)} as suggested above.

packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -58,9 +57,10 @@
     # to get a structure like: scores_dict[metric][stream][run_id] = plot
     scores_dict = defaultdict(lambda: defaultdict(dict))
 
-    for run_id, run in runs.items():
+    # common plotting options across streams and run_ids. e.g. vertical ranges.
+    plotting_opts = cfg.get("plotting_opts", {})
```
mlangguth89: Maybe, we should indicte in the variable name that these are **global** plotting options.

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -187,23 +213,33 @@ def map(
...
             plt.title(
-                f"{self.stream}, {var} : fstep = {self.fstep:03} ({da['valid_time'][0].values})"
+                f"{self.stream}, {var} : fstep = {self.fstep:03} ({da['valid_time'][0].values.astype('datetime64[s]')})"
             )
             ax.set_global()
             ax.gridlines(draw_labels=False, linestyle="--", color="black", linewidth=1)
 
+            # store range for consistency across streams/fsteps
+            if not vmin or not vmax:
+                vmin = scatter_plt.get_clim()[0]
```
mlangguth89, line 310: Wouldn't it be better to derive these parameters from all the samples that should be plotted rather than from the first one? Basing this on a single sample may be sub-optimal for data with a pronounced sample-to-sample variation, e.g. due to diurnal and seasonal cycles (e.g. 2m temperature for a winter sample from the CERRA data).
Not sure if this is too complex for now and if we should leave it open for future work.

mlangguth89 (commented):
Thanks @iluise for your work on this and the nice realization. I've added a couple of comments, hoping to further improve the implementation.

packages/evaluate/src/weathergen/evaluate/plotter.py
clessig: Yes, please: we need consistent vmin and vmax for a forecast sequence. Bonus is to do this consistent across multiple streams (this is #623) but for one stream would be a good first step.

iluise: @Michael: I thought about it, but then we would need to cache all the plots, adjust the colorbars and save them from cache, but for seviri this would probably very quickly lead to OOM. Or you have a better solution? 
for me adjusting on the first sample + the option of a custom range from the config is already the best compromise. 

My idea is that: if the forecast is stable getting the range from the first one is not an issue. If the forecast is exploding then you can adjust the range as you like, but having one range that covers all steps is in any case suboptimal for some fsteps, either the first ones or the last ones. That's why in these cases there's the option to adjust the range as you like from the config. I find this "human-supervised" way more useful for debugging than trying to over-optimize the ranges automatically, what do you think? 

@Christian: what do yo mean? this already gives a consistent range across fsteps, streams and run_ids

mlangguth89: @iluise 
One could calculate the min and max-values from all samples of a stream-specific channel by calling `.min()` and `.max()` on the data arrays (of the first stream). This adds computational complexity, but should not lead to OOM-issues thanks to dask.
Assuming that the value range does not significantly change accross different datasets (e.g. t2m from ERA5 should be similar to t2m from CERRA), this would be a robust approach.

clessig: The color range will change significantly between ERA5 and CERRA due to the poles, tropics etc in ERA5, e.g. for t2m.

iluise: Yeah, that's what I thought as well. We would need to load all steps and requested samples of all streams in advance just to get the ranges, that goes against the idea of fast evaluation to me 😅  

Can we give it a try by using the first forecast step for now and if we see that it is really suboptimal and users need to tune parameters too often to min and max we will change it accordingly? 

Having ranges that are too wide when using min and max imo is also suboptimal as it would hide features. There's no ideal solution if the forecasts are diverging too much, users will just customize them according to what they want to see.

clessig: We need consistent and valid plot range across forecast steps. Across streams is more difficult and it would be ok if there's a more complicated way to do it.

iluise: One solution here would be to compute max and min per variable in the ZarrIO class and store them as attributes in the zarr, like what is done in anemoi-datasets right now. This will make the retrieval so much easier. 

Proposal: Can we merge it like this using the first forecast step and open a follow up issue to get min and max from the ZarrIO class + get the ranges from it?

The only cases where this is might be suboptimal are when one wants to plot a specific subregion. But these cases would be covered by the custom vmin and vmax option in the config.

packages/evaluate/src/weathergen/evaluate/plot_inference.py
iluise: I renamed it "common_ranges"

mlangguth89: We should not narrow it ranges. One likely also wants to customize the colormap or provide levels instead of `vmin` and `vmax` in the future.

Issue #604, opened 2025-07-28
Proper defaults for all paths in evaluation-script plot_inference


mlangguth89:
### Is your feature request related to a problem? Please describe.

Currently, the path to save the results from the model evaluation (from `plot_inference.py`) must be defined in the configuration script. Furthermore, some filenames of the created plots are arbitrary. 
Thus, a proper naming scheme applied to all generated files (plots and JSON-diagnostics) should be set up with sensible defaults.

### Describe the solution you'd like

The following changes will be implemented:
- Relative paths in the config should be replaced by defaults pointing to the standard `results`-directory of the WeatherGenerator
- All model-specific plots should (by default) be saved under `${WEATHERGENERATOR_HOME}/results/<run_id>/plots`
- All model-specific evaluation JSON-files should (by default) be saved `${WEATHERGENERATOR_HOME}/results/<run_id>/evaluation/`
- Plots for model comparison should include a sorted list of run IDs in their filename (instead of ambigious namings such as `plots/png/line_plots/compare_rmse_ERA5_z_500.png`. They are stored in `comparison_plots_dir` which defaults to `./plots/`.

### Describe alternatives you've considered

N/A

### Additional context

Please add further suggestions in the comments. This will be accepted unless they increase the scope of this issue-branch too significantly.

### Organisation

JSC

PR #652, opened 2025-07-31
Implement regional evaluation 

### LINKED ISSUE #540: Scores need to be computed per hemisphere

Currently, scores are computed locally. WMO/ECMWF standard is scores per NHem, Tropics, SHem.

Add score computation per NHem, Tropics, SHem. This is high priority since we need it for a fair comparison to standard eval scores.

CC @mlangguth89 , @iluise 

_No response_

### Additional context

_No response_

### Organisation

_No response_

mlangguth89:
## Description

This PR adds support for regional evaluation of the WeatherGenerator output.
Currently, three regions are supported next to global evaluation: ["northern_hemisphere", "southern_hemisphere", "tropics"]. 

Note that the extension to regional evaluation required renaming of the metric JSON-files:
The region identifier is now appended after the stream name in the filename.
```
Example: `h63rbq4y_ERA5_tropics_l1_epoch00000.json` <-- ERA5 output evaluated over the tropical region
```
The results of the global evaluation (previously the default) will now be stored under 
```
`h63rbq4y_ERA5_global_l1_epoch00000.json`.
```
As a result of this change, previously generated evaluation files (which lack the region identifier) will not be recognized by the updated code. One has two options:
   - Manually rename the old metric files to include` _global_` in the correct position.
   - Re-run the evaluation, which will create new files using the new naming convention.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #540 

## Code Compatibility

### Dependencies

### Documentation

The documentation in the Wiki must be updated after merging.

Example config YAML-file (focussed on evaluation):
```
verbose: true
results_dir : "/p/home/jusers/langguth1/juwels/WeatherGenerator_ownfork/results/"
output_plotting_dir: "/p/home/jusers/langguth1/juwels/WeatherGenerator_ownfork/plots/"
output_scores_dir : "/p/home/jusers/langguth1/juwels/WeatherGenerator_ownfork/results/"
dpi_val : 300
create_html : false
summary_plots : true
print_summary: false

evaluation:
  regions: ["global", "tropics"]


#MTM models
run_ids :
  h63rbq4y:
    streams:
      ERA5:
        channels: ["2t", "10u"]
        evaluation:
          forecast_step: "all"
          sample: "all"
        plotting:
          plot_maps: false
          plot_histograms: false
    label: "MTM ERA5"
    epoch: 0
    rank: 0
```

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/utils.py
+++ b/packages/evaluate/src/weathergen/evaluate/utils.py
@@ -20,7 +20,7 @@ from tqdm import tqdm
 from weathergen.common.io import ZarrIO
 from weathergen.evaluate.plotter import DefaultMarkerSize, LinePlots, Plotter
 from weathergen.evaluate.score import VerifiedData, get_score
-from weathergen.evaluate.score_utils import to_list, RegionBoundingBox
+from weathergen.evaluate.score_utils import RegionBoundingBox, to_list
 
 _logger = logging.getLogger(__name__)
 _logger.setLevel(logging.INFO)


packages/evaluate/src/weathergen/evaluate/score_utils.py
```diff
@@ -30,3 +36,100 @@ def to_list(obj: Any) -> list:
...
+
+
+class RegionLibrary:
+    """
+    Predefined bounding boxes for known regions.
+    """
+
+    REGIONS: ClassVar[dict[str, tuple[float, float, float, float]]] = {
+        "global": (-90.0, 90.0, -180.0, 180.0),
+        "northern_hemisphere": (0.0, 90.0, -180.0, 180.0),
```
iluise: I'd use "nhem", "shem", "tropics" as the ECMWF's convention

packages/evaluate/src/weathergen/evaluate/score_utils.py
```diff
@@ -30,3 +36,100 @@ def to_list(obj: Any) -> list:
...
+        try:
+            return cls(*RegionLibrary.REGIONS[region])
+        except KeyError as err:
+            raise ValueError(
+                f"Region '{region}' is not supported. "
+                f"Available regions: {', '.join(RegionLibrary.REGIONS.keys())}"
+            ) from err
+
+
+class RegionLibrary:
```
iluise: minor: can we put this before RegionBoundingBox?

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -229,22 +246,45 @@ def calc_scores_per_stream(
...
+        else:
+            # depending on the datset, there might be no data (e.g. no CERRA in southern hemisphere region)
+            _logger.warning(
+                f"No data available for stream {stream} at forecast step {fstep} in region {region}. Skipping metrics calculation."
+            )
 
-        _logger.debug(f"Running computation of metrics for stream {stream}...")
-        combined_metrics = combined_metrics.compute()
-        combined_metrics = scalar_coord_to_dim(combined_metrics, "channel")
+            combined_metrics = xr.DataArray(
```
iluise: uhm, this will write a json full of nans right? Maybe we can just skip it with an "else: continue"? 
The only place where I can think of where skipping instead of having nans is in the summary_plots, no?

If it complains one can refine the 'if exists' checks like for the `run_id`: https://github.com/ecmwf/WeatherGenerator/blob/develop/packages/evaluate/src/weathergen/evaluate/utils.py#L507

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -371,31 +411,31 @@ def metric_list_to_json(
...
-    streams : list[str]
+    streams :
         Stream names.
-    metric_dir : Path
+    region :
+        Region name.
+    metric_dir :
         Output directory.
-    run_id : str
+    run_id :
```
iluise, line 423: can you fill these again?

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -485,60 +530,77 @@ def plot_summary(cfg: dict, scores_dict: dict, print_summary: bool):
...
-                    scores_dict[metric][stream][run_id]["channel"].values
+            # get total list of channels
+            # TODO: improve this
+            channels_set = list(
+                set(
+                    value
+                    for run_id in runs
+                    for stream in scores_dict.get(metric).get(region).keys()
+                    if run_id
+                    in scores_dict.get(metric, {})
```
iluise, line 546: with the `continue` logic this needs to be as for the run_id

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -195,7 +212,7 @@ def calc_scores_per_stream(
     if fsteps == "all":
         fsteps = None
 
-    output_data = get_data(cfg, run_id, stream, return_counts=True)
+    output_data = get_data(cfg, run_id, stream, region=region, return_counts=True)
```
iluise, line 214: With this logic we need to call get_data for each region. Which means we need to repeat:
```
Processing wz9rbl3w - ERA5 - 0: 100%|█| 244/244 [00:15<00:00, 16.06it/s
```
each time. This takes 15 seconds each time we call it for the 2 month evaluation and I guess minutes for high-res yearly preds. How difficult would it be to call get_data once and loop over regions right after?
I'm also fine to merge it as it is and improve it in a second step when/if this becomes a bottleneck

iluise (commented):
Super, thanks for implementing it. I like the way it is implemented. I just put a couple of suggestions to make it faster. 

One thing: the folder becomes difficult to navigate when plotting all supported regions. Can we add per region sub-folders when saving the summary plots?

packages/evaluate/src/weathergen/evaluate/utils.py
mlangguth89: Yes, this is not super smart by design so far. Give me a try to improve it in the revision - in principle this should be feasible, but I remember that I encountered some difficulties.

packages/evaluate/src/weathergen/evaluate/utils.py
mlangguth89: Yes, the reason why I implemented this was the data retrieval in `summary_plots`. 
I strumbled with the complex nested looping + if-statement [here](https://github.com/ecmwf/WeatherGenerator/blob/develop/packages/evaluate/src/weathergen/evaluate/utils.py#L507), and continued therefore in a pragmatic way.
We should definately find a way to make the latter easier to read/handle.

packages/evaluate/src/weathergen/evaluate/score_utils.py
mlangguth89: Sure!

packages/evaluate/src/weathergen/evaluate/score_utils.py
mlangguth89: Would also be great if someone (e.g. @clessig=) could double-check the correctness of the tropics-region definition.

packages/evaluate/src/weathergen/evaluate/utils.py
mlangguth89: As @tjhunter pointed out, this is redundant with the type-indication in the function signature.

mlangguth89:
> Super, thanks for implementing it. I like the way it is implemented. I just put a couple of suggestions to make it faster.
> 
> One thing: the folder becomes difficult to navigate when plotting all supported regions. Can we add per region sub-folders when saving the summary plots?

Maybe we do this in #604 where we specifically revise the output directory-structure?

packages/evaluate/src/weathergen/evaluate/utils.py
mlangguth89: Can we have a follow-up issue on this?
I would need to refactor substantial parts of the code, since the region-handling has implications in `plot_inference.py` (see e.g. [here](https://github.com/ecmwf/WeatherGenerator/blob/ccd7894b16de00838bf528a714631237a61a453e/packages/evaluate/src/weathergen/evaluate/plot_inference.py#L80)). This would delay the implementation.

iluise: yes, sounds good. Can you open the issue or should I? 
Oks, for adding the folders in the other PR.

clessig (approved):

PR #677, opened 2025-08-05
Refactor path-setting for 'model' and 'results' to be dynamic (no relative paths) (Closes #591)

moritzhauschulz:
## Description

Refactor the path-setting in train_continue and inference functions to not rely on relative paths, but instead use 'path_shared_working_dir' and build 'model' and 'result' paths dynamically. This is to ensure that models can be transferred between HPC systems more smoothly, and follows the general practice of not using relative paths. (Closes #591) 

## Type of Change

-   [x] Bug fix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update

## Issue Number

Closes #591 

## Code Compatibility

-   [x] I have performed a self-review of my code

### Code Performance and Testing

-   [x] I ran the `uv run train` and (if necessary) `uv run evaluate` on a least one GPU node and it works 
-   [ ] If the new feature introduces modifications at the config level, I have made sure to have notified the other software developers through Mattermost and updated the paths in the `$WEATHER_GENERATOR_PRIVATE` directory

The change is designed to allow models trained on one HPC to be loaded on another. We tried migrating models between HPC2020 and jewels-booster in both directions, which worked smoothly. Additional testing may be useful, but I do not expect this to cause any breaking changes.

### Dependencies

-   [ ] I have ensured that the code is still pip-installable after the changes and runs
-   [ ] I have tested that new dependencies themselves are pip-installable.
-   [ ] I have not introduced new dependencies in the inference portion of the pipeline

### Documentation

-   [x] My code follows the style guidelines of this project
-   [x] I have updated the documentation and docstrings to reflect the changes
-   [x] I have added comments to my code, particularly in hard-to-understand areas

## Additional Notes

We were not able to reproduce the error described in #591, but performed the changes regardless, to reduce the number of references to relative paths in the codebase. This is in line with agreed best practices.

Issue #591, opened 2025-07-25
Paths from private config should always be read at launch time


clessig:
### Is your feature request related to a problem? Please describe.

Currently, starting a new training run reads the paths from the private config, but train_continue and inference re-use them. This leads to problems when the paths change or when a model changes machines. 

### Describe the solution you'd like

train, train_continue and inference should always read paths from the private config at start time.

PR #681, opened 2025-08-05
Revised path defaults and output dirctory structure for fast evaluation

### LINKED ISSUE #604: Proper defaults for all paths in evaluation-script plot_inference

mlangguth89:
## Description

This PR removes the necessity to set any path-parameters in the YAML config file for the fast evaluation (aka `plot_inference.py`). By default, the inference output data is assumed to be available under `WeatherGenerator/results/<run_id>`. Overwriting is however possible to support custom directory paths by setting `results_base_dir` in the configuring YAML-file.
Additionally, the organization of the fast evaluation output is revised following the description in #604. 

## Type of Change

-   [ ] Bug fix (non-dwabreaking change which fixes an issue)
-   [ ] Documentation update

## Issue Number

Closes #604

## Code Compatibility

### Dependencies

### Documentation

Setting `results_base_dir` is now optional. Likewise, the base directories for the evaulation output can be modified via the following optional parameters:
- `runplot_base_dir`: base directory where the map plots and histograms are saved
- `metric_base_dir`: base directory where the JSON  metric files are saved or expected to be located if they should be read in)
- `summary_dir`: base directory where the summary plots are saved
The sub-directory structure remains unaffected when setting these parameters, and follows the structure as described in #604.

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/plot_inference.py
+++ b/packages/evaluate/src/weathergen/evaluate/plot_inference.py
@@ -15,7 +15,7 @@ from pathlib import Path
 
 from omegaconf import DictConfig, OmegaConf
 
-from weathergen.utils.config import load_config, set_paths, _REPO_ROOT
+from weathergen.src.weathergen.utils.config import load_config, set_paths, _REPO_ROOT
 from weathergen.evaluate.utils import (
     calc_scores_per_stream,
     metric_list_to_json,
@@ -41,6 +41,7 @@ def run_main(cfg: DictConfig) -> None:
     )  # base directory where summary plots will be stored
 
     metrics = cfg.evaluation.metrics
+    regions = cfg.evaluation.get("regions", ["global"])
 
     # to get a structure like: scores_dict[metric][region][stream][run_id] = plot
     scores_dict = defaultdict(lambda: defaultdict(lambda: defaultdict(dict)))
@@ -55,19 +56,21 @@ def run_main(cfg: DictConfig) -> None:
         if results_base_dir is None:
             cf_run = load_config(private_paths, run_id, run["epoch"])
             cf_run = set_paths(cf_run)
-            results_base_dir = cf_run["run_path"]
+            results_dir = cf_run["run_path"]
+            logging.info(f"Results directory obtained automatically: {results_dir}")
         else:
-            results_base_dir = Path(results_base_dir)
+            results_dir = Path(results_base_dir) / run_id
+            logging.info(f"Results directory parsed: {results_dir}")
 
         runplot_base_dir = Path(
-            run.get("runplot_base_dir", results_base_dir)
+            run.get("runplot_base_dir", results_base_dir.parent)
         )  # base directory where map plots and histograms will be stored
         metric_base_dir = Path(
             run.get("metric_base_dir", results_base_dir)
         )  # base directory where score files will be stored
 
         streams = run["streams"].keys()
-        metric_dir = metric_base_dir / run_id / "evaluation"
+        metric_dir = metric_base_dir / "evaluation"
 
         for stream in streams:
             _logger.info(f"RUN {run_id}: Processing stream {stream}...")
@@ -77,7 +80,7 @@ def run_main(cfg: DictConfig) -> None:
             if stream_dict.get("plotting"):
                 _logger.info(f"RUN {run_id}: Plotting stream {stream}...")
                 _ = plot_data(
-                    cfg, results_base_dir, runplot_base_dir, run_id, stream, stream_dict
+                    cfg, results_dir, runplot_base_dir, stream, stream_dict
                 )
 
             if stream_dict.get("evaluation"):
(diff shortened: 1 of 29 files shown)


packages/evaluate/src/weathergen/evaluate/plot_inference.py
```diff
@@ -25,27 +26,39 @@
 
 _logger = logging.getLogger(__name__)
 
+_DEFAULT_RESULT_PATH = _REPO_ROOT / "results"
```
clessig: Shouldn't we get this from the private repo? Then we wouldn't have to rely on that the softlinks are set correctly.

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -14,41 +14,47 @@
 _logger = logging.getLogger(__name__)
 _logger.setLevel(logging.INFO)
 
+_REPO_ROOT = Path(
+    __file__
+).parent.parent.parent.parent.parent.parent  # TODO use importlib for resources
+_DEFAULT_RESULT_PATH = _REPO_ROOT / "results"
```
clessig: Same as above

packages/evaluate/src/weathergen/evaluate/plot_inference.py
mlangguth89: Please provide some more information here (cf. my comment rely below).

packages/evaluate/src/weathergen/evaluate/plotter.py
mlangguth89: Side note: One should import `_REPO_ROOT` from the common-package.

moritzhauschulz:
> Reading the default paths for models and results from the private repo into the runtime config is currently being implemented by @moritzhauschulz in #677, right. This should be applicable here too and I'd suggest to read the results path from the config, such as done in @kctezcan's #631.

In my code I haven't used `_REPO_ROOT' yet (I was not aware of this). I think however it might be almost equivalent to using the `path_shared_working_dir' (from paths.yaml) which is what I am doing. Please let me know which way is preferred, or if there is a major difference that I am not seeing. At first glance, using `_REPO_ROOT' seems reasonable.

mlangguth89:
> > > Thanks for the idea. One fundamental question: shouldn't the default results dir come from the private repo?
> > 
> > 
> > @clessig I'm not fully getting the point here. Can you provide more detail how 'getting the default from the private-repo' can be realized in code and how it differs from the implemented approach?
> > Actually, setting up the repo as written in the Wiki by running
> > ```
> > ./WeatherGenerator/scripts/actions.sh create-links
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > creates symlinks for the `results`, `plots`, ... under `./WeatherGenerator`. These linked directories are used as defaults here with the help of `_REPO_ROOT`. The introduction of `_REPO_ROOT` has been introduced in consolutation with @tjhunter into the common-package.
> > If users deviate from the common directory structure or have their results stored at a different location (I have, for instance, a `resutlts_private`-directory since data has been deleted twice in the shared work space), they need to set `results_base_dir` in the config-file for the evaluation.
> 
> Yes, people should create the symlinks. But it's ultimately an optional step that doesn't need to be there to run the model. The more robust is to use the same `results` default as the model. And this comes from the private repo.
> 
> @moritzhauschulz @MatKbauer : can you help here? I think you solved the same issue for #591

To be clear about it: The info comes from the HPC-specific `path.yaml`, from what the user sets via the env variable `WEATHERGEN_PRIVATE_CONF` or from a user specified path to a proper YAML-file (`private_home`), as can be seen [here](https://github.com/ecmwf/WeatherGenerator/blob/8a0782fbd1b708d6acd19c1ad3ea3a1eb63c0525/src/weathergen/utils/config.py#L238). The default fallback does exactly the same thing as the code in this PR:  Set `_REPO_RESULT_RUN` which is equivalent to `_REPO_ROOT / "results"` (see here: https://github.com/ecmwf/WeatherGenerator/blob/8a0782fbd1b708d6acd19c1ad3ea3a1eb63c0525/src/weathergen/utils/config.py#L345)

clessig:
> > Reading the default paths for models and results from the private repo into the runtime config is currently being implemented by @moritzhauschulz in #677, right. This should be applicable here too and I'd suggest to read the results path from the config, such as done in @kctezcan's #631.
> 
> In my code I haven't used `_REPO_ROOT` yet (I was not aware of this). I think however it might be almost equivalent to using the `path_shared_working_dir` (from paths.yaml) which is what I am doing. Please let me know which way is preferred, or if there is a major difference that I am not seeing. At first glance, using `_REPO_ROOT` seems reasonable.

Yes, let's please use the info from `paths.yaml` via `path_shared_working_dir`

mlangguth89:
> > > Reading the default paths for models and results from the private repo into the runtime config is currently being implemented by @moritzhauschulz in #677, right. This should be applicable here too and I'd suggest to read the results path from the config, such as done in @kctezcan's #631.
> > 
> > 
> > In my code I haven't used `_REPO_ROOT` yet (I was not aware of this). I think however it might be almost equivalent to using the `path_shared_working_dir` (from paths.yaml) which is what I am doing. Please let me know which way is preferred, or if there is a major difference that I am not seeing. At first glance, using `_REPO_ROOT` seems reasonable.
> 
> Yes, let's please use the info from `paths.yaml` via `path_shared_working_dir`

I've updated the code to obtain the default-configs via `load_model_config` or from `paths.yaml` if provided. This is now in-line with code in main repository. 
Additionally, the user can now even provide a custom result path per evaluated `run_id` in case that inference results are stored in a non-standard directory.
An example config looks as follows:
```
verbose: true
summary_dir: /p/home/jusers/langguth1/juwels/WeatherGenerator_ownfork/plots
dpi_val : 300
create_html : true
summary_plots : true
print_summary: false

evaluation:
  regions: ["global", "tropics"]


#MTM models
run_ids :
  h63rbq4y:
    results_dir: /p/home/jusers/langguth1/juwels/WeatherGenerator_ownfork/results_private
    label: "MTM ERA5 and IMERG"
    epoch: 0
    rank: 0
    streams:
      ERA5:
        channels: ["2t", "10u"]
        evaluation:
          forecast_step: "all"
          sample: "all"
        plotting:
          sample: [1, 2]
          plot_maps: true
          plot_histograms: true
      IMERG:
        forecast_step: [0]
        channels: ["tp_imerg_0"]
        evaluation: 
          forecast_step: "all"
          sample: "all"
        plotting:
          sample: [1, 2, 3]
          plot_maps:
            marker_size: 0.3
          plot_histograms: false
```

clessig:
Thanks! But I don't think the per-run plots are stored in `path_shared_working_dir/results`. This is what we had agreed on in the second but last evaluation meeting.

MatKbauer:
I like the intention to structure the output as described [here](https://github.com/ecmwf/WeatherGenerator/issues/604#issuecomment-3133187591). After running `uv run evaluate`, I
-  found the plots in the `./WeatherGenerator/results/{RUN_ID}/plots/` directory, separated by stream, pred/target, and map/hist,
- obtained `./WeatherGenerator/results/{RUN_ID}/evaluation/` directory with separate `.json` result files per region (e.g., global and tropics), and
- saw the inter-model comparison plots were written to `./WeatherGenerator/plots/line_plots/`.

This looks great! Now, adding custom directories to the `eval_config.yml` file, such as
```
summary_dir: "./plots2/"
metric_base_dir: "./plots2/metrics/"
runplot_base_dir: "./plots2/runplots/"
```
rooted the writing of the comparison plots to `./WeatherGenerator/plots2/line_plots/`, whereas the metrics and runplots ended up in the same directory as in the default above. Is this intended and what are `metric_base_dir` and `runplot_base_dir` supposed to control?

MatKbauer (approved):
New evaluation directory structure works. We can merge this PR and discuss the custom `metric_base_dir` in issue #737.

PR #694, opened 2025-08-08
Set default value for config file

### LINKED ISSUE #693: Default path for evaluation config file

When running `uv run evaluate`, the is no default config file provided. It must, instead, be explicitly passed via `--config path/to/eval/config/file`.

Modify the `run_evaluate()` function to deliver a default name for the evaluation configuration file if no `--config` argument is not passed.

_No response_

### Additional context

_No response_

### Organisation

ECMWF

MatKbauer:
## Description

Added a default value to the `run_evaluate.py` to deliver `eval_config.yml` if no `--config` argument is passed.

## Type of Change

-   [-] New feature (non-breaking change which adds functionality)
-   [ ] Documentation update

## Issue Number

resolves #693 

## Code Compatibility

-   [x] I ran `uv run evaluate` and it works 

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/run_evaluation.py
+++ b/packages/evaluate/src/weathergen/evaluate/run_evaluation.py
@@ -33,13 +33,20 @@ def evaluate() -> None:
     parser.add_argument(
         "--config",
         type=str,
-        help="Path to the configuration yaml file for plotting. e.g. config/plottig_config.yaml",
+        help="Path to the configuration yaml file for plotting. e.g. config/eval_config.yml",
+        default="config/eval_config.yml"
     )
 
     args = parser.parse_args()
 
     # configure logging
     logging.basicConfig(level=logging.INFO)
+
+    assert Path(args.config).exists(), (
+        "The evaluation configuration file does not exist. Its "
+        "default location would be 'config/eval_config.yml'. Either create it and restart "
+        "evaluation, or pass the config file with the --config argument."
+    )
 
     # load configuration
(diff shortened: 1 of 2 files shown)


packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -33,14 +33,21 @@ def evaluate() -> None:
     parser.add_argument(
         "--config",
         type=str,
-        help="Path to the configuration yaml file for plotting. e.g. config/plottig_config.yaml",
+        help="Path to the configuration yaml file for plotting. e.g. config/eval_config.yml",
+        default="config/eval_config.yml"
```
mlangguth89, line 37: This is a  hard-coded, relative paths, i.e. its validity depends on the directory from where `run_evaluation.py` is called.
Please make use of `_REPO_ROOT` from `config.py` here.

uv.lock
```diff
@@ -2380,6 +2381,7 @@ requires-dist = [
     { name = "torch", marker = "sys_platform == 'macosx'", specifier = "==2.6.0", index = "https://download.pytorch.org/whl/cpu" },
     { name = "tqdm" },
     { name = "weathergen-common", editable = "packages/common" },
+    { name = "weathergen-evaluate", editable = "packages/evaluate" },
```
mlangguth89, line 2384: :+1:

mlangguth89 (changes requested):
Thanks a lot for picking this up.
Only one thing to address -> We need to avoid hard-coded relative paths.

PR #705, opened 2025-08-11
Implement first function for latitude weighting

### LINKED ISSUE #466: Latitude-weighted loss function

_No response_

_No response_

_No response_

### Additional context

_No response_

### Organisation

_No response_

Jubeku:
## Description

Adding optional latitude weighting functions to loss calculations. 

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #466 

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/config/streams/streams_anemoi/era5.yml
+++ b/config/streams/streams_anemoi/era5.yml
@@ -13,7 +13,6 @@ ERA5 :
   source_exclude : ['w_', 'skt', 'tcw', 'cp', 'tp']
   target_exclude : ['w_', 'slor', 'sdor', 'tcw', 'cp', 'tp']
   loss_weight : 1.
-  location_weight : cosine_latitude
   masking_rate : 0.6
   masking_rate_none : 0.05
   token_size : 8


src/weathergen/train/loss_calculator.py
```diff
@@ -140,7 +154,8 @@ def _loss_per_loss_function(
 
         ctr_substeps = 0
         for mask_t in substep_masks:
-            assert mask_t.sum() == len(weights_locations) if weights_locations else True
+            if weights_locations is not None:
```
clessig: This is equivalent to the old version, not?

src/weathergen/train/loss_calculator.py
```diff
@@ -105,6 +105,26 @@ def _get_weights(self, stream_info):
 
         return stream_info_loss_weight, weights_channels
 
+    def cosine_latitude(self, stream_data, forecast_offset, fstep, min_value=1e-3, max_value=1.0):
```
clessig: This function should be outside of the class

src/weathergen/train/loss_calculator.py
```diff
@@ -105,6 +105,26 @@ def _get_weights(self, stream_info):
...
+        latitudes_radian = (
+            stream_data.target_coords_raw[forecast_offset + fstep][:, 0] * np.pi / 180
+        )
+        return (max_value - min_value) * np.cos(latitudes_radian) + min_value
+
+    def _get_location_weights(self, stream_info, stream_data, forecast_offset, fstep):
+        stream_info_location_weight = stream_info.get("location_weight", None)
```
clessig: stream_info_location_weight -> location_weight_name or location_weight_type

src/weathergen/train/loss_calculator.py
```diff
@@ -105,6 +105,26 @@ def _get_weights(self, stream_info):
...
+        weights_locations_fct = (
+            getattr(self, stream_info_location_weight) if stream_info_location_weight else None
+        )
+        weights_locations = (
+            weights_locations_fct(stream_data, forecast_offset, fstep)
+            if weights_locations_fct
+            else None
+        )
+        weights_locations = weights_locations.to(device=self.device, non_blocking=True)
```
clessig, line 114: This will fail when weights_locations is None, i.e. when weights_locations_fct is None

src/weathergen/train/loss_calculator.py
```diff
@@ -105,6 +105,26 @@ def _get_weights(self, stream_info):
...
 
+    def cosine_latitude(self, stream_data, forecast_offset, fstep, min_value=1e-3, max_value=1.0):
+        latitudes_radian = (
+            stream_data.target_coords_raw[forecast_offset + fstep][:, 0] * np.pi / 180
+        )
+        return (max_value - min_value) * np.cos(latitudes_radian) + min_value
+
+    def _get_location_weights(self, stream_info, stream_data, forecast_offset, fstep):
+        stream_info_location_weight = stream_info.get("location_weight", None)
+        weights_locations_fct = (
```
clessig: I would early out before this function.

```
if weights_locations_fct is None:
  return None
```

src/weathergen/train/loss.py
```diff
@@ -135,9 +135,10 @@ def mse_channel_location_weighted(
     mask_nan = ~torch.isnan(target)
     pred = pred[0] if pred.shape[0] == 0 else pred.mean(0)
 
-    diff2 = torch.square(torch.where(mask_nan, target, 0) - torch.where(mask_nan, pred, 0)).mean(0)
-    wl = weights_points
-    loss_chs = ((diff2.transpose(1, 0) * wl).transpose(1, 0) if wl else diff2).mean(0)
+    diff2 = torch.square(torch.where(mask_nan, target, 0) - torch.where(mask_nan, pred, 0))
```
clessig, line 138: Thanks for the fix here. This was actually incorrect (and only worked since we didn't have weights).

clessig (commented):
Looks overall good, just some minor comments. We should get this merged quite quickly and have separate PRs for adding more complex lat- and variable weighting schemes.

src/weathergen/train/loss_calculator.py
Jubeku: For me the in-line if condition doesn't work if weights_locations is a tensor but checking if it's not None does. Does this make sense?

clessig: But would this work:

```
assert mask_t.sum() == len(weights_locations) if weights_locations is not None else True
```

Jubeku: Not sure if I tried -  let me test again.
Btw, I had the same issue [here](https://github.com/ecmwf/WeatherGenerator/pull/705/files#diff-7b5c433e5f3049bb43897cfc93a3d0cd1ffabeadde5a754a5f7a268bc9566428R139).

src/weathergen/train/loss_calculator.py
Jubeku: Moved it to loss.py

src/weathergen/train/loss_calculator.py
Jubeku: That works. Fixed it.

src/weathergen/train/loss_calculator.py
Jubeku: I check now even before if `location_weight_type` (before named `stream_info_location_weight`) is None and return, this way I don't have to check afterwards. This also fixes `.to(device)` as it will never receive None. 
Checking early simplified the whole function, thanks.

src/weathergen/train/loss_calculator.py
Jubeku: See comment in https://github.com/ecmwf/WeatherGenerator/pull/705#discussion_r2276111271

clessig (approved):
Ready to be merged.

PR #732, opened 2025-08-13
Update handling of `run_path` and `model_path` (Closes #716)

### LINKED ISSUE #716: config.get("path_shared_working_dir") + "results" -> unsupported operand type(s) for +: 'NoneType' and 'str'
### What happened?

in the line https://github.com/ecmwf/WeatherGenerator/blob/acb0be2e2a9ae65f0d9aa68d7beb95c7bb42a489/src/weathergen/utils/config.py#L352
I am getting the error 
```
  File "<string>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```
even though the `run_path` is actually set.
because it is first evaluating the fail condition and 1) `config.get("path_shared_working_dir")` is None because I don't have that in my config and 2) None + "results" is not a valid string concatenation.

this is not backward compatible and forcing me to set `path_shared_working_dir` even though i have my results path already set

run ltest develop without the entry `path_shared_working_dir`

### Version

0.1

linux

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

moritzhauschulz:
## Description

Update handling of run_path and model_path to allow them to be specified even if `path_shared_working_dir` is not (Closes #716).

## Type of Change

-   [ ] Documentation update

## Issue Number

#716 

## Code Compatibility

UNTESTED

### Dependencies

### Documentation

## Additional Notes

Alternatively, must always specify `path_shared_working_dir` in private config.

### DIFF
--- a/src/weathergen/utils/config.py
+++ b/src/weathergen/utils/config.py
@@ -342,18 +342,20 @@ def load_streams(streams_directory: Path) -> list[Config]:
 def set_paths(config: Config) -> Config:
     """Set the configs run_path model_path attributes to default values if not present."""
     config = config.copy()
-    config.run_path = _get_config_attribute(config=config, 
-                                            attribute_name="run_path", fallback="results")
-    config.model_path = _get_config_attribute(config=config,
-                                              attribute_name="model_path", fallback="models")
+    config.run_path = _get_config_attribute(
+        config=config, attribute_name="run_path", fallback="results"
+    )
+    config.model_path = _get_config_attribute(
+        config=config, attribute_name="model_path", fallback="models"
+    )
 
     return config
 
 
 def _get_config_attribute(config: Config, attribute_name: str, fallback: str) -> str:
     """Get an attribute from a Config. If not, fall back to path_shared_working_dir concatenated
-    with the desired fallback path. Raise an error if neither the attribute nor 
-    is specified.""" 
+    with the desired fallback path. Raise an error if neither the attribute nor
+    is specified."""
     attribute = OmegaConf.select(config, attribute_name)
     fallback_root = OmegaConf.select(config, "path_shared_working_dir")
     assert attribute is not None or fallback_root is not None, (


moritzhauschulz:
@MatKbauer

src/weathergen/utils/config.py
```diff
@@ -342,12 +342,22 @@ def load_streams(streams_directory: Path) -> list[Config]:
 def set_paths(config: Config) -> Config:
     """Set the configs run_path model_path attributes to default values if not present."""
     config = config.copy()
-    config.run_path = config.get("run_path", config.get("path_shared_working_dir") + "results")
-    config.model_path = config.get("model_path", config.get("path_shared_working_dir") + "models")
+    config.run_path = get_private_config_attribute(config, "run_path", "results")
+    config.model_path = get_private_config_attribute(config, "model_path", "models")
```
MatKbauer: I'm a fan of explicit argument passing. Could you change to `get_config_attribute(private_config=config, attribute="run_path", fallback="results")`?

src/weathergen/utils/config.py
```diff
@@ -342,12 +342,22 @@ def load_streams(streams_directory: Path) -> list[Config]:
...
 
     return config
 
 
+def get_private_config_attribute(config: Config, attribute: str, fallback: str):
```
MatKbauer: Even though I have suggested to name the function `get_private_config_attribute`, let's rename to `get_config_attribute`, as it is not for private configs only. Let's have a DocString too.

MatKbauer (commented):
Looks good; I like we can reuse the `get_config_attribute` in other places. Some comments and we are ready to

src/weathergen/utils/config.py
```diff
@@ -342,12 +342,22 @@ def load_streams(streams_directory: Path) -> list[Config]:
...
+    # Check if attribute is available in config and fall back to "path_shared_working_dir" if not
+    attribute = OmegaConf.select(config, attribute)
+    fallback_root = OmegaConf.select(config, "path_shared_working_dir")
+    assert attribute is not None or fallback_root is not None, (
+        "Must specify `attribute` in config if `path_shared_working_dir` is None in config"
+    )
+    attribute = attribute if attribute else fallback_root + fallback
```
MatKbauer, line 364: Add `return attribute` or `return attribute if attribute else fallback_root + fallback` here

src/weathergen/utils/config.py
MatKbauer: Let's also add a trailling underscore, i.e., `_get_private_config_attribute()`, to emphasize this function is meant for use within this `config.py` file only.

MatKbauer (changes requested):
Tested and found some issues that need to be resolved.

src/weathergen/utils/config.py
moritzhauschulz: I am not sure if I agree on this since there is no clear preference for `results` or `models`, so the default behaviour is not obvious. Also, I don't think `=config` would work here since config is undefined. Happy to do it though if you think it's better style.

Issue #746, opened 2025-08-14
Add context manger of NVTX profiling


javak87:
### Is your feature request related to a problem? Please describe.

In order to have a consistent way of NVTX profiling, we need a class feature.

### Organisation

JSC

florianscheidl:
I see the associated PR was closed - shall we close this issue?

clessig:
> I see the associated PR was closed - shall we close this issue?

Do we want need the functionality? I don't think it was merged

florianscheidl:
> > I see the associated PR was closed - shall we close this issue?
> 
> Do we want need the functionality? I don't think it was merged

It wasn't merged. @javak87 could you provide the context since you closed the PR?

javak87:
> > > I see the associated PR was closed - shall we close this issue?
> > 
> > 
> > Do we want need the functionality? I don't think it was merged
> 
> It wasn't merged. [@javak87](https://github.com/javak87) could you provide the context since you closed the PR?

It should be closed since we are not using NVTX context manger for profiling.

PR #747, opened 2025-08-14
add nvtx contex manager

javak87:
## Description
NVTX context manager for Nsight system profiling.

## Type of Change

-   [ ] Documentation update

## Issue Number
https://github.com/ecmwf/WeatherGenerator/issues/746

## Code Compatibility

### Dependencies

### Documentation

## Additional Notes
This is an example how we can use NVTX profiling:
```
with ProfilerSection(name=f"Backward", profile=True):
         self.grad_scaler.scale(loss_values.loss).backward()
```

### DIFF
--- a/src/weathergen/utils/profiler.py
+++ b/src/weathergen/utils/profiler.py
@@ -1,0 +1,40 @@
+import torch
+
+class ProfilerSection:
+    """
+    Context manager for profiling sections of PyTorch code using NVTX ranges.
+
+    Attributes:
+        name (str): The name of the section for profiling.
+        profile (bool): Whether profiling is enabled.
+    """
+
+    def __init__(self, name: str, profile: bool = False):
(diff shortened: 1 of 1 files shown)


src/weathergen/utils/profiler.py
```diff
@@ -0,0 +1,40 @@
+import torch
+
```
tjhunter, line 2: I suggest to have a module flag to turn on or off all profiling. It is more convenient than enabling on case by case basis, and it offers a default.

```py
nvtx_profiler_enabled: bool = false

  def __init__(self, name: str, profile: bool | None = None):
    self._profile: bool = profile if profile is not None else nvtx_profiler_enabled
```

javak87: Sure.
I'll do that!

PR #753, opened 2025-08-15
Channel weighting in loss computation

### LINKED ISSUE #752: Channel weighting for loss computation

By adding channel weights the model can be told to emphasize certain variables more during training, hence forcing the model to improve "more important" variables for example. 

You can define channel weights in the stream config as a list for each variable. 

_No response_

### Additional context

_No response_

### Organisation

ECMWF

Jubeku:
## Description

You can define channel weights in the stream config as a list for each variable and by this force the model to focus on certain channels more during training.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #752 

## Code Compatibility

### Dependencies

### Documentation


src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -143,6 +143,23 @@ def __init__(
                 stream_info[str(self._stage) + "_source_channels"] = ds.source_channels
                 stream_info[str(self._stage) + "_target_channels"] = ds.target_channels
 
+                stream_info["target_channel_weights"] = [
```
clessig: This should go into the data reader (in a separate function) and the code here in multi_stream_data_sample should look like in 143. Please also keep in mind that all data_readers need to have `channel_weights` so it should be defined in the base class (probably as None) and maybe also the parsing can go to data_reader_base.

src/weathergen/datasets/data_reader_base.py
```diff
@@ -292,6 +293,7 @@ def __init__(
 
         self.time_window_handler = tw_handler
         self.stream_info = stream_info
+        self.target_channel_weights = None
```
Jubeku, line 316: FYI, adding initialization to None here so we don't have to modify all other data readers (not sure about best practice).

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -143,6 +143,11 @@ def __init__(
                 # MODIFIES config !!!
                 stream_info[str(self._stage) + "_source_channels"] = ds.source_channels
                 stream_info[str(self._stage) + "_target_channels"] = ds.target_channels
+                stream_info["target_channel_weights"] = (
+                    ds.target_channel_weights
+                    if ds.target_channel_weights is not None
+                    else [1.0 for _ in ds.target_channels]
+                )
```
Jubeku, line 155: This block could be simplified if we define `ds.target_channel_weights` in all data readers, instead of initializing it with None in the data_reader_base.

src/weathergen/datasets/data_reader_base.py
wael-mika: We need to have a look on the design aspect and how to make it work best for the newly added data sets/readers. But as a quick solution, it works!

src/weathergen/train/loss.py
wael-mika, line 1: The weights_channels is None when printing out the values directly:
    loss = torch.mean(loss_chs * weights_channels if weights_channels is not None else loss_chs)
    print(f"channel weights: {weights_channels}")
    print(f"mse loss per channel: {loss_chs[:5]}")
    print(f"mse loss (weighted): {loss}")
channel weights: None
mse loss per channel: tensor([1.1883, 1.6215, 1.6979, 1.1762, 1.1030], device='cuda:0', grad_fn=<SliceBackward0>)
mse loss (weighted): 1.159118890762329

src/weathergen/train/loss_calculator.py
```diff
@@ -94,8 +94,10 @@ def _get_weights(self, stream_info):
             # set loss_weights to 1. when not specified
             stream_info_loss_weight = stream_info.get("loss_weight", 1.0)
             weights_channels = (
-                torch.tensor(stream_info["channel_weight"]).to(device=device, non_blocking=True)
-                if "channel_weight" in stream_info
+                torch.tensor(stream_info["target_channel_weights"]).to(
+                    device=device, non_blocking=True
+                )
+                if "target_channel_weights" in stream_info
```
wael-mika, line 100: "target_channel_weights" is not found in stream_info, rather should be taken from the data reader class

src/weathergen/train/loss.py
Jubeku: What stream config are you testing it with @wael-mika ?

wael-mika: The one you provided, ERA5.yml

src/weathergen/train/loss_calculator.py
Jubeku: It is added to the stream_into [here](https://github.com/ecmwf/WeatherGenerator/blob/jk/develop/ch_weights/src/weathergen/datasets/multi_stream_data_sampler.py#L146), just like source channels and target channels.

wael-mika: I know, but when it comes to loss calculations, the weights are not there!
The only way I got your script to work by adding this line to  parse_target_channel_weights function in the base class
```
self.stream_info["target_channel_weights"] = target_channel_weights
``` 
AND also specifying the target channels and channel weights in the stream config, and they should be MATCHING. This is why we need to find a way to set the weight of a channel to 1 in case if it was not found in the stream config

Jubeku: Thanks Wael for testing, let me look into this!

Jubeku: Hi @wael-mika, I think the problem was that I had the weight parsing commented out in the anemoi reader (probably for testing if it works if other datareaders don't parse the weights).
Can you test again? Thanks :)

wael-mika: Hello @Jubeku, I had it commented out from the first round, and I am testing it again with the line uncommented; the same results persist: channel weights: None. 
The issue is that "target_channel_weights" is not defined anywhere in the code, not in the stream config, not in data reader base or anemoi and not in the multidatasampler. I think you need to update self.stream_info in the parse_target_channel_weights so it contains the newly parsed target_channel_weights and then the channel weights will  be populated as you see below: 
The output with the line
```
self.stream_info["target_channel_weights"] = target_channel_weights
channel weights: tensor([1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 1.0000, 0.8000, 0.5000, 1.5000, 1.2000], device='cuda:0')
```

The output without the line:
```
channel weights: None
```

Jubeku: Thanks Wael. Just making sure we are talking about the same line? https://github.com/ecmwf/WeatherGenerator/blob/jk/develop/ch_weights/src/weathergen/datasets/data_reader_anemoi.py#L114
And you pulled the latest update?

It's weird, I introduced a breakpoint in the loss function and was passed the channel weights without any problem.

wael-mika: Unfortunately, there is no compute time for me at the moment to try the code again. I will try it the first thing in the morning and let you know!

src/weathergen/datasets/data_reader_base.py
```diff
@@ -441,6 +444,28 @@ def get_geoinfo_size(self) -> int:
         """
         return len(self.geoinfo_idx)
 
+    def parse_target_channel_weights(
+        self,
+    ) -> list[float] | None:
+        target_channel_weights = [
+            self.stream_info["channel_weights"].get(ch, 1.0)
```
kacpnowak, line 471: Hi @Jubeku, in this case you are exactly matching channel names to the ones in dictionary with weights. So it's not possible to assign weight to entire class of channels (eg. u_=2.0). Would it be possible to add this feature? I think it would be really helpful.

Jubeku: Thanks @kacpnowak for the suggestion. I think we originally decided to specify the channels explicitly (even though this can result into a long list in the stream config). 
Do others have strong opinions on this design question? cc @clessig, @tjhunter, @grassesi

clessig: Yes, this should be specified explicitly. Having an implicit dependence on the channels in the dataset makes it difficult to read and understand the stream configs. In the future, we will also change this for `source` and `target`, i.e. they (always?) need to be specified explicitly.

Jubeku:
Hi @wael-mika, I added a wiki page for loss weighting options: https://gitlab.jsc.fz-juelich.de/esde/WeatherGenerator-private/-/wikis/home/Model-and-training-configuration/Loss-weighting

PR #648, opened 2025-07-31
Sgrasse/develop/issue 616

grassesi:
## Description

refactors (first 4 commints) and fix #616 

## Type of Change

-   [x] Bug fix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Documentation update

## Issue Number

closes #616 

## Code Compatibility

-   [x] I have performed a self-review of my code

### Code Performance and Testing

-   [x] I ran the `uv run train` and (if necessary) `uv run evaluate` on a least one GPU node and it works 
-   [ ] If the new feature introduces modifications at the config level, I have made sure to have notified the other software developers through Mattermost and updated the paths in the `$WEATHER_GENERATOR_PRIVATE` directory

### Dependencies

-   [ ] I have ensured that the code is still pip-installable after the changes and runs
-   [ ] I have tested that new dependencies themselves are pip-installable.
-   [x] I have not introduced new dependencies in the inference portion of the pipeline

### Documentation

-   [x] My code follows the style guidelines of this project
-   [ ] I have updated the documentation and docstrings to reflect the changes
-   [x] I have added comments to my code, particularly in hard-to-understand areas

PR #755, opened 2025-08-15
Enable correct reading of channels, forecast_step, sample variables in plot config file (Closes #717)

### LINKED ISSUE #717: Allow custom forecast steps in plotting
### What happened?

Actual results: Changed forecast steps in `plot_config.yml` after first run, but it still produced plot with forecast steps from previous config.
Expected results: Changes in config file should produce plot with the corresponding change.

Follow on to #701.

first run:

forecast_steps: [1,2,3]

second run:

forecast_steps: [1]

### Version

acb0be2e2a9ae65f0d9aa68d7beb95c7bb42a48

N/A

### Relevant log output

```shell
N/A
```

### Accompanying data

N/A

### Organisation

JSC

moritzhauschulz:
## Description

PLEASE READ – THIS PR WAS UPDATED SIGNIFICANTLY

Refactor fast evaluation config handling comprehensively. This PR fixes how channels, samples and forecast steps are handled as the three main configurable dimensions of the evaluation within each stream. Previously, an updated config (e.g. adding a new channel/forecast_step/sample) would not be picked up by the code and the samples would be produced based on what is available in the saved json. Now, before producing the evaluation and plots, it is checked 

a) which of the requested channels/forecast_steps/samples are present in the JSON
b) if a channel/forecast_step/sample is not in the JSON, whether it is present in the Zarr file produced at inference

If a), the plots are produced from the given JSON.
If b), the JSON is recomputed for the requested channels based on the Zarr. 
If neither a) or b), an error with explanation is thrown – the config has to be adapted to be a subset of the available data. 

Also fixing a small bug, where a `nan` data array is generated when both `samples` and `forecast_steps` have multiple list entries.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #717 (but arguable the scope has increased a bit – sorry)

## Code Compatibility

### Dependencies

### Documentation

## Additional Notes

@iluise if you are familiar with the 'plot_subtimesteps' argument in the plot config, could you please check if this is still working as expected. I am not sure about the purpose and desired behaviour of this attribute. Otherwise, I mainly tested with variations of the below config:

```
verbose: true
results_dir : "./results/"
output_plotting_dir: "./plots/"
output_scores_dir : "./jsons/"
dpi_val : 300
summary_plots : true
print_summary: false #print out score values on screen. it can be verbose

evaluation:
  metrics  : ["rmse"]

#MTM models
run_ids :
  tskmqw5d:
     streams:
       ERA5:
        channels: ["10v", "10u"]
        evaluation:
          forecast_step: [1,2,3]
          sample: [1,2]
        plotting: 
          sample: [1,2]
          forecast_step: [1,2]
          plot_maps: true
          plot_histograms: true
     label: "MTM ERA5"
     epoch: 0
     rank: 0
```

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/utils.py
+++ b/packages/evaluate/src/weathergen/evaluate/utils.py
@@ -373,10 +373,6 @@ def plot_data(
         cfg, stream, results_dir, mode="plotting"
     )
 
-    # plot_samples = plot_settings.get("sample", None)
-    # plot_fsteps = plot_settings.get("forecast_step", None)
-    # plot_chs = stream_cfg.get("channels")
-
     # Check if maps should be plotted and handle configuration if provided
     plot_maps = plot_settings.get("plot_maps", False)
     if not isinstance(plot_maps, bool):


packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -132,9 +132,7 @@ def evaluate_from_args(argl: list[str]) -> None:
                             )
 
                             # check if channels unchanged from previous config
-                            channels = cfg["run_ids"][run_id]["streams"][stream].get(
-                                "channels"
-                            )
+                            channels = stream_dict.get("channels")
```
clessig: Are we guaranteed that stream_dict has channels? Maybe safer to have `[]` as default for `get()`?

Also stream_dict has `source_channels` and `target_channels` but what is `channels`?

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -144,6 +142,18 @@ def evaluate_from_args(argl: list[str]) -> None:
...
+                            elif sorted(
+                                [
+                                    int(fstep)
+                                    for fstep in stream_dict["evaluation"].get(
+                                        "forecast_step"
+                                    )
+                                ]
+                            ) != sorted(metric_data["forecast_step"].values):
+                                _logger.info(
+                                    "Forecast steps different from previous config. Recomputing if all Forecast steps are available in the Zarr file."
```
clessig: Capital "F"

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -203,8 +204,12 @@ def calc_scores_per_stream(
         f"RUN {run_id} - {stream}: Calculating scores for metrics {metrics}..."
     )
 
-    samples = cfg.evaluation.get("sample", None)
-    fsteps = cfg.evaluation.get("forecast_step", None)
+    samples = cfg.run_ids.get(run_id).streams.get(stream).evaluation.get("sample", None)
+    fsteps = (
```
clessig: Wouldn't the best solution to take the min of requested and available forecast steps. E.g. if I request to process forecast steps [1, ... 8] and only [1, ... , 4] are available it seems best to generate a warning but proceed with what can be done.

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
moritzhauschulz: Thanks Christian. As per your first question, currently, if you leave channels empty (or if the dict does not have channels), it will produce all channels. I think this makes sense, and this can equivalently be achieved by specifying `channels: "all"`. Unless you want that in the case of empty `channels` case we throw an error?

As per your second question, I think since the config is specifically for plotting, it will always only have channels – the output channels to be plotted... I think this is not an issue.

clessig: 1. Ok, fine if there's no error that can be generated if `channels` is not there.

2. In principle there are `source_channels` and `target_channels`. We could match source_channels and target_channels to the channels that are specified in the config, , which can and will be different for various datasets.

clessig: But this should be a differnet PR

moritzhauschulz: Okay. I agree, perhaps this should be a new issue. But to be honest, I don't yet see exactly the problem. Since the plotting works with saved Zarr files which are produced in a prior step. The channels that are specified in the plot config are checked against the channels present in the Zarr file, as far as I understand. Also, as far as I understand, these would be a subset of `target_channels`.

packages/evaluate/src/weathergen/evaluate/utils.py
moritzhauschulz: I implement this in the next commit...

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
clessig: The issue is this PR: https://github.com/ecmwf/WeatherGenerator/pull/648.

moritzhauschulz: Okay, I get the problem. Would be more sensible for this to be fixed in #648, right? Any particular reason why #648 has not been merged yet?

moritzhauschulz: Apart from the issue with channels, I think everything should be fixed now.

mlangguth89: > 1. Ok, fine if there's no error that can be generated if `channels` is not there.
> 
>     2. In principle there are `source_channels` and `target_channels`. We could match source_channels and target_channels to the channels that are specified in the config, , which can and will be different for various datasets.

@clessig @moritzhauschulz @grassesi 
Important note regarding 2):
The xarray-interface gives back a DataArray that has a `channel`-dimension in any case. The distinction between `source_channels` and `target_channels` is done more low-level in the I/O-interface and thus not relevant here.

clessig: Thanks @mlangguth89. And there are separate xarrays for source, target, prediction?

mlangguth89: To go back to the change:
This is indeed a proper, more compact way to access `channels` from `stream_dict`. However, as pointed out by Chrustian, there is no gurantee that `channels`  is provided in the stream configuration as pointed out by Christian. Thus, the code will fail in its current form if omitted. 
However, with an empty list as the default return value for the `get`-method, correct behaviour is not achieved. In this case, missing channels won't be detected at all subsequently. So, we must find a way to get a list of all channels in the stream of the respective inference run, in case `channels` is not provided in the evaluation config-file. Using `peek_tar_channels` might be a way to do so (cf. also its usage in the `get_data`-[method](https://github.com/ecmwf/WeatherGenerator/blob/77f6fd159e5e84220ac58739473c289fa15c5b6e/packages/evaluate/src/weathergen/evaluate/utils.py#L94)).

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -144,6 +142,18 @@ def evaluate_from_args(argl: list[str]) -> None:
                                     f"Channels {missing_channels} do not appear in saved scores for {metric}. Recomputing."
                                 )
                                 metrics_to_compute.append(metric)
+                            elif sorted(
```
mlangguth89: The metric-handling from [l. 123 onwards](https://github.com/ecmwf/WeatherGenerator/blob/69849961c00bb5d724ed1c9ec34c5f124cdd8be9/packages/evaluate/src/weathergen/evaluate/run_evaluation.py#L123) has become very complex due to the various if-clauses. The code in `run_evaluation.py` should be modular. So, I suggest to move the whole metric-handling into a separate method under `utils.py`.

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
clessig: We have target_channels and source_channels for all streams. This should be enough, not?

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -116,53 +117,61 @@ def get_data(
             points_per_sample = None
 
         for fstep in fsteps:
-            _logger.info(f"RUN {run_id} - {stream}: Processing fstep {fstep}...")
-            da_tars_fs, da_preds_fs = [], []
-            pps = []
+            try:
```
mlangguth89: Please keep this PR as focused and self-contained as possible (see comment above).
Handling cases where `forecast_steps` are missing when reading the data is outside the scope of this PR.  
This PR should only address the scenario where the `channels` used for metric evaluation change when running it a second time.

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
mlangguth89: There is no `target_channels` and `source_channels` when reading the data with `as_xarray` (see here: https://github.com/ecmwf/WeatherGenerator/blob/77f6fd159e5e84220ac58739473c289fa15c5b6e/packages/common/src/weathergen/common/io.py#L90). This also does not change with PR #648. So, no need to adapt any logic in the fast evaluation pipeline.

packages/evaluate/src/weathergen/evaluate/utils.py
mlangguth89: I strongely suggest to do this in a separate issue/PR. 
This PR should only address the scenario where the `channels` used for metric evaluation change when running it a second time. The suggested adaption requires re-factoring of the `get_data`-method and likely downstream changes in other routines (creating map plots, histograms and metric computation which all call the `get_data`-method).

packages/evaluate/src/weathergen/evaluate/utils.py
mlangguth89: Much easier than having a try-except construct is to check if `fstep` is part of `zio_forecast_steps` (see [here](https://github.com/ecmwf/WeatherGenerator/blob/77f6fd159e5e84220ac58739473c289fa15c5b6e/packages/evaluate/src/weathergen/evaluate/utils.py#L92C9-L92C28)) which is directly deduced from the inference output.

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -116,53 +117,61 @@ def get_data(
...
-            da_tars_fs = xr.concat(da_tars_fs, dim="ipoint")
-            da_preds_fs = xr.concat(da_preds_fs, dim="ipoint")
+                    da_tars_fs.append(target.squeeze())
+                    da_preds_fs.append(pred.squeeze())
+                    pps.append(npoints)
 
-            if set(channels) != set(all_channels):
                 _logger.debug(
-                    f"Restricting targets and predictions to channels {channels} for stream {stream}..."
+                    f"Concatenating targets and predictions for stream {stream}, forecast_step {fstep}..."
```
mlangguth89: We are not concatenating over channels here, but filtering. So please don't change the logging-message.

moritzhauschulz: Thanks. Not sure how this got there, might have been a merge conflict.

moritzhauschulz:
Not sure why I can't reply directly @mlangguth89 to this https://github.com/ecmwf/WeatherGenerator/pull/755#discussion_r2284316016

I don't fully understand why you quote this section of the code. In the section you quote, this issue was that `cfg.evaluation.get("sample", None)` always returned none, since `evaluation` is a nested attribute at the stream level... So I think my change made sense. But please let me know if I am missing the point here.

moritzhauschulz:
Moved to draft. Not ready for review...

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -193,6 +193,8 @@ def create_histograms_per_sample(
             # Remove NaNs
             targ = targ.dropna(dim="ipoint")
             prd = prd.dropna(dim="ipoint")
+            assert targ.size > 0, "Data array must not be empty or contain only NAs"
```
iluise, line 196: can we use the logger and throw an error trough the logger if tag.size == 0 or pre.size == 0?
smth like: 

`After dropping all NaNs the dataset is empty. Please check.`

packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -370,6 +371,9 @@ def create_maps_per_sample(
                 if valid_time is not None:
                     _logger.debug(f"Plotting map for {var} at valid_time {valid_time}")
 
+                da_t = da_t.dropna(dim="ipoint")
+                assert da_t.size > 0, "Data array must not be empty or contain only NAs"
```
iluise, line 375: same here, can we pass through the logger?

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -125,7 +126,9 @@ def evaluate_from_config(cfg):
 
             if stream_dict.get("plotting"):
                 _logger.info(f"RUN {run_id}: Plotting stream {stream}...")
-                _ = plot_data(cfg, cf_run, results_dir, runplot_dir, stream)
+                _ = plot_data(
+                    cfg, run_id, run, results_dir, runplot_dir, stream, stream_dict
```
iluise: we will change it in another pr, but for now we need to stick to the old convention and pass also the cf_run to get the time steps correctly. we should open an issue about it

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -143,30 +146,28 @@ def evaluate_from_config(cfg):
                                 metric,
                                 run.epoch,
                             )
-
-                            # check if channels unchanged from previous config
-                            channels = cfg["run_ids"][run_id]["streams"][stream].get(
-                                "channels"
+                            checked, (channels, fsteps, samples) = check_availability(
+                                cfg, run_id, run, stream, results_dir, metric_data
```
iluise: you don't actually need cfg, run_id and run all together. you can retrieve run from cfg and run_id

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -176,7 +177,9 @@ def get_data(
                 da_preds.append(da_preds_fs)
             if return_counts:
                 points_per_sample.loc[{"forecast_step": fstep}] = np.array(pps)
-        
+
+        fsteps_final = list(sorted(set(fsteps_final)))
```
iluise: this should have been solved by: https://github.com/ecmwf/WeatherGenerator/issues/852. can you remove it? If people do not specify the forecast steps in the right order this will silently re-shuffle the fsteps and `da_tars` and `da_preds` below will have incorrect fsteps vs data matching

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -304,40 +309,41 @@ def calc_scores_per_stream(
...
+    stream_cfg:
+        Stream sub-config.
     Returns
     -------
     List of plot names generated during the plotting process.
     """
-    run_id = results_dir.name
-
-    # get stream dict from evaluation config (assumed to be part of cfg at this point)
-    stream_dict = cfg["run_ids"][run_id]["streams"][stream]
```
iluise, line 339: we should restore the old stream_dict for now

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -355,16 +361,18 @@ def plot_data(
         "image_format": cfg.get("image_format", "png"),
         "dpi_val": cfg.get("dpi_val", 300),
         "fig_size": cfg.get("fig_size", (8, 10)),
-        "plot_subtimesteps": get_stream_attr(
-            run_config, stream, "tokenize_spacetime", False
-        ),
+        "plot_subtimesteps": stream_cfg.get("tokenize_spacetime", False),
```
iluise: this is not in the stream_cfg, but in the run config. we need to retrieve it from the attrs "tokenize_spacetime" in the run_config unfortunately. see my branch

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -355,16 +361,18 @@ def plot_data(
...
+    plotter = Plotter(plotter_cfg, plot_base_dir)
+
+    check, (plot_chs, plot_fsteps, plot_samples) = check_availability(
+        cfg, run_id, run_cfg, stream, results_dir, eval=False
+    )
 
-    plot_samples = plot_settings.get("sample", None)
-    plot_fsteps = plot_settings.get("forecast_step", None)
-    plot_chs = stream_dict.get("channels")
+    # plot_samples = plot_settings.get("sample", None)
```
iluise: remove commented lines

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -827,25 +836,192 @@ def scalar_coord_to_dim(da: xr.DataArray, name: str, axis: int = -1) -> xr.DataA
...
-    for stream in config.get("streams", []):
-        if stream.get("name") == stream_name:
-            return stream.get(key, default)
-    return default
+
+    channels, fsteps, samples = _get_channels_fsteps_samples(
+        cfg, run_id, stream, eval=eval
+    )
+
+    channels = set(channels) if channels is not None else None
```
iluise: with set we loose the order. This is used only in this function to compare with Zio, and not in the rest of the code, right?

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -827,25 +836,192 @@ def scalar_coord_to_dim(da: xr.DataArray, name: str, axis: int = -1) -> xr.DataA
...
+        if metric_data is not None and zio_samples != available_samples:
+            _logger.info(
+                "Requested all samples, but previous config was a strict subset of samples in Zarr. Recomputing."
+            )
+            return False, (channels, fsteps, samples)
+
+    # config must be a subset of zio, otherwise error
+    if not channels <= zio_channels:
+        raise ValueError(
+            f"Requested channel(s) {channels - zio_channels} do(es) not exist in the Zarr file. Channels must be a subset of {zio_channels}"
```
iluise: it should fail silently and just throw a warning saying that these Charles will be excluded as they are not in the zarr file. same for forecast steps and samples. the syntax here https://github.com/ecmwf/WeatherGenerator/tree/iluise/test717 should work, but please test it!

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -827,25 +836,192 @@ def scalar_coord_to_dim(da: xr.DataArray, name: str, axis: int = -1) -> xr.DataA
...
+
+    with ZarrIO(fname_zarr) as zio:
+        zio_fsteps = zio.forecast_steps
+        zio_samples = zio.samples
+        zio_channels = peek_tar_channels(zio, stream, zio_fsteps[0])
+        zio_fsteps = set([int(fstep) for fstep in zio_fsteps])
+        zio_samples = set([int(sample) for sample in zio_samples])
+        zio_channels = set(zio_channels)
+
+    # available must equal zio, otherwise recompute
```
iluise: I suggest a more compact solution in branch https://github.com/ecmwf/WeatherGenerator/tree/iluise/test717

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -143,30 +146,28 @@ def evaluate_from_config(cfg):
...
+                                    )
                                 )
-
-                        # TODO update retrieve_metric_from_json to avoid having to catch errors
-                        except (FileNotFoundError, KeyError, ValueError):
+                        except (FileNotFoundError, KeyError):
                             metrics_to_compute.append(metric)
 
                     if metrics_to_compute:
+                        checked, _ = check_availability(
```
iluise: I am a bit confused, why do we need to re-check it here if it was already checked above?

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -34,6 +33,7 @@ class WeatherGeneratorOutput:
     points_per_sample: xr.DataArray | None
 
 
+# TODO: This function needs some careful refactoring.
```
iluise, line 36: I agree, can you open an issue for that? do you have suggestions?

packages/evaluate/src/weathergen/evaluate/plotter.py
clessig: The alternative to the assert would be to throw a ValueException, not the logger (it wouldn't abort). But we agreed that assert's are the way to handle it for the time being, though I think we should change.

packages/evaluate/src/weathergen/evaluate/plotter.py
clessig: Same as above.

packages/evaluate/src/weathergen/evaluate/plotter.py
moritzhauschulz: Thanks. So I keep it as assert for the time being? I think a logger message is too easily overlooked. I placed the assert there because there was actually a bug in the old code that lead to nans, and it took me 5+ hours to figure out where it came from. So aborting here seems reasonable in my opinion.

iluise: yes, oks, let's keep the assert

packages/evaluate/src/weathergen/evaluate/utils.py
moritzhauschulz: This actually addresses another issue. I think it is unrelated to #852. Basically, the loop preceding this assignment can cause duplicate entries in final_fsteps, which then in turn causes a mismatch in the zipping that follows. This is necessary here, but can also be addressed in another way if you have a preference.

packages/evaluate/src/weathergen/evaluate/utils.py
moritzhauschulz: Since we return it also, the order will be lost. I think this does not cause any issue thought since at least in get_data the list is sorted anyways. I may be wrong though, so please let me know where a problem could arise. I will push an update shorty that changes the handling sightly, so please wait for this before looking into this. Thank you.

packages/evaluate/src/weathergen/evaluate/utils.py
moritzhauschulz: Makes sense, I have to adapt the code slightly though for 'calc_scores_per_stream' to correctly handle this. Will be in the next push.

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
moritzhauschulz: It is technically necessary because it is a separate loop, but I moved this now into calc_scores_per_stream. Otherwise we need to refactor a lot in `evaluate_from_config` which I would prefer not to do in this PR. Hope this is okay (see next push).

packages/evaluate/src/weathergen/evaluate/utils.py
moritzhauschulz: I think I fixed some issues with the function since I wrote this. I think with our check_availability function, some of the checks have become outdated so it could be shortened quite a bit.

packages/evaluate/src/weathergen/evaluate/utils.py
iluise: well if you look at the fix: https://github.com/ecmwf/WeatherGenerator/pull/853/files I think this is solved now, right? It was indeed a bug in the implementation and we were storing fstep multiple times, but it should be fixed now

packages/evaluate/src/weathergen/evaluate/utils.py
iluise: Thanks for opening the other issue! please ping me when you have the new version :)

packages/evaluate/src/weathergen/evaluate/utils.py
iluise: uhm, I always prefer to keep the order, but I am fine with sets as soon as we check that in get_data we assign the correct fstep to the correct data.

moritzhauschulz: I am not super familiar with xarrays/zarr IO, do you think you could check this perhaps? As I said, in get_data some reordering (with`sorted()`) is already done for fsteps and samples (I think since your first version), so as long as this is valid, our change on top of this should also be valid. To make sure, I guess we could put another sort at the end of `check_availability` – should I just do that?

packages/evaluate/src/weathergen/evaluate/utils.py
moritzhauschulz: Ahh I see now. Sorry, I was not careful here. Your implementation is definitely safer. I will merge yours and remove mine. Thanks for spotting this.

packages/evaluate/src/weathergen/evaluate/utils.py
iluise: yes please!

packages/evaluate/src/weathergen/evaluate/utils.py
moritzhauschulz: Is this resolved? not sure what you mean...

packages/evaluate/src/weathergen/evaluate/utils.py
moritzhauschulz: Should be resolved, thanks.

PR #756, opened 2025-08-15
Combine masking strategies during training, with appropriate masking_…

### LINKED ISSUE #757: Combine masking strategies during training
During training, sample a masking strategy according to some probability chosen by the user. Therefore mix different masking strategies during training.

shmh40:
…strategy_config

NOTE: default and era5 config have been changed for ease of testing, they will be reverted before merge.

## Description

Combine masking strategies during training, each sample has a different strategy which is randomly sampled according to a user specified probability in masking_strategy_config.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #757

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/datasets/masking.py
+++ b/src/weathergen/datasets/masking.py
@@ -56,12 +56,9 @@ class Masker:
         self.perm_sel: list[np.typing.NDArray] = None
 
         # Per-batch strategy tracking
-        # masking_strategy_per_batch is the strategy
-        # used for the current batch.
         self.same_strategy_per_batch = self.masking_strategy_config.get(
             "same_strategy_per_batch", False
         )
-        self.masking_strategy_per_batch = None
         self.batch_strategy_set = False
 
         # Check for required masking_strategy_config at construction time


config/streams/streams_anemoi/era5.yml
```diff
@@ -11,7 +11,7 @@ ERA5 :
   type : anemoi
   filenames : ['aifs-ea-an-oper-0001-mars-o96-1979-2022-6h-v6.zarr']
   source_exclude : ['w_', 'skt', 'tcw', 'cp', 'tp']
-  target_exclude : ['w_', 'slor', 'sdor', 'tcw', 'cp', 'tp']
```
sophie-xhonneux: this change should be a different PR maybe

sophie-xhonneux (approved):
Looks good to me, we should think about refactoring such that we don't create a new class everytime for a new batch

config/streams/streams_anemoi/era5.yml
shmh40: Yes sorry, just put here for ease of testing with the channels masking

config/default_config.yml
```diff
@@ -88,13 +88,16 @@ masking_rate: 0.6
 masking_rate_sampling: True
 # sample a subset of all target points, useful e.g. to reduce memory requirements (also can specify per-stream)
 sampling_rate_target: 1.0
-# include a masking strategy here, currently only supporting "random", "block", "healpix" and "channel"
+# include a masking strategy here, currently only supporting "random", "block", "healpix", "channel", "combination"
```
clessig, line 97: Why not have `masking_strategy` as a list. Then we wouldn't need `strategies` below.

clessig: Or I think we should remove `masking_strategy` entirely and just use `masking_strategy_config`. And we should think a bit more what a good strategy for it is.

src/weathergen/datasets/masking.py
```diff
@@ -132,6 +133,17 @@ def mask_source(
             source_data = [data[~p] for data, p in zip(tokenized_data, self.perm_sel, strict=True)]
             return source_data
 
+        # if we have masking_strategy "combination"
+        # at each pass we will sample a different masking_strategy
```
clessig: What do you mean by pass here?

clessig (changes requested):

src/weathergen/datasets/masking.py
shmh40: I mean per-batch and per-stream, we call batchify_source and target in multi_stream_data_sampler. This raises a point - I like having a different strategy per-batch, but we should probably have the same strategy per-stream for the same batch right? Will work on implementing this.

config/default_config.yml
shmh40: I agree, we should just use masking_strategy_config. I have some ideas about how to rewrite this to make it much more flexible, particularly specifying a masking_rate for each individual strategy (relevant for the combination case), rather than taking a global masking_rate for all, and having the strategy-specific arguments being cleanly associated with a particular strategy.

I think this should be in a separate PR though. Can we merge this one first (and the causal one, removing the current causal_masking_rate), and then it will be easier to do a separate PR "restructure masking strategy config" that addresses all these cases in one go?

src/weathergen/datasets/masking.py
shmh40: Or have an option to do this per-stream too or not

config/default_config.yml
```diff
@@ -94,13 +94,16 @@ masking_rate: 0.6
...
 masking_strategy: "random"
 # masking_strategy_config is a dictionary of additional parameters for the masking strategy
 # required for "healpix" and "channel" masking strategies
 # "healpix": requires healpix mask level to be specified with `hl_mask`
-# "channel": requires "mode" to be specified, "per_cell" or "global", 
-masking_strategy_config: {"hl_mask": 3}
+# "channel": requires "mode" to be specified, "per_cell" or "global",
+masking_strategy_config: {"strategies": ["random", "healpix", "channel"], 
+                          "probabilities": [0.34, 0.33, 0.33],
```
sophie-xhonneux, line 104: can we set the masking rate here for each strategy as well?

src/weathergen/datasets/masking.py
```diff
@@ -39,6 +39,7 @@ class Masker:
     def __init__(self, cf: Config):
         self.masking_rate = cf.masking_rate
         self.masking_strategy = cf.masking_strategy
```
sophie-xhonneux, line 41: maybe we can call this `sampled_masking_strategy`?

src/weathergen/datasets/masking.py
shmh40: @clessig I have now written this so that for each batch, we sample the same masking strategy for every stream in that batch. If you could have a quick look that would be great, then I can rewrite the masking_strategy_config early next week.

Separately this also raises the question of whether we should have the *same* random or healpix mask for every stream in each batch. At the moment, the different streams have different randomly sampled masks. I suppose this will affect whether the model primarily learns cross-stream or single-stream correlations. Currently with the different masks per-stream, it seems likely that for e.g. a masked cell of ERA5, instead  e.g. mostly using the unmasked ERA5 cells, it may instead learn to use perhaps an unmasked cell of SYNOP that is the same cell that was masked in the ERA5 stream.

sophie-xhonneux: I think both have their merit (thinking out loud now:
- if a cell has sometimes ERA5 and sometimes CERRA it will create some resilience if we don't have have either ERA5 or CERRA at inference time
- if a cell has nothing, we need to learn the spatial correlation between cells, which probably gives good latents

given our high masking rate, we might get a bit of of both? is that true given the high number of points per cell for say CERRA?

config/default_config.yml
shmh40: Next PR!

src/weathergen/datasets/masking.py
shmh40: We could do, I think I will leave it for now though as sometimes it is not sampled and we still use the same variable. If it is really unclear let me know.

config/default_config.yml
sophie-xhonneux: Okay!

src/weathergen/datasets/masking.py
sophie-xhonneux: I think it is confusing to me, but I might be alone in this. At its core in my head, the name in the config should be the same in the class and not change, and the variable that changes should have an alternate name, each masking_strategy_this_batch or something

src/weathergen/datasets/masking.py
clessig: I think the masking strategies should be independent per stream. A key part of the pre-training is to learn the relationship/correlation between the streams and then having non-aligned missing pieces across streams really helps.

PR #760, opened 2025-08-17
Fix IO when targets/preds are empty.

### LINKED ISSUE #759: Inference output cannot handle empty targets/predictions
### What happened?

Inference I/O fails when some output streams are empty.

Use, e.g, a network with len_hrs=3 and step_hrs=3 and trained with ERA5 and CERRA (6-hourly)

### Version

develop

ALL

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

clessig:
## Description

Fix IO when targets/preds are empty.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #759 

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/packages/common/src/weathergen/common/io.py
+++ b/packages/common/src/weathergen/common/io.py
@@ -59,7 +59,7 @@ class OutputDataset:
     item_key: ItemKey
 
     # (datapoints, channels, ens)
-    data: zarr.Array  # wrong type => arary like
+    data: zarr.Array  # wrong type => array like
 
     # (datapoints,)
     times: zarr.Array
@@ -125,16 +125,16 @@ class OutputDataset:
 class OutputItem:
     def __init__(
         self,
-        target: OutputDataset,
-        prediction: OutputDataset,
+        key: ItemKey,
+        target: OutputDataset | None = None,
+        prediction: OutputDataset | None = None,
         source: OutputDataset | None = None,
     ):
         """Collection of possible datasets for one output item."""
+        self.key = key
         self.target = target
         self.prediction = prediction
         self.source = source
-
-        self.key = self.target.item_key
 
         self.datasets = [self.target, self.prediction]
 
@@ -181,6 +181,7 @@ class ZarrIO:
             name: OutputDataset(name, key, **dict(dataset.arrays()), **dataset.attrs)
             for name, dataset in group.groups()
         }
+        datasets["key"] = key
 
         return OutputItem(**datasets)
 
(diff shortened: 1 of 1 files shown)


packages/common/src/weathergen/common/io.py
```diff
@@ -318,19 +325,25 @@ def extract(self, key: ItemKey) -> OutputItem:
...
+                self.targets[forecast_step][stream_idx][0][datapoints].cpu().detach().numpy()
+            )
+            preds_data = (
+                self.predictions[forecast_step][stream_idx][0]
+                .transpose(1, 0)
+                .transpose(1, 2)[datapoints]
+                .cpu()
+                .detach()
+                .numpy()
+            )
```
grassesi, line 342: Instead of inventing values here, I would suggest we just set `OutputItem(source=source_dataset, target=None, prediction=None)`. This will save a lot of code and clearly indicate that there are no valid predictions/targets for this `ItemKey`.

packages/common/src/weathergen/common/io.py
```diff
@@ -303,8 +304,14 @@ def extract(self, key: ItemKey) -> OutputItem:
...
-        start = sum(lens[:sample])
-        n_samples = lens[sample]
+
+        # empty target/prediction
+        if len(lens) == 0:
+            start = 0
+            n_samples = 0
+        else:
+            start = sum(lens[:sample])
+            n_samples = lens[sample]
```
grassesi, line 314: this is not necessary if the suggestion below is followed.

packages/common/src/weathergen/common/io.py
```diff
@@ -125,17 +125,17 @@ def as_xarray(self, chunk_nsamples=CHUNK_N_SAMPLES) -> xr.Dataset:
 class OutputItem:
     def __init__(
         self,
-        target: OutputDataset,
-        prediction: OutputDataset,
+        key: ItemKey,
+        target: OutputDataset | None = None,
+        prediction: OutputDataset | None = None,
```
grassesi, line 130: Since we add datasets 0 sized arrays in this case, this should never be `None` (except there is a different case where `target`/`prediction` could be none => but than it would need to be properly handled by `ZarrIO._write_dataset`)

packages/common/src/weathergen/common/io.py
```diff
@@ -125,17 +125,17 @@ def as_xarray(self, chunk_nsamples=CHUNK_N_SAMPLES) -> xr.Dataset:
...
         source: OutputDataset | None = None,
     ):
         """Collection of possible datasets for one output item."""
+        self.key = key
         self.target = target
         self.prediction = prediction
         self.source = source
 
-        self.key = self.target.item_key
-
```
grassesi, line 138: since we guarantee `target` to be present (even if its values are zero-sized) we can always grab the key from there

packages/common/src/weathergen/common/io.py
```diff
@@ -181,6 +181,7 @@ def load_zarr(self, key: ItemKey) -> OutputItem:
             name: OutputDataset(name, key, **dict(dataset.arrays()), **dataset.attrs)
             for name, dataset in group.groups()
         }
+        datasets["key"] = key
```
grassesi, line 184: see above

packages/common/src/weathergen/common/io.py
```diff
@@ -381,6 +394,7 @@ def extract(self, key: ItemKey) -> OutputItem:
             source_dataset = None
 
         return OutputItem(
+            key=key,
```
grassesi, line 397: see above

Issue #762, opened 2025-08-18
Stream name with non-alphanumeric character causes crash


kacpnowak:
### What happened?

When steam name has non-alphanumeric character (eg. `FESOM_NODE`) it's being cleaned and before being uses as a key here: https://github.com/ecmwf/WeatherGenerator/blob/6a16363956f6c8b08ec95a01c00d3f9a33584660/src/weathergen/utils/train_logger.py#L123

This causes a crash as it was not treated in this way when creating `losses_all` dictionary, thus such key doesn't exist. Removing call of this function fixes the issue 

### What are the steps to reproduce the bug?

On levante run: 

`uv run train --config=/work/ab0995/a270225/WeatherGenerator2/da_best_config_ever.yml  --private_config=/work/ab0995/a270225/WeatherGenerator-private/`

### Version

0.0

### Platform (OS and architecture)

Levante

### Relevant log output

```shell
Traceback (most recent call last):
  File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/run_train.py", line 170, in train_with_args
    trainer.run(cf)
  File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/train/trainer.py", line 322, in run
    self.train(epoch)
  File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/train/trainer.py", line 506, in train
    self._log(TRAIN)
  File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/train/trainer.py", line 697, in _log
    self.train_logger.add_train(
  File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/utils/train_logger.py", line 124, in add_train
    loss = losses_all[st_name]
           ~~~~~~~~~~^^^^^^^^^
KeyError: 'FESOMNODE'
> /work/ab0995/a270225/WeatherGenerator2/src/weathergen/utils/train_logger.py(124)add_train()
-> loss = losses_all[st_name]
(Pdb) print(list(losses_all.keys()))
['FESOM_NODE', 'FESOM_ELEM', 'IFS_ATMO']
```

### Organisation

AWI

PR #763, opened 2025-08-18
Remove cleaning stream name when logging loss

kacpnowak:
## Description

Removes cleaning stream names when logging losses 

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/762

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/datasets/masking.py
+++ b/src/weathergen/datasets/masking.py
@@ -74,7 +74,9 @@ class Masker:
                 # check explicit includes
                 source_include = stream.get("source_include", [])
                 target_include = stream.get("target_include", [])
-                assert set(source_include) == set(target_include), ()
+                assert set(source_include) == set(target_include), (
+                    "Source and target channels not identical. Required for masking_mode=channel"
+                )
                 # check excludes
                 source_exclude = stream.get("source_exclude", [])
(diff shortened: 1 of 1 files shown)


src/weathergen/utils/train_logger.py
```diff
@@ -166,7 +166,7 @@ def add_val(
         log_vals += [samples]
 
         for st in self.cf.streams:
-            st_name = _clean_name(st["name"])
+            st_name = st["name"]
```
clessig, line 169: Does this work with all stream names, e.g. `NPP, ATMS. `losses_all` comes from model and we need to match the key that was used there. Since `_clean_name` is defined here, I can see that there's a bug at the moment. But it would be important to check that we really don't break anything else, including `plot_train``

kacpnowak: This change shouldn't affect plotting. This change only impacts retrival of values from dictionary as the names are cleaned separately when logging https://github.com/ecmwf/WeatherGenerator/blob/6a16363956f6c8b08ec95a01c00d3f9a33584660/src/weathergen/utils/train_logger.py#L437

PR #794, opened 2025-08-21
Modify DataReaderObs to get base_yyyy... from stream config

### LINKED ISSUE #793: Move base_yyyy... to constructor in DataReaderObs for obs datasets with different base dates
Currently the base date of the index of zarr data for DataReaderObs is hard coded as 197001010000. This should be passed via the stream config for different obs datasets with different base dates (e.g. long obs datasets with records back to 1700 and something. 

stream_info is passed already, and we need a sensible default of 197001010000 as is widely used for other obs datasets.

shmh40:
…it in the ctor, with default of 19700101. Use it in _setup_sample_index. Remove loading obs_id attr. Add igra.yml with example usage.

## Description

base_yyyymmddhhmm was hard coded in DataReaderObs. This is now passed to the class from the stream config and set in the constructor, and used to read in the hrly_index of the Zarr, and used in _setup_sample_index.

In this PR, we also remove loading the obs_id attribute from the Zarr as it is unused, and we add igra.yml which shows example usage of base_yyyymmddhhmm in a stream config.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #793 

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/config/streams/igra/igra.yml
+++ b/config/streams/igra/igra.yml
@@ -1,7 +1,11 @@
-# obs_types
-# 0 : polar orbiting satellites
-# 1 : geostationay satellites
-# 2 : conventional observations
+# (C) Copyright 2025 WeatherGenerator contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
 
 IGRA :
   type : obs
--- a/config/streams/igra/igra.yml
+++ b/config/streams/igra/igra.yml
@@ -10,7 +10,7 @@
 IGRA :
   type : obs
   filenames : ['igra.zarr']
-  base_yyyymmddhhmm : 175001010000
+  base_datetime : '1750-01-01T00:00:00'
   loss_weight : 1.0
   #masking_rate_none : 0.05
   token_size : 64
--- a/config/streams/igra/igra.yml
+++ b/config/streams/igra/igra.yml
@@ -1,0 +1,30 @@
+# obs_types
+# 0 : polar orbiting satellites
+# 1 : geostationay satellites
+# 2 : conventional observations
+
+IGRA :
+  type : obs
+  filenames : ['igra.zarr']
+  base_yyyymmddhhmm : 175001010000
+  loss_weight : 1.0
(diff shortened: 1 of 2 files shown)


src/weathergen/datasets/data_reader_obs.py
```diff
@@ -201,7 +199,6 @@ def _load_properties(self) -> None:
 
         self.properties["means"] = self.data.attrs["means"]
         self.properties["vars"] = self.data.attrs["vars"]
-        self.properties["obs_id"] = self.data.attrs["obs_id"]
```
shmh40, line 204: We remove the obs_id here - @clessig I hope this is ok to do now.

config/streams/igra/igra.yml
```diff
@@ -0,0 +1,34 @@
+# (C) Copyright 2025 WeatherGenerator contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
```
shmh40, line 8: Just including an example of how base_datetime is used here

src/weathergen/datasets/data_reader_obs.py
clessig: Yes, remove

config/streams/igra/igra.yml
```diff
@@ -0,0 +1,34 @@
...
+
+IGRA :
+  type : obs
+  filenames : ['igra.zarr']
+  base_yyyymmddhhmm : 175001010000
```
clessig: I think it would be better to use

base_datetime: 1750-01-1T00:00:00

We will enforce ISO datetime format at some point but I don't see much benefit using the old format here.

src/weathergen/datasets/data_reader_obs.py
```diff
@@ -33,7 +33,8 @@ def __init__(self, tw_handler: TimeWindowHandler, filename: Path, stream_info: d
         self.z = zarr.open(filename, mode="r")
         self.data = self.z["data"]
         self.dt = self.z["dates"]  # datetime only
-        self.hrly_index = self.z["idx_197001010000_1"]
+        self.base_yyyymmddhhmm = stream_info.get("base_yyyymmddhhmm", 197001010000)
```
clessig: Would need to be adjusted if ISO format is used. Next line should do the conversion from str to datetime.

PR #798, opened 2025-08-21
Implement causal masking as MTM strategy

### LINKED ISSUE #797: Implement causal masking
Implement a causal masking strategy, where later timesteps of tokenized_data are masked, and predicted from earlier timesteps, with the fraction masked set by the masking rate.

shmh40:
## Description

Implement causal masking, where a fraction of the later timesteps in tokenized_data are masked out and predicted from the earlier timesteps as a masked token modelling strategy.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #797 

## Code Compatibility

### Dependencies

### Documentation


src/weathergen/datasets/masking.py
```diff
@@ -132,6 +133,17 @@ def mask_source(
             source_data = [data[~p] for data, p in zip(tokenized_data, self.perm_sel, strict=True)]
             return source_data
 
+        # if we have masking_strategy "combination"
```
shmh40: NOTE: this also contains the code for the combination strategy

src/weathergen/datasets/masking.py
```diff
@@ -386,3 +414,56 @@ def _generate_channel_mask(
...
+                full_mask.append(np.array([], dtype=bool))
+            elif len(token_data) == 1:
+                # for 1 timestep, don't mask target
+                full_mask.append(np.zeros(1, dtype=bool))
+            else:
+                # create temporal mask
+                num_time_windows = len(token_data)
+                mask = np.zeros(num_time_windows, dtype=bool)
+                # mask the final time windows based on rate
+                num_future_to_mask = int(rate * num_time_windows)
```
shmh40: This could be floor or ceiling, or as it is.

src/weathergen/datasets/masking.py
```diff
@@ -386,3 +414,56 @@ def _generate_channel_mask(
...
+                # set the mask to True for the final time windows
+                start_mask_idx = num_time_windows - num_future_to_mask
+                # force start_mask_idx is at least 1, keep one source time
+                start_mask_idx = max(1, start_mask_idx)
```
shmh40: Here we make sure that we have at least one timestep of token unmasked (and hence in source), so there is always some input to predict future times of the tokens, regardless of how high the masking rate is

config/streams/era5_1deg/era5.yml
```diff
@@ -34,4 +34,4 @@ ERA5 :
     # sampling_rate : 0.2
   pred_head :
     ens_size : 1
-    num_layers : 1
+    num_layers : 1
```
sophie-xhonneux, line 37: remove, unless we want to change the default config?

src/weathergen/datasets/masking.py
```diff
@@ -214,7 +239,6 @@ def mask_target(
 
         processed_target_tokens = []
 
-        # process all healpix cells used for embedding
```
sophie-xhonneux, line 265: why did we remove this comment?

src/weathergen/datasets/masking.py
```diff
@@ -386,3 +414,56 @@ def _generate_channel_mask(
...
+    ) -> list[np.typing.NDArray]:
+        """
+        Generates a causal mask, masking the latest times
+        in each tokenized_data according to the masking rate.
+
+        Args:
+            tokenized_data (list[torch.Tensor]): A list of tensors. Most will have a shape of
+                                                (time, num_tokens, num_channels), but some may
+                                            be empty with a shape of (0,), no data in cell
+            rate (float): The desired masking rate.
```
sophie-xhonneux: Is this a separate masking rate, i.e. different from the spatial masking rate, if not I think it should be!

src/weathergen/datasets/masking.py
sophie-xhonneux: I think floor makes the most sense, easiest to reason about, in the worst case we simply don't mask causally

config/streams/era5_1deg/era5.yml
shmh40: No actual difference as far as I can tell :/

src/weathergen/datasets/masking.py
shmh40: They are really the tokens, not the cells

sophie-xhonneux: can we adapt the comment then rather than remove it?

src/weathergen/datasets/masking.py
shmh40: Nice

src/weathergen/datasets/masking.py
shmh40: done

src/weathergen/datasets/masking.py
shmh40: Done!

config/default_config.yml
```diff
@@ -83,12 +83,15 @@ freeze_modules: ""
 training_mode: "masking"
 # masking rate when training mode is "masking"; ignored in foreacast mode
 masking_rate: 0.6
+# separate masking rate for strategy "causal"
+causal_masking_rate : 0.6
```
clessig: Why can't this go into the masking_config dict? And why do we need a separate rate for causal? Is this the "pivot" for causality?

config/default_config.yml
```diff
@@ -83,12 +83,15 @@ freeze_modules: ""
...
 # sample the masking rate (with normal distribution centered at masking_rate)
 # note that a sampled masking rate leads to varying requirements
 masking_rate_sampling: True
+causal_masking_rate_sampling: True
```
clessig: Again, can't we put this into the masking config?

src/weathergen/datasets/masking.py
```diff
@@ -38,8 +38,13 @@ class Masker:
 
     def __init__(self, cf: Config):
         self.masking_rate = cf.masking_rate
+        self.causal_masking_rate = cf.get("causal_masking_rate", cf.masking_rate)
```
clessig: If we have a config dict then we could probably keep the params just there.

src/weathergen/datasets/masking.py
```diff
@@ -170,6 +189,16 @@ def mask_source(
                 else:
                     source_data.append(data)
 
+        elif self.masking_strategy == "causal":
+            # Only select unmasked timesteps
```
clessig, line 228: What happens for data that doesn't have time steps?

src/weathergen/datasets/masking.py
```diff
@@ -170,6 +189,16 @@ def mask_source(
...
+            self.perm_sel = mask
+            source_data = []
+            for data, p in zip(tokenized_data, self.perm_sel, strict=True):
+                if len(data) > 0:
```
clessig: `source_data.append(data[~p] if len(data)>0 else data)`

src/weathergen/datasets/masking.py
```diff
@@ -386,3 +428,56 @@ def _generate_channel_mask(
...
+                            in tokenized_data.
+        """
+
+        if not tokenized_data:
+            return []
+
+        rate = self._get_sampling_rate()
+
+        full_mask = []
+        for token_data in tokenized_data:
```
clessig: This loop is over all healpix cells? I think we should vectorize

src/weathergen/datasets/masking.py
shmh40: We simply include the data as source, with no corresponding target from that stream in the case of just one time step. Else we just append empty data.

Issue #749, opened 2025-08-14
logger breaks in notebook


clessig:
### What happened?

assert type(ostr) is str or type(ostr) is pathlib.Path, however, the ostr is of the type `ipykernel.iostream.OutStream'

### What are the steps to reproduce the bug?

Run in notebook

### Version

develop

### Platform (OS and architecture)

ALL

### Relevant log output

```shell

```

PR #803, opened 2025-08-22
Train continue function with arguments

### LINKED ISSUE #802: train continue with args

Currently, we cannot call train_continue as a function with args.

Add a parser function which parses the cli arguments to a function with arguments.

_No response_

### Additional context

_No response_

### Organisation

ECMWF

Jubeku:
## Description

Currently, we cannot call train_continue as a function with args.
This PR adds a parser function which parses the cli arguments to a function with arguments.

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #802 

## Code Compatibility

### Dependencies

### Documentation


src/weathergen/run_train.py
```diff
@@ -113,6 +127,10 @@ def train_continue() -> None:
     )
     cf = config.set_run_id(cf, args.run_id, args.reuse_run_id)
 
+    print("-----------------")
```
clessig: This should be `logger.info()`. But we have plenty of logging statements in the config code, including which run_id is loaded etc. I don't think it should be here. Otherwise we would need to also add something to `train()` and `inference()`.

clessig (approved):
Please remove the print statements, see my comments there. Otherwise fine to be merged.

src/weathergen/run_train.py
Jubeku: For some reason, in the notebook, the logs are printed during inference but not during training. This means in the notebook people cannot access the run id for fine-tuning. Something must be slightly different in the loggers of these two. Do you have an idea what?
I thought I could add the print just as for the trainable parameters which are only printed in interactive mode - but ruff is indeed complaining.

clessig: I think this issue from last week contains the solution; https://github.com/ecmwf/WeatherGenerator/issues/749.

Can you try to init the loggers with `ipykernel.iostream.OutStream` and see if the messages re-appear?

Jubeku: Didn't get that work yet. 
Tried `init_loggers(logging_level=logging.DEBUG, debug_output_streams=ipykernel.iostream.OutStream)`

clessig: @tjhunter : any ideas?

PR #815, opened 2025-08-25
add tokenizer base class

### LINKED ISSUE #207: Base class for tokenizers

Currently, the two tokenizers have identical code, e.g. in the ctor.

Create base class for tokenizer that implements shared code.

_No response_

### Additional context

_No response_

### Organisation

_No response_

TillHae:
## Description

Adds a base class for the tokenizers that implements shared code from tokenizer_forecast.py and tokenizer_masking.py

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #207 

## Code Compatibility

### Dependencies

### Documentation


shmh40 (changes requested):
This looks good, thanks @TillHae.

There is also another shared bit of code between tokenizer_masking and _forecast, where the source_centroids are computed:

      
     source_tokens_lens = torch.tensor([len(s) for s in source_tokens_cells], dtype=torch.int32)

        if source_tokens_lens.sum() > 0:
            source_means = [
                (
                    self.hpy_verts[-1][i].unsqueeze(0).repeat(len(s), 1)
                    if len(s) > 0
                    else torch.tensor([])
                )
                for i, s in enumerate(source_tokens_cells)
            ]
            source_means_lens = [len(s) for s in source_means]
            # merge and split to vectorize computations
            source_means = torch.cat(source_means)
            source_centroids = torch.cat(
                [source_means.to(torch.float32), r3tos2(source_means).to(torch.float32)], -1
            )
            source_centroids = torch.split(source_centroids, source_means_lens)

Could you include this in the base class? Just needs a compute_source_centroids function that takes self, source_tokens_cells as inputs and returns the source_centroids as seen in both tokenizer_masking and _forecast.

Also small thing but can you move the self.size_time_embedding to above the first verts00, it is a bit hidden down below given it is one line.

TillHae:
Thanks for the review @shmh40.

I tried to implement all your suggestions, but unfortunately source_tokens_lens is needed for the else statement in tokenizer_masking, which does not exist in tokenizer_forecast. Therefore I had to keep it in the subclasses.

I also removed the abstraction of the base_class. It only adds complexity.

shmh40 (approved):
Looks good, thank you @TillHae ! Tested with masking and forecast for multiple streams. Could have an init for TokenizerForecast but I think what you have is probably best practice to keep it clean.

Issue #816, opened 2025-08-26
Using list of forecasting steps causes crash


kacpnowak:
### What happened?

When using sequential forecast training mode there's a crash, making it impossible to use this mode. 

### What are the steps to reproduce the bug?

Use config with: 

forecast_steps: [2, 3, 4, 5, 6, 7, 8, 9, 10]
forecast_policy: "sequential"

### Version

0.1

### Platform (OS and architecture)

Levante

### Relevant log output

```shell
1: Traceback (most recent call last):
1:   File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/run_train.py", line 170, in train_with_args
1:     trainer.run(cf)
1:   File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/train/trainer.py", line 136, in run
1:     self.init(cf)
1:   File "/work/ab0995/a270225/WeatherGenerator2/src/weathergen/train/trainer.py", line 61, in init
1:     assert cf.forecast_policy if cf.forecast_steps > 0 else True
1:                                  ^^^^^^^^^^^^^^^^^^^^^
1: TypeError: '>' not supported between instances of 'ListConfig' and 'int'
```

### Organisation

AWI

kacpnowak:
assign me

SavvasMel:
@MatKbauer has observed the same isse

PR #824, opened 2025-08-27
[Hotfix] Fix crash when using list of forecasting steps

kacpnowak:
## Description

Fixes issue when using list of forecasting steps 

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/816

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/train/trainer.py
+++ b/src/weathergen/train/trainer.py
@@ -16,6 +16,8 @@ from typing import Any
 import numpy as np
 import torch
 import tqdm
+import weathergen.utils.config as config
+from omegaconf import ListConfig
 from torch import Tensor
 from torch.distributed.fsdp import FullOptimStateDictConfig, FullStateDictConfig, StateDictType
 from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
@@ -26,8 +28,6 @@ from torch.distributed.fsdp.fully_sharded_data_parallel import (
 from torch.distributed.fsdp.wrap import (
     size_based_auto_wrap_policy,  # default_auto_wrap_policy,
 )
-
-import weathergen.utils.config as config
 from weathergen.datasets.multi_stream_data_sampler import MultiStreamDataSampler
 from weathergen.model.model import Model, ModelParams
 from weathergen.model.utils import freeze_weights
@@ -58,7 +58,13 @@ class Trainer(TrainerBase):
 
         assert cf.samples_per_epoch % cf.batch_size_per_gpu == 0
         assert cf.samples_per_validation % cf.batch_size_validation_per_gpu == 0
-        assert cf.forecast_policy if cf.forecast_steps > 0 else True
+
+        if isinstance(cf.forecast_steps, int):
+            assert cf.forecast_policy if cf.forecast_steps > 0 else True
+        elif isinstance(cf.forecast_steps, ListConfig):
+            assert cf.forecast_policy if len(cf.forecast_steps) > 0 else True
+        else:
+            raise TypeError("Forecast steps must either an integer or a non-empty list of integers")
 
         self.mixed_precision_dtype = get_dtype(cf.attention_dtype)
 


src/weathergen/train/trainer.py
```diff
@@ -58,7 +59,13 @@ def init(
...
         assert cf.samples_per_epoch % cf.batch_size_per_gpu == 0
         assert cf.samples_per_validation % cf.batch_size_validation_per_gpu == 0
-        assert cf.forecast_policy if cf.forecast_steps > 0 else True
+
+        if isinstance(cf.forecast_steps, int):
+            assert cf.forecast_policy if cf.forecast_steps > 0 else True
+        elif isinstance(cf.forecast_steps, ListConfig):
+            assert cf.forecast_policy if len(cf.forecast_steps) > 0 else True
+        else:
+            raise TypeError("Forecast steps must either an integer or a non-empty list of integers")
```
TillHae: raise TypeError("Forecast steps must be an integer or a non-empty list of integers")

kacpnowak: Thanks! Fixed that

clessig: @TillHae : did you test? Then I am happy to approve

TillHae: @kacpnowak @clessig: I just tested it with an empty list as forecast_steps and got the following error:

```
  File "/p/project1/hclimrep/hauer1/WeatherGenerator_kacpnowak/src/weathergen/run_train.py", line 184, in train_with_args
    trainer.run(cf)
  File "/p/project1/hclimrep/hauer1/WeatherGenerator_kacpnowak/src/weathergen/train/trainer.py", line 145, in run
    self.dataset = MultiStreamDataSampler(
                   ^^^^^^^^^^^^^^^^^^^^^^^
  File "/p/project1/hclimrep/hauer1/WeatherGenerator_kacpnowak/src/weathergen/datasets/multi_stream_data_sampler.py", line 139, in __init__
    fsm = self.forecast_steps[0]
          ~~~~~~~~~~~~~~~~~~~^^^
IndexError: index 0 is out of bounds for axis 0 with size 0
```

But I think this is a different issue.

src/weathergen/train/trainer.py
```diff
@@ -58,7 +59,13 @@ def init(
 
         assert cf.samples_per_epoch % cf.batch_size_per_gpu == 0
         assert cf.samples_per_validation % cf.batch_size_validation_per_gpu == 0
-        assert cf.forecast_policy if cf.forecast_steps > 0 else True
+
+        if isinstance(cf.forecast_steps, int):
+            assert cf.forecast_policy if cf.forecast_steps > 0 else True
+        elif isinstance(cf.forecast_steps, ListConfig):
+            assert cf.forecast_policy if len(cf.forecast_steps) > 0 else True
```
TillHae: use ```sum(cf.forecast_steps) > 0``` instead of ```len(cf.forecast_steps) > 0``` for also skipping the ```forecast_policy``` check if all forecast_steps are 0

src/weathergen/train/trainer.py
TillHae: > Thanks! Fixed that

@kacpnowak The "be" is still missing in the error message.
Replace
```raise TypeError("Forecast steps must an integer or a non-empty list of integers")```
with
```raise TypeError("Forecast steps must be an integer or a non-empty list of integers")```

File: src/weathergen/train/trainer.py (hunk 2 above)
Comment (kacpnowak) [reply src/weathergen/train/trainer.py]: I'm not sure if that's a good solution. In principle list of [-1, 1] would bypass this check. I think we can make this more robust by adding `and all(step > 0 for step in cf.forecast_steps)`

File: src/weathergen/train/trainer.py (hunk 1 above)
Comment (kacpnowak) [reply src/weathergen/train/trainer.py]: Ok, fixed that too :-)

Comment (kacpnowak) [reply src/weathergen/train/trainer.py]: > @kacpnowak @clessig: I just tested it with an empty list as forecast_steps and got the following error:
> 
> ```
> Traceback (most recent call last):
>   File "/p/project1/hclimrep/hauer1/WeatherGenerator_kacpnowak/src/weathergen/run_train.py", line 184, in train_with_args
>     trainer.run(cf)
>   File "/p/project1/hclimrep/hauer1/WeatherGenerator_kacpnowak/src/weathergen/train/trainer.py", line 145, in run
>     self.dataset = MultiStreamDataSampler(
>                    ^^^^^^^^^^^^^^^^^^^^^^^
>   File "/p/project1/hclimrep/hauer1/WeatherGenerator_kacpnowak/src/weathergen/datasets/multi_stream_data_sampler.py", line 139, in __init__
>     fsm = self.forecast_steps[0]
>           ~~~~~~~~~~~~~~~~~~~^^^
> IndexError: index 0 is out of bounds for axis 0 with size 0
> ```
> 
> But I think this is a different issue.

It's a different issue, but we can address it in this PR if you see this fit @clessig

File: src/weathergen/train/trainer.py (hunk 2 above)
Comment (TillHae) [reply src/weathergen/train/trainer.py]: Yes, this is a better idea. But I think we have to do it like this:
```assert cf.forecast_policy if any(n > 0 for n in cf.forecast_steps) else True```

Then we can also remove the len() check as any returns False on empty lists, which then executes the else statement. It's also more efficient this way as any finishes on the first occurrence and does not have to go through the whole list. Nevertheless this is neglectable at the expected list sizes.

Comment (kacpnowak) [reply src/weathergen/train/trainer.py]: I think it's also a good idea, but it doesn't catch situation when there's negative values. How about we combine both into: `assert cf.forecast_policy and all(step > 0 for step in cf.forecast_steps) if any(n > 0 for n in cf.forecast_steps) else True`

File: src/weathergen/train/trainer.py [hunk 3]
```diff
@@ -58,7 +59,17 @@ def init(
 
         assert cf.samples_per_epoch % cf.batch_size_per_gpu == 0
         assert cf.samples_per_validation % cf.batch_size_validation_per_gpu == 0
-        assert cf.forecast_policy if cf.forecast_steps > 0 else True
+
+        if isinstance(cf.forecast_steps, int):
+            assert cf.forecast_policy if cf.forecast_steps > 0 else True
+        elif isinstance(cf.forecast_steps, ListConfig):
+            assert (
+                cf.forecast_policy and all(step > 0 for step in cf.forecast_steps)
```
TillHae: replace ```all(step > 0 for step in cf.forecast_steps)```with ```all(step >= 0 for step in cf.forecast_steps)``` for also making lists like [0] or [3, 5, 0] for forecast_steps possible.

src/weathergen/train/trainer.py
```diff
@@ -58,7 +59,17 @@ def init(
...
+                if len(cf.forecast_steps) > 0
```
TillHae: replace ```len(cf.forecast_steps) > 0``` with ```any(n > 0 for n in cf.forecast_steps)``` to stop checking forecast_policy if forecast_steps is [0] and therefore forecast_policy is not necessary.

kacpnowak: So how about this: `assert cf.forecast_policy and all(step >= 0 for step in cf.forecast_steps) if any(n > 0 for n in cf.forecast_steps) else True`? 

It should catch negative values and doesn't check forecast_policy if forecast_steps is [0]

TillHae: Yes, perfect

kacpnowak: @TillHae done, you can test and approve

src/weathergen/train/trainer.py
TillHae: @kacpnowak you missed this part

kacpnowak: Good catch

TillHae:
@clessig: From my side everything looks good and can be merged. I've tested all changes and I'll address the other issue we've detected in a separate PR.

TillHae (approved):

TillHae:
@kacpnowak can you quickly add the highlighted part so we also raise the error when forecast_steps is an empty list:

<pre><code>
if isinstance(cf.forecast_steps, int):
            assert cf.forecast_policy if cf.forecast_steps > 0 else True
        elif isinstance(cf.forecast_steps, ListConfig) <mark>and len(cf.forecast_steps) > 0:</mark>
            assert (
                cf.forecast_policy and all(step >= 0 for step in cf.forecast_steps)
                if any(n > 0 for n in cf.forecast_steps)
                else True
            )
        else:
            raise TypeError("Forecast steps must be an integer or a non-empty list of integers")
</code></pre>

kacpnowak:
@TillHae done

MatKbauer:
Looks great, I have tested and added another check to prevent negative `forecast_steps` integers. Also, I modularized the checks in a separate function for readability and added some assertion messages.

@kacpnowak, can you please verify the PR I have created against the branch on your fork: https://github.com/kacpnowak/WeatherGenerator2/pull/13

Once approved, we can update here and merge into develop.

src/weathergen/train/trainer.py
```diff
@@ -79,6 +80,56 @@ def init(
         self.init_perf_monitoring()
         self.train_logger = TrainLogger(cf, config.get_path_run(self.cf))
 
+    @staticmethod
```
clessig: Thanks for implementing this. But it should go somewhere else, likely config.py. 

@grassesi , @tjhunter : any opinion where such consistency checks for the config should go? I would put it config.py until it becomes too much.

clessig (commented):
Thanks for implementing the check. 

But is forecast_steps being a list still supported?

clessig:
> > Thanks for implementing the check.
> > But is forecast_steps being a list still supported?
> 
> It works when this issue is fixed, so I guess it is 😄 With all seriousness, it's a very useful feature in my opinion and supporting it is worth the effort

No, I was just double-checking that we retain the support for lists, which we need. 

We can merge once the function has moved.

MatKbauer (approved):
Looks all good. Let's merge.

Issue #845, opened 2025-09-02
Remove default --analysis_streams_output in cli.py


jpolz:
The default --analysis_streams_output is ERA5 which is overwriting the model config. 
Why is this an issue?
1. It requires setting analysis_streams_output in inference if not using ERA5 even though it was set in the config.yml during training
2. It leads to empty outputs in other streams without error if ERA5 was in the training streams. Not getting the outputs is confusing because the overwrite in cli.py is not obvious.

Change is one line. Happy to get your opinion on this

clessig:
@jpolz : thanks for opening this. The default should be to have all streams in a model/config as output. I think we should also think of a better name than `analysis_streams_output`. Maybe just `streams_output` ?

jpolz:
I opened a PR, for now only removing the default. I will test if the behavior is as expected (default from training/model config is used). I will also test if not setting this during training has any effect which later leads to desired outputs not being written. 

Renaming can be added. `streams_output` sounds good.

PR #850, opened 2025-09-03
Remove default for analysis_streams_output argument

jpolz:
## Description

Remove default argument completely as it is set in default_config.yml

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/845

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/utils/cli.py
+++ b/src/weathergen/utils/cli.py
@@ -61,7 +61,6 @@ def get_inference_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "--analysis_streams_output",
         nargs="+",
-        default=["ERA5"],
         help="Analysis output streams during inference.",
     )
 


clessig:
@jpolz : many thanks. I just tested and I needed this patch for the code to run:

```
diff --git a/src/weathergen/utils/validation_io.py b/src/weathergen/utils/validation_io.py
index 5f9dd07..b119202 100644
--- a/src/weathergen/utils/validation_io.py
+++ b/src/weathergen/utils/validation_io.py
@@ -30,9 +30,10 @@ def write_output(
     sample_idxs,
 ):
+    analysis_streams_output = cf.get( 'analysis_streams_output', None)
     if cf.streams_output is not None:
         output_stream_names = cf.streams_output
-    elif cf.analysis_streams_output is not None:  # --- to be removed at some point ---
+    elif analysis_streams_output is not None:  # --- to be removed at some point ---
         output_stream_names = cf.analysis_streams_output  # --- to be removed at some point ---
     else:
         output_stream_names = None
 ```
 
`analysis_streams_output` is not (necessarily) in the config. Otherwise it worked for me. So can you please apply it and then ping me so that we can merge.

File: src/weathergen/utils/validation_io.py [hunk 1]
```diff
@@ -30,7 +30,13 @@ def write_output(
     sample_idxs,
 ):
     stream_names = [stream.name for stream in cf.streams]
-    output_stream_names = cf.analysis_streams_output
+    if cf.streams_output is not None:
+        output_stream_names = cf.streams_output
+    elif cf.analysis_streams_output is not None:  # --- to be removed at some point ---
```
clessig: Does not exist any longer in config.

jpolz: Does not exist because this PR removes it or does not exist for other reasons? This line is to ensure backward compatibility when older model configs have this argument

jpolz: ah now I got it

jpolz: will apply the fix

PR #860, opened 2025-09-04
Clessig/develop/synop datareader

### LINKED ISSUE #862: Data reader for simple station/synop data in netCDF format

Various datasets can conveniently be stored in a simple netCDF format with an essentially tabular format and regular time steps (i.e. if a value/station is missing for a time step it is NaN). We need a data reader for this format

Data reader for the above format

_No response_

### Additional context

_No response_

### Organisation

_No response_

clessig:
## Description

Data reader for station-like data in simple netCDF format

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes #862 

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/datasets/data_reader_synop.py
+++ b/src/weathergen/datasets/data_reader_synop.py
@@ -50,6 +50,8 @@ class DataReaderSynop(DataReaderTimestep):
         None
         """
 
+        np32 = np.float32
+
         # open  dataset to peak that it is compatible with requested parameters
         ds = xr.open_dataset(filename, engine="netcdf4")
 
@@ -58,12 +60,10 @@ class DataReaderSynop(DataReaderTimestep):
             name = stream_info["name"]
             _logger.warning(f"{name} is not supported over data loader window. Stream is skipped.")
             super().__init__(tw_handler, stream_info)
-            self.init_empty()
+            self._init_empty()
             return
 
-        kwargs = {}
         if "frequency" in stream_info:
-            # kwargs["frequency"] = str_to_timedelta(stream_info["frequency"])
             assert False, "Frequency sub-sampling currently not supported"
 
         period = (ds.time[1] - ds.time[0]).values
@@ -82,7 +82,7 @@ class DataReaderSynop(DataReaderTimestep):
         )
         # If there is no overlap with the time range, no need to keep the dataset.
         if tw_handler.t_start >= data_end_time or tw_handler.t_end <= data_start_time:
-            self.init_empty()
+            self._init_empty()
             return
         else:
             self.ds = ds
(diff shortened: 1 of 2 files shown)


src/weathergen/datasets/data_reader_synop.py
```diff
@@ -0,0 +1,274 @@
...
+        # If there is no overlap with the time range, no need to keep the dataset.
+        if tw_handler.t_start >= data_end_time or tw_handler.t_end <= data_start_time:
+            self._init_empty()
+            return
+        else:
+            self.ds = ds
+            self.len = len(ds)
+
+        self.offset_data_channels = 4
+        self.fillvalue = ds["air_temperature"][0, 0].values.item()
```
tnipen, line 92: I think we should force the user to set a _FillValue attribute in the NetCDF files if there are missing data. I forgot to do this with my file (well actually, if you are using the default fill value of 9.96921e+36, then you don't need to set the flag, however xarray doesn't recognize this).

When _FillValue is set, xr.open_dataset fill automatically convert missing values to NaNs.

src/weathergen/datasets/data_reader_synop.py
```diff
@@ -0,0 +1,274 @@
...
+        )
+        # If there is no overlap with the time range, no need to keep the dataset.
+        if tw_handler.t_start >= data_end_time or tw_handler.t_end <= data_start_time:
+            self._init_empty()
+            return
+        else:
+            self.ds = ds
+            self.len = len(ds)
+
+        self.offset_data_channels = 4
```
tnipen, line 91: Line 91 can be removed now that the user specifically select channels

src/weathergen/datasets/data_reader_synop.py
```diff
@@ -0,0 +1,274 @@
...
+    TimeWindowHandler,
+    TIndex,
+    check_reader_data,
+)
+
+_logger = logging.getLogger(__name__)
+
+
+class DataReaderSynop(DataReaderTimestep):
+    "Wrapper for SYNOP datasets from MetNo in NetCDF"
```
tnipen, line 30: Would it be useful to specify the requirements of the NetCDF file? Here is a suggestion:

```
Generic parser for station data in NetCDF format. The file must have 2 dimensions: time and location.
- Data variables must have dimensions in the following order (time, location). The names of the dimension can be anything.
- Geoinfo variables must have dimension (location,)
- Any variable with missing values must have the _FillValue attribute set
- A latitude and longitude variable with dimension (location,) must be provided. The units must be degrees and the variable name can be configured.
- A variable called time must be provided and have dimension (time,). Units must follow CF-conventions and the variable must have a units attribute.
```

cristianlussana:
@clessig One very small remark: In the latest develop, in the config file for a data stream, the geoinfo are called `geoinfo_channels` while here they are called `geoinfos`. That could be a reason for keeping this distinction but if not, then we may decide to call `geoinfo_channels` also in this case.

Issue #792, opened 2025-08-21
Optional logarithmic y-axis for score plots + grid lines


Jubeku:
### Is your feature request related to a problem? Please describe.

In some cases we would like to have the option to plot the scores on logarithmic scale. In the example below log-scale would help see the differences in the initial forecast steps. 
Also grid lines would be helpful. 

<img width="1251" height="653" alt="Image" src="https://github.com/user-attachments/assets/b8697b2d-a4f2-4be9-9168-ef21814d296e" />

### Describe the solution you'd like

Add option in `config_eval.yml` to plot scores on log-scale. 
Add grid lines by default. 

### Organisation

ECMWF

Jubeku:
cc @iluise

tjhunter:
I see the need, but in the case of this plot: would it actually lead to a different scientific conclusion?

PR #865, opened 2025-09-05
add log scale and refactor plot_summary

iluise:
## Description

- refactor plot summary by moving functions to `plot_utils.py`
- add log scale
- add grid lines
- 
## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/792

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/plot_utils.py
+++ b/packages/evaluate/src/weathergen/evaluate/plot_utils.py
@@ -1,0 +1,62 @@
+# (C) Copyright 2025 WeatherGenerator contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+import numpy as np
+from weathergen.evaluate.plotter import LinePlots, Plotter
+import logging
+_logger = logging.getLogger(__name__)
+
+def collect_streams(runs):
+    """Get all unique streams across runs, sorted."""
+    return sorted({s for run in runs.values() for s in run["streams"].keys()})
+
+
+def collect_channels(scores_dict, metric, region, runs):
+    """Get all unique channels available for given metric and region across runs."""
+    channels = set()
+    if metric not in scores_dict or region not in scores_dict[metric]:
+        return []
+    for stream, run_data in scores_dict[metric][region].items():
+        for run_id in runs:
+            if run_id not in run_data:
+                continue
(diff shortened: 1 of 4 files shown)


packages/evaluate/src/weathergen/evaluate/plotter.py
```diff
@@ -326,20 +326,10 @@ def create_maps_per_sample(
...
-            {
-                key: value
-                for key, value in map_kwargs.copy().items()
-                if key not in variables
-            }
-            if map_kwargs is not None
-            else {}
-        )
-
-        if not hasattr(map_kwargs_save, "marker_size"):
```
SavvasMel, line 341: I am not sure if we should remove this. The idea of this part is to copy all the elements of `map_kwargs` that are not variable specific to `map_kwargs_save` . Those elements might be for example the `marker_size `or `linewidths` etc. but not `vmax `or `vmin`. In this way the below makes sense:

```
scatter_plt = ax.scatter(
            data["lon"],
            data["lat"],
            c=data,
            cmap="coolwarm",
            s=marker_size,
            marker=marker,
            transform=ccrs.PlateCarree(),
            vmin=vmin,
            vmax=vmax,
            linewidths=0.0,  # only markers, avoids aliasing for very small markers
            **map_kwargs_save,
        )
```

Specifically, `map_kwargs_save` should include global options/arguments only related to `scatter` and that are not bound to channels/variables. This allows also not to have to explicitly state all the arguments in `scatter` such as `cmap` or `norm` or `marker` or `s`

By passing `**map_kwargs_save[varname]` afterwards you do not allow for non-variable global options.

iluise: Hi Savvas, thanks for the suggestion. I updated it and made it more clear. I tested it with this configuration and it works as expected for me:
```
global_plotting_options:
  ERA5:
    marker_size: 2
    scale_marker_size: 1
    marker: "o"
    alpha: 0.5
    2t: 
      vmin: 250
      vmax: 300
    10u:
      vmin: -40
      vmax: 40
```

I know you can't test it right now, but can we merge it so I can open a few other merge requests today and if there's something odd we fix it later? if we move to plotly this might change anyway also..

packages/evaluate/src/weathergen/evaluate/plot_utils.py
```diff
@@ -0,0 +1,66 @@
...
+# nor does it submit to any jurisdiction.
+
+import logging
+
+import numpy as np
+
+_logger = logging.getLogger(__name__)
+
+
+def collect_streams(runs):
```
tjhunter: you should put types on your functions, it can really help in finding basic bugs when refactoring.

tjhunter (approved):

PR #868, opened 2025-09-05
Sophiex/dev/fsdp2

### LINKED ISSUE #863: Logger should automatically create an output file

The logger that prints to stdout should also save everything in the same way to an output file, so that in interactive jobs you can easily look at the output again

_No response_

_No response_

### Additional context

_No response_

### Organisation

_No response_

### LINKED ISSUE #867: Upgrade FSDP to FSDP2

As per the title, code changes throughout model and trainer to change from FSDP to FSDP2

### Organisation

ECMWF

sophie-xhonneux:
## Description

Upgrade FSDP to FSDP2, due to its better representation internally of Tensors, allowing for more flexibility later on.

## Type of Change

-   [ ] Documentation update

Non of the above, really more of an upgrade, the goal is to be as non-breaking as possible

## Issue Number

Closes #867

Closes #863 

## Code Compatibility

### Dependencies

### Documentation


src/weathergen/train/trainer.py
Jubeku, line 1: For what is this needed in `save_model`? Can it be removed? (It also contains still `ddp_model`.)
```
 if self.cf.with_ddp and self.cf.with_fsdp:
            with FSDP.state_dict_type(
                self.ddp_model,
                StateDictType.FULL_STATE_DICT,
                FullStateDictConfig(rank0_only=False),
                FullOptimStateDictConfig(rank0_only=False),
            ):
                state = self.ddp_model.state_dict()
```

src/weathergen/train/trainer.py
```diff
@@ -625,25 +664,152 @@ def validate(self, epoch):
...
+            meta_sharded_sd = self.model.state_dict()
+            sharded_sd = {}
+            for param_name, full_tensor in params.items():
+                sharded_meta_param = meta_sharded_sd.get(param_name)
+                sharded_tensor = distribute_tensor(
+                    full_tensor,
+                    sharded_meta_param.device_mesh,
+                    sharded_meta_param.placements,
+                )
+                sharded_sd[param_name.replace("module.", "")] = nn.Parameter(sharded_tensor)
```
Jubeku: Add torch --> `torch.nn.Parameter(shared_tensor)`

src/weathergen/train/trainer.py
```diff
@@ -625,25 +664,152 @@ def validate(self, epoch):
...
+        else:
+            maybe_sharded_sd = {}
+            for k in params.keys():
+                maybe_sharded_sd[k.replace("module.", "")] = params[k]
+        # choose `assign=True` for sharded model since we cannot call `copy_` on meta tensor
+        mkeys, ukeys = self.model.load_state_dict(
+            maybe_sharded_sd, strict=False, assign=is_model_sharded
```
Jubeku: Still need to work on how to load the "maybe_sharded_sd", but if I change this line to `self.model.load_state_dict(sharded_sd, ...)` it works for me. Though I still need to double check if the weights are actually working. 
Torchrun and pdb don't seem to like each other?!

src/weathergen/train/trainer.py
```diff
@@ -226,25 +230,56 @@ def run(self, cf, run_id_contd=None, epoch_contd=None):
...
 
-        self.model_params = ModelParams().create(cf).to("cuda")
+        # load model if specified
+        if run_id_contd is None:
+            self.model.to_empty(device="cuda")
+            self.model.reset_parameters()
+        else:
+            _logger.info(f"Continuing run with id={run_id_contd} at epoch {epoch_contd}.")
+            self.load_model(run_id_contd, epoch_contd)
+            _logger.info(f"Loaded model id={run_id_contd}.")
```
Jubeku: change two appearances of `_logger.info` to `logger.info`.

src/weathergen/train/trainer.py
sophie-xhonneux: discussed this with Christian `nn.Parameter` is fine

src/weathergen/train/trainer.py
sophie-xhonneux: I think in general pdb in multiple processes is tricky :/

src/weathergen/train/trainer.py
Jubeku: But you need to add `import torch.nn as nn`

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -409,7 +410,7 @@ def worker_workset(self):
         worker_info = torch.utils.data.get_worker_info()
 
         if worker_info is None:
-            assert self.num_ranks == 1
+            assert self.world_size == 1
```
tjhunter: `assert self.world_size == 1, self.world_size` <- make the assertions actionable

src/weathergen/model/model.py
```diff
@@ -34,16 +36,79 @@
...
 from weathergen.utils.config import Config, get_dtype
-from weathergen.utils.logger import logger
 
+logger = logging.getLogger(__name__)
 
 class ModelParams(torch.nn.Module):
     """Creation of query and embedding parameters of the model."""
 
-    def __init__(self) -> None:
+    def __init__(self, cf) -> None:
```
tjhunter, line 48: `cf: Config`

src/weathergen/model/model.py
```diff
@@ -387,6 +453,15 @@ def create(self) -> "Model":
 
         return self
 
+    def reset_parameters(self):
```
tjhunter, line 461: note that it does not need to be a method, it could be a separate function

src/weathergen/run_train.py
```diff
@@ -59,13 +65,24 @@ def inference_from_args(argl: list[str]):
...
-    init_loggers(logging_level=logging.DEBUG, debug_output_streams=fname_debug_logging)
+    devices = Trainer.init_torch()
+    cf = Trainer.init_ddp(cf)
+
+    if cf.rank == 0:
+        # Get current time
+        now = datetime.now()
+        timestamp = now.strftime("%Y-%m-%d-%H%M")
+
+        output_dir = f"./output/{timestamp}-{cf.run_id}"
```
tjhunter: Can we do the opposite? Typically we communicate around the run_id when having to do experiments (it is the primary search key) and this pattern would break autocomplete in the terminal. My preference in order would be:

```
output/{run_id}/log_{time}.txt
```

or

```
output/{run_id}/{time}/log.txt
```

src/weathergen/run_train.py
```diff
@@ -127,8 +142,19 @@ def train_continue_from_args(argl: list[str]):
...
```
tjhunter: see comment above. 

Also, since we are repeating the logic: can we have a tiny function that returns the output dir? We tend to adjust frequently these path-related functions. Also, use instead `src/weathergen/utils/config.py` : `_REPO_ROOT`

The current path makes the assumption that the process is launched from the `WeatherGenerator` directory. We have had issues with that when importing and moving CLI code.

src/weathergen/run_train.py
```diff
@@ -169,8 +195,19 @@ def train_with_args(argl: list[str], stream_dir: str | None):
...
```
tjhunter: same

src/weathergen/train/trainer.py
```diff
@@ -49,10 +58,7 @@ def __init__(self, checkpoint_freq=250, print_freq=10):
         self.checkpoint_freq = checkpoint_freq
         self.print_freq = print_freq
 
-    def init(
-        self,
-        cf: Config,
-    ):
+    def init(self, cf: Config, devices):
```
tjhunter, line 63: type?

src/weathergen/train/trainer.py
```diff
@@ -186,18 +189,11 @@ def run(self, cf, run_id_contd=None, epoch_contd=None):
...
-            self.model.load(run_id_contd, epoch_contd)
-            if cf.with_ddp and cf.with_fsdp:
-                FSDP.set_state_dict_type(
-                    self.model,
-                    StateDictType.FULL_STATE_DICT,
-                    FullStateDictConfig(rank0_only=False),
-                )
-            logger.info(f"Loaded model id={run_id_contd}.")
+        device_type = torch.accelerator.current_accelerator()
+        device = torch.device(f"{device_type}:{cf.rank}")
```
tjhunter: probably not relevant yet, but I will be curious to see if it works on CPU too.

src/weathergen/train/trainer.py
```diff
@@ -625,25 +641,152 @@ def validate(self, epoch):
...
+            maybe_sharded_sd, strict=False, assign=is_model_sharded
+        )
+
+        if len(mkeys) > 0:
+            logger.warning(f"Missing keys when loading model: {mkeys}")
+
+        if len(ukeys) > 0:
+            logger.warning(f"Unused keys when loading model: {mkeys}")
+
+    def load_optim(self):
```
tjhunter, line 729: I would leave it to a follow up work.

src/weathergen/train/trainer_base.py
```diff
@@ -68,90 +66,62 @@ def init_torch(use_cuda=True, num_accs_per_task=1, multiprocessing_method="fork"
...
-            cf.with_ddp = False
-            cf.rank = rank
-            cf.num_ranks = num_ranks
-            _logger.info(
-                "DDP not initialized. MASTER_ADDR not set. Running in single process mode."
-            )
-            _logger.info(f"rank: {rank} has run_id: {cf.run_id}")
+
+        if not dist.is_available():
+            print("Distributed training is not available.")
```
tjhunter, line 74: logger.warning?

tjhunter (changes requested):
I did not try it, my main comments are about the log files path. Happy to discuss if there is a workflow that I am missing here.

src/weathergen/model/attention.py
```diff
@@ -155,6 +155,7 @@ def sparsity_mask(score, b, h, q_idx, kv_idx):
             return flex_attention(qs, ks, vs, score_mod=sparsity_mask)
 
         self.compiled_flex_attention = torch.compile(att, dynamic=False)
+        # self.compiled_flex_attention = att
```
clessig: Remove

src/weathergen/model/attention.py
```diff
@@ -238,6 +239,7 @@ def mask_block_local(batch, head, idx_q, idx_kv):
         )
         # compile for efficiency
         self.flex_attention = torch.compile(flex_attention, dynamic=False)
+        # self.flex_attention = flex_attention
```
clessig: Remove

src/weathergen/model/embeddings.py
```diff
@@ -152,7 +155,8 @@ def forward_channels(self, x_in, centroids):
         x = peh(checkpoint(self.embed, x_in.transpose(-2, -1), use_reentrant=False))
 
         for layer in self.layers:
-            x = checkpoint(layer, x, use_reentrant=False)
+            # x = checkpoint(layer, x, use_reentrant=False)
```
clessig: Is this required for FSDP2? Otherwise let's leave it. I think we also need the checkpoint here for larger configs.

src/weathergen/model/embeddings.py
```diff
@@ -124,7 +124,10 @@ def __init__(
             assert self.unembed_mode == "block"  # only supported mode at the moment
             # padding needed if the unembedded columns cannot be concatenated to dim_out (e.g GPSRO)
             self.pad = self.dim_out % token_size
-            self.out_pad = torch.nn.Parameter(torch.zeros(self.pad))
+            self.out_pad = torch.nn.Parameter(
```
clessig, line 127: It shouldn't get gradients, but let's do it in a separate PR

src/weathergen/model/engines.py
```diff
@@ -384,7 +384,7 @@ def __init__(
                 self.pred_heads[-1].append(fal)
 
     #########################################
-    @torch.amp.custom_fwd(cast_inputs=torch.float32, device_type="cuda")
+    # @torch.amp.custom_fwd(cast_inputs=torch.float32, device_type="cuda")
```
clessig: Let's leave it. We always needed this for stability. I would be uncomfortable removing it.

src/weathergen/model/model.py
```diff
@@ -34,16 +36,79 @@
...
+
+        self.healpix_level = cf.healpix_level
+        self.num_healpix_cells = 12 * 4**cf.healpix_level
+        self.dtype = get_dtype(cf.attention_dtype)
+
+        bs = cf.batch_size_per_gpu
+        nqs = 9
+        s = [bs, self.num_healpix_cells, cf.ae_local_num_queries, cf.ae_global_dim_embed]
+
+        ### POSITIONAL EMBEDDINGS ###
```
clessig, line 61: `# positional embeddings` (it's not Fortran, we don't need all caps ;))

src/weathergen/model/model.py
```diff
@@ -34,16 +36,79 @@
...
+        pe = torch.zeros(
+            self.num_healpix_cells,
+            cf.ae_local_num_queries,
+            cf.ae_global_dim_embed,
+            dtype=self.dtype,
+        )
+        self.pe_global = torch.nn.Parameter(pe, requires_grad=False)
+        # self.pe_global = pe
+
+        ### HEALPIX NEIGHBOURS ###
```
clessig, line 77: `# healpix neighbors`

src/weathergen/model/model.py
```diff
@@ -787,19 +863,19 @@ def predict(
...
-                            tc_embed,
-                            streams_data[i_b][ii].target_coords[fstep],
-                            use_reentrant=False,
-                        )
-                        if len(streams_data[i_b][ii].target_coords[fstep].shape) > 1
-                        else streams_data[i_b][ii].target_coords[fstep]
-                        for i_b in range(len(streams_data))
-                    ]
-                )
+            # with torch.amp.autocast("cuda", dtype=torch.float32, enabled=False):
```
clessig: Is this not compatible with FSDP2?

src/weathergen/run_train.py
clessig: Yes, we should start with the run_id

src/weathergen/run_train.py
```diff
@@ -59,13 +65,24 @@ def inference_from_args(argl: list[str]):
     )
     cf = config.set_run_id(cf, args.run_id, args.reuse_run_id)
 
-    fname_debug_logging = f"./logs/debug_log_{cf.run_id}.txt"
-    init_loggers(logging_level=logging.DEBUG, debug_output_streams=fname_debug_logging)
+    devices = Trainer.init_torch()
```
clessig, line 64: I really don't like the logic here. `init_torch` and `init_ddp` should stay in the Trainer.

src/weathergen/run_train.py
```diff
@@ -59,13 +65,24 @@ def inference_from_args(argl: list[str]):
...
+    cf = Trainer.init_ddp(cf)
+
+    if cf.rank == 0:
```
clessig: Why only for rank 0? We need proper logging for on all workers.

src/weathergen/run_train.py
clessig: This should not happen raw here but in `init_loggers()` or some utility function. It's also duplicated in inference(), train_continue(), train().

src/weathergen/train/trainer_base.py
```diff
@@ -68,90 +66,62 @@ def init_torch(use_cuda=True, num_accs_per_task=1, multiprocessing_method="fork"
...
+            if torch.accelerator.is_available():
+                device_type = torch.accelerator.current_accelerator()
+                device = torch.device(f"{device_type}:{rank}")
+                torch.accelerator.set_device_index(rank)
+                print(f"DDP initialization: rank={rank}, world_size={world_size}")
+            else:
+                device = torch.device("cpu")
+                print(f"Running on device {device}")
+
+            backend = torch.distributed.get_default_backend_for_device(device)
```
clessig, line 106: We should insist on NCCL.

src/weathergen/train/trainer_base.py
```diff
@@ -68,90 +66,62 @@ def init_torch(use_cuda=True, num_accs_per_task=1, multiprocessing_method="fork"
...
+            torch.distributed.init_process_group(
+                backend=backend,
+                world_size=world_size,
+                device_id=device,
+                rank=rank,
+                init_method=f"tcp://{master_addr}:{master_port}",
+            )
+            print(f"Process group initialized ({backend}).")
+
+            if is_root():
```
clessig, line 116: This statement should come after the barrier

src/weathergen/utils/logger.py
```diff
@@ -43,48 +43,8 @@ def format(self, record, *args, **kwargs):
...
 @cache
-def init_loggers(
-    logging_level=logging.DEBUG,
-    critical_output_streams=sys.stderr,
-    error_output_streams=sys.stderr,
-    warning_output_streams=sys.stderr,
-    info_output_streams=sys.stdout,
-    debug_output_streams=sys.stdout,
-):
+def init_loggers(filename, logging_level_file=logging.DEBUG, logging_level_console=logging.DEBUG):
```
clessig: We need separate files for the different logging levels. We had it before that people overlooked error messages because it was 1 line in hundreds of debug messages. Having a clean separation there is the reason for having the different logging levels.

src/weathergen/utils/logger.py
```diff
@@ -103,32 +63,30 @@ def init_loggers(
...
+    # make sure the path is independent of path where job is launched
+    if not ofile.is_absolute():
+        work_dir = pathlib.Path(_load_private_conf().get("path_shared_working_dir"))
+        ofile = work_dir / ofile
+    # make sure the parent directory exists
+    pathlib.Path(ofile.parent).mkdir(parents=True, exist_ok=True)
+    logging.basicConfig(level=logging_level_file,
+                    format=format_str,
+                    datefmt='%m-%d %H:%M',
+                    filename=ofile,
```
clessig: This didn't go through ruff, did it?

src/weathergen/model/engines.py
sophie-xhonneux: You have to do this differently in FSDP2, trying to figure out the exact dynamics

src/weathergen/model/model.py
sophie-xhonneux: I like it in comments, but okay

src/weathergen/model/model.py
sophie-xhonneux: it does in Pytorch and it is the dominant custom. In pytorch each nn.Module that PyTorch provides has a `self.reset_parameters()`, we should mimic that

src/weathergen/model/model.py
sophie-xhonneux: Don't think it is impossible, but haven'y yet fully solved the casting sharding flow chart in my head, better to do it via the MixedPrecisionPolicy

src/weathergen/run_train.py
sophie-xhonneux: they are? but they are static functions and I call them there

src/weathergen/run_train.py
sophie-xhonneux: this awful when going through the directories, because they are listed alphabetically, by having the time first, we get it chronologically. So I am strongly against this

src/weathergen/train/trainer_base.py
sophie-xhonneux: init_ddp comes before the loggers are initialised

src/weathergen/train/trainer_base.py
sophie-xhonneux: maybe a warning?

src/weathergen/train/trainer_base.py
sophie-xhonneux: why?

src/weathergen/utils/logger.py
sophie-xhonneux: shouldn't we crash on errors, rather than continue silently?

src/weathergen/utils/logger.py
sophie-xhonneux: not yet

src/weathergen/run_train.py
sophie-xhonneux: This creates the file and sets up the loggers, this should happen only once, else each worker creates a logger with a different run id file

src/weathergen/run_train.py
sophie-xhonneux: the solution to do this search is fuzzy search/autocomplete

src/weathergen/utils/logger.py
tjhunter: some exceptions get logged at warning levels before a fatal error happens further down. I am agnostic personally because just the error without context is usually enough for a quick introspection but not enough to diagnose the issue (you need the info/debug logs before). Having a complete log separately is also a must.

src/weathergen/utils/logger.py
```diff
@@ -103,32 +116,33 @@ def init_loggers(
...
+                filename = f"{output_dir}/{run_id}-{v}"
+                ofile = pathlib.Path(filename)
+                # make sure the path is independent of path where job is launched
+                if not ofile.is_absolute():
+                    work_dir = pathlib.Path(_load_private_conf().get("path_shared_working_dir"))
+                    ofile = work_dir / ofile
+                pathlib.Path(ofile.parent).mkdir(parents=True, exist_ok=True)
+                handler[k] = ofile
+            else:
+                continue
```
grassesi, line 144: This just injects the proper log output file into each handler that defines a `"filename"` field right? If so you can just write:
```python
for handler in logging_config["handlers"].values():
    if "filename" in handler:
        filename = f"{output_dir}/{run_id}-{v}"
        ofile = pathlib.Path(filename)
        # make sure the path is independent of path where job is launched
        if not ofile.is_absolute():
            work_dir = pathlib.Path(_load_private_conf().get("path_shared_working_dir"))
            ofile = work_dir / ofile
        pathlib.Path(ofile.parent).mkdir(parents=True, exist_ok=True)
        handler[k] = ofile
```

The path handling should be done in advance (ideally by the caller => see my comments in `run_train`). If this is done this simplifies even more:
```python
for handler in logging_config["handlers"].values():
    if "filename" in handler:
        log_file = output_dir /  v # assume output_dir is properly set up
        log_file.parent.mkdir(parents=True, exist_ok=True)

        handler["filename"] = log_file
```

src/weathergen/utils/logger.py
```diff
@@ -7,14 +7,67 @@
...
+import json
 import logging
+import logging.config
 import os
 import pathlib
-import sys
+from datetime import datetime
 from functools import cache
 
 from weathergen.utils.config import _load_private_conf
```
grassesi: This is an implementation detail of the config module, instead pass what you need from the private config into the functions.

src/weathergen/run_train.py
```diff
@@ -59,13 +61,17 @@ def inference_from_args(argl: list[str]):
     )
     cf = config.set_run_id(cf, args.run_id, args.reuse_run_id)
 
-    fname_debug_logging = f"./logs/debug_log_{cf.run_id}.txt"
-    init_loggers(logging_level=logging.DEBUG, debug_output_streams=fname_debug_logging)
+    devices = Trainer.init_torch()
+    cf = Trainer.init_ddp(cf)
+
+    init_loggers(cf.run_id)
+
```
grassesi, line 68: Pass a absolute version of the output (sub)directory here, that way you dont have to deal with it in the function:
```python
init_loggers(Path(cf.path_shared_working_dir) / "output" / cf.run_id)
```

src/weathergen/run_train.py
```diff
@@ -127,8 +131,10 @@ def train_continue_from_args(argl: list[str]):
     )
     cf = config.set_run_id(cf, args.run_id, args.reuse_run_id)
 
-    fname_debug_logging = f"./logs/debug_log_{cf.run_id}.txt"
-    init_loggers(logging_level=logging.DEBUG, debug_output_streams=fname_debug_logging)
+    devices = Trainer.init_torch()
+    cf = Trainer.init_ddp(cf)
+
+    init_loggers(cf.run_id)
```
grassesi, line 138: see above

src/weathergen/run_train.py
```diff
@@ -169,8 +175,15 @@ def train_with_args(argl: list[str], stream_dir: str | None):
...
-    fname_debug_logging = f"./logs/debug_log_{cf.run_id}.txt"
-    init_loggers(logging_level=logging.DEBUG, debug_output_streams=fname_debug_logging)
+    devices = Trainer.init_torch()
+    cf = Trainer.init_ddp(cf)
+
+    # if cf.rank == 0:
+    # this line should probably come after the processes have been sorted out else we get lots
+    # of duplication due to multiple process in the multiGPU case
+    init_loggers(cf.run_id)
+
```
grassesi, line 180: see above

src/weathergen/utils/logger.py
```diff
@@ -43,48 +96,8 @@ def format(self, record, *args, **kwargs):
...
-def init_loggers(
-    logging_level=logging.DEBUG,
-    critical_output_streams=sys.stderr,
-    error_output_streams=sys.stderr,
-    warning_output_streams=sys.stderr,
-    info_output_streams=sys.stdout,
-    debug_output_streams=sys.stdout,
-):
+def init_loggers(run_id, logging_config=None):
     """
```
grassesi, line 100: Pass a absolute version of the output (sub)path here:
```python
def init_loggers(output_dir: Path, logging_config=None):
...
```

(Maybe rename `output_dir` to `logging dir`? I think this is more meaningful)

src/weathergen/utils/logger.py
```diff
@@ -103,32 +116,33 @@ def init_loggers(
...
-
-
-# TODO: remove, it should be module-level loggers
-logger = logging.getLogger("weathergen")
+    # Get current time
+    now = datetime.now()
+    timestamp = now.strftime("%Y-%m-%d-%H%M")
+
+    output_dir = f"./output/{timestamp}-{run_id}"
+
```
grassesi: why not just have all the logs of a run in a dedicated subdirectory?

src/weathergen/run_train.py
tjhunter: I would keep it inside the function, I am not sure what we are gaining to create the path here.

grassesi (changes requested):
The logger setup should not depend on `config._load_private_conf`. This can be circumvented by just injecting a properly set up path when calling `init_loggers`. This will also significantly simplify the loop for setting the proper filename.

src/weathergen/run_train.py
grassesi: Its more of a conceptual thing for me: What the logger really needs is not `run_id` or `path_shared_working_dir`, but it just needs to know where to put the logs. That is why I would just give it that path. Obviously implementing it multiple times in `run_train` is also not the best scope to put that logic: `config` module should be really responsible for this and return the proper path via something like `get_path_logging(config: Config)` (similar to `get_path_run` / `get_path_model`) then this simplifies to `init_loggers(config.get_path_logger(cf))`. But I did not suggest this since I wanted to keep this PR small and this something that can be easily realized as a refactoring at a later point. :shrug:

src/weathergen/run_train.py
grassesi: > this awful when going through the directories, because they are listed alphabetically, by having the time first, we get it chronologically. So I am strongly against this

I would agree on trying to keep the semantics the same. If need be the files can be listed chronologically using `ls -rt` / `ls -t`

sophie-xhonneux:
> With `train_continue` I get:
> 
> ```
> 0: [rank0]: Traceback (most recent call last):
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/src/weathergen/run_train.py", line 208, in <module>
> 0: [rank0]:     train_continue()
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/src/weathergen/run_train.py", line 89, in train_continue
> 0: [rank0]:     train_continue_from_args(sys.argv[1:])
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/src/weathergen/run_train.py", line 149, in train_continue_from_args
> 0: [rank0]:     trainer.run(cf, devices, args.from_run_id, args.epoch)
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/src/weathergen/train/trainer.py", line 402, in run
> 0: [rank0]:     self.train(epoch)
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/src/weathergen/train/trainer.py", line 555, in train
> 0: [rank0]:     preds, posteriors = self.model(
> 0: [rank0]:                         ^^^^^^^^^^^
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1739, in _wrapped_call_impl
> 0: [rank0]:     return self._call_impl(*args, **kwargs)
> 0: [rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1845, in _call_impl
> 0: [rank0]:     return inner()
> 0: [rank0]:            ^^^^^^^
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1772, in inner
> 0: [rank0]:     args_kwargs_result = hook(self, args, kwargs)  # type: ignore[misc]
> 0: [rank0]:                          ^^^^^^^^^^^^^^^^^^^^^^^^
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/distributed/fsdp/_fully_shard/_fsdp_state.py", line 71, in fsdp_hook_wrapper
> 0: [rank0]:     return torch._dynamo.disable(func, recursive=True)(*args, **kwargs)
> 0: [rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/_dynamo/eval_frame.py", line 745, in _fn 
> 0: [rank0]:     return fn(*args, **kwargs)
> 0: [rank0]:            ^^^^^^^^^^^^^^^^^^^
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/distributed/fsdp/_fully_shard/_fsdp_state.py", line 231, in _pre_forward
> 0: [rank0]:     args, kwargs = self._root_pre_forward(module, args, kwargs)
> 0: [rank0]:                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/distributed/fsdp/_fully_shard/_fsdp_state.py", line 123, in _root_pre_forward
> 0: [rank0]:     self._lazy_init()
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/distributed/fsdp/_fully_shard/_fsdp_state.py", line 187, in _lazy_init
> 0: [rank0]:     state._fsdp_param_group.lazy_init()
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/distributed/fsdp/_fully_shard/_fsdp_param_group.py", line 239, in lazy_init
> 0: [rank0]:     self._validate_no_meta_params()
> 0: [rank0]:   File "/users/lessig/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/distributed/fsdp/_fully_shard/_fsdp_param_group.py", line 644, in _validate_no_meta_params
> 0: [rank0]:     raise RuntimeError(
> ```
> 
> This already contains the following patch:
> 
> ```
> diff --git a/src/weathergen/train/trainer.py b/src/weathergen/train/trainer.py
> index 6e1b846..bd71925 100644
> --- a/src/weathergen/train/trainer.py
> +++ b/src/weathergen/train/trainer.py
> @@ -707,12 +710,12 @@ class Trainer(TrainerBase):
>                  )
>                  sharded_sd[param_name.replace("module.", "")] = nn.Parameter(sharded_tensor)
>          else:
> -            maybe_sharded_sd = {}
> +            meta_sharded_sd = {}
>              for k in params.keys():
>                  maybe_sharded_sd[k.replace("module.", "")] = params[k]
>          # choose `assign=True` for sharded model since we cannot call `copy_` on meta tensor
>          mkeys, ukeys = self.model.load_state_dict(
> -            maybe_sharded_sd, strict=False, assign=is_model_sharded
> +            meta_sharded_sd, strict=False, assign=is_model_sharded
>          )
> ```

I haven't seen this issue before, what HPC and what config did you use?

src/weathergen/utils/logger.py
```diff
@@ -103,32 +116,33 @@ def init_loggers(
...
-        logger = init_logger_per_stream(logger, min(streams), stream_handle)
-
-
-# TODO: remove, it should be module-level loggers
-logger = logging.getLogger("weathergen")
+    # Get current time
+    now = datetime.now()
+    timestamp = now.strftime("%Y-%m-%d-%H%M")
+
+    output_dir = f"./output/{timestamp}-{run_id}"
```
tjhunter: my previous comment seems to have been lost. It would be good to have all the files from a run in organized together: `output/{run_id}/{timestamp}_{err|debug|...}.txt`
- the run id is the logical unit that we have to organize our runs. all the other elements (models, ...) are organized this way
- putting the timestamp last is not friendly to tab completion. Reflecting on the level of technical sophistication of all of us, I suppose very few people have fuzzy search enabled or even know it exists. We need to cater to the most common use cases.

clessig:
Resolved via a different PR

Issue #855, opened 2025-09-04
Implement forecast activity metric


MatKbauer:
### Is your feature request related to a problem? Please describe.

As of now, the current issue of freezing models (predicting no changes after some rollout steps) can hardly be quantified with our metrics.

### Describe the solution you'd like

To understand whether and at what lead time a model predicts smaller changes and eventually freezes, we can implement a `forecast activity` metric, which we define as the difference between predicted subsequent time steps, such as

`fact = \hat{y}_{t+1} - \hat{y}_{t}`

Ideally, `fact_{var} \in \mathbb{R}^T` is a one-dimensional time-series per variable to characterize the change in the respective time step. Reducing the spatial dimensions can be realized by calculating the norm of the difference, while maintaining the time dimension.

### Organisation

ECMWF

MatKbauer:
Good points, @jpolz, thanks for bringing them up. The `\hat{}` stands for prediction. I'd like to keep the effort minimal and would like to come up with a simple metric that quantifies the actifity of a forecast over time. Something like the folling:

<img width="895" height="519" alt="Image" src="https://github.com/user-attachments/assets/c929ac43-c41b-405e-b19f-205c7717c361" />

Given the name "forecast activity" stands in conflict with existing definitions, I suggest to name our metric froct for "forecast rate of change over time". Sounds a bit clumpsy, but should do for now.

jpolz:
Sounds good. But _var would still be variance?

PR #892, opened 2025-09-10
Implement forecast activity metrics

jpolz:
## Description

Implement forecast activity metrics

## Type of Change

-   [ ] Documentation update

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/855

## Code Compatibility

### Dependencies

### Documentation

## Additional Notes

CC @MatKbauer

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/score.py
+++ b/packages/evaluate/src/weathergen/evaluate/score.py
@@ -604,6 +604,72 @@ class Scores:
 
         return rmse
 
+    def calc_fact(
+        self,
+        p: xr.DataArray,
+        p1: xr.DataArray,
+        group_by_coord: str | None = None,
+    ):
+        """
+        Calculate simple forecast activity
+
+        Parameters
+        ----------
+        p: xr.DataArray
+            Forecast data array
+        p1: xr.DataArray
+            Next forecast step data array
+        group_by_coord: str
+            Name of the coordinate to group by.
+            If provided, the coordinate becomes a new dimension of the FACT score.
+        """
+        if self._agg_dims is None:
+            raise ValueError(
+                "Cannot calculate forecast activity without aggregation dimensions (agg_dims=None)."
+            )
+
+        fact = p1 - p
+
+        if group_by_coord:
+            fact = fact.groupby(group_by_coord)
+
+        fact = self._mean(fact)
+
+        return fact
+
+    def calc_fact2(
+        self,
+        p: xr.DataArray,
+        clim_mean: xr.DataArray,
+        group_by_coord: str | None = None,
+    ):
+        """
+        Calculate forecast activity after Ben Bouallègue et al. (2024)
+
+        Parameters
+        ----------
+        p: xr.DataArray
+            Forecast data array
+        clim_mean: xr.DataArray
+            Climatological mean data array
+        group_by_coord: str
+            Name of the coordinate to group by.
+            If provided, the coordinate becomes a new dimension of the FACT score.
+        """
+        if self._agg_dims is None:
+            raise ValueError(
+                "Cannot calculate forecast activity without aggregation dimensions (agg_dims=None)."
+            )
+
+        fact2 = p - clim_mean
+
+        if group_by_coord:
+            fact2 = fact2.groupby(group_by_coord)
+
+        fact = fact.std(dim=self._agg_dims)
+
+        return fact2
+
     def calc_acc(
         self,
         p: xr.DataArray,
--- a/packages/evaluate/src/weathergen/evaluate/utils.py
+++ b/packages/evaluate/src/weathergen/evaluate/utils.py
@@ -267,10 +267,25 @@ def calc_scores_per_stream(
         },
     )
 
+    # fsteps ordered numerically
+    sfsteps = sorted(fsteps)
+
     for (fstep, tars), (_, preds) in zip(
         da_tars.items(), da_preds.items(), strict=False
     ):
         _logger.debug(f"Verifying data for stream {stream}...")
+
+        # Prepare any metric-specific kwargs, similar for ACC
+        metrics_kwargs = {}
+        for metric in metrics:
+            metrics_kwargs[metric] = {}
+
+        if "fact" in metrics:
+            # index of fstep in sorted fsteps
+            fstep_idx = sfsteps.index(fstep)
+            next_fstep = sfsteps[fstep_idx + 1] if fstep_idx + 1 < len(sfsteps) else None
+            preds_next = da_preds.get(next_fstep, None)
+            metrics_kwargs["fact"] = {"next_prediction": preds_next}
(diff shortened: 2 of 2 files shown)


packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -267,11 +267,26 @@ def calc_scores_per_stream(
...
         da_tars.items(), da_preds.items(), strict=False
     ):
         _logger.debug(f"Verifying data for stream {stream}...")
 
+        # Prepare any metric-specific kwargs, similar for ACC
+        metrics_kwargs = {}
+        for metric in metrics:
+            metrics_kwargs[metric] = {}
+
+        if "fact" in metrics:
```
jpolz: To be discussed where this has to go

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -267,11 +267,42 @@ def calc_scores_per_stream(
...
+
+        if "froct" in metrics:
+            # index of fstep in sorted fsteps
+            fstep_idx = sfsteps.index(fstep)
+            next_fstep = sfsteps[fstep_idx + 1] if fstep_idx + 1 < len(sfsteps) else None
+            if next_fstep is not None:
+                preds_next = da_preds.get(next_fstep, None)
+            else:
+                preds_next = da_preds.get(fstep, None)
+                # set all values to nan if there is no next step
```
Jubeku: Is this the reason we see the 0 in the last forecast step? Would it be better to report `len(fsteps)-1` steps?
image (2)

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -267,11 +267,42 @@ def calc_scores_per_stream(
         },
     )
 
+    # fsteps ordered numerically
+    sfsteps = sorted(fsteps)
```
Jubeku: In which situations are they not sorted? cc @iluise

packages/evaluate/src/weathergen/evaluate/score.py
```diff
@@ -604,6 +606,78 @@ def calc_rmse(
...
+    
+    def calc_troct(
+        self,
+        p: xr.DataArray,
+        gt: xr.DataArray,
+        gt1: xr.DataArray,
+        group_by_coord: str | None = None,
+    ):
+        """
+        Calculate forecast rate of change over time
```
iluise: froct and troct have the same description. Can you add a few more details?

packages/evaluate/src/weathergen/evaluate/utils.py
iluise: This was before we implemented this function: https://github.com/ecmwf/WeatherGenerator/blob/develop/packages/evaluate/src/weathergen/evaluate/io_reader.py#L387

Now everything should be sorted by default

packages/evaluate/src/weathergen/evaluate/score.py
```diff
@@ -604,6 +606,78 @@ def calc_rmse(
 
         return rmse
 
+    def calc_froct(
```
iluise, line 648: since froct and troct have the same underlying formula I propose to refactor them so we call the same method (that can be named e.g. `calc_activity`) from `calc_froct` and `calc_troct`.

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -267,11 +267,42 @@ def calc_scores_per_stream(
...
         da_tars.items(), da_preds.items(), strict=False
     ):
         _logger.debug(f"Verifying data for stream {stream}...")
 
+        # Prepare any metric-specific kwargs, similar for ACC
+        metrics_kwargs = {}
+        for metric in metrics:
+            metrics_kwargs[metric] = {}
+
+        if "froct" in metrics:
```
iluise: we should avoid having metric-related exceptions in these functions as it will soon become messy. 
My solution would be to modify the VerifiedData class https://github.com/ecmwf/WeatherGenerator/blob/develop/packages/evaluate/src/weathergen/evaluate/score.py#L65-L73

And add also the next step. You can fill it here: https://github.com/ecmwf/WeatherGenerator/blob/develop/packages/evaluate/src/weathergen/evaluate/utils.py#L99
with smth like: 
```
score_data = VerifiedData(preds, tars, next_pred, next_tar)
```
and then in here: https://github.com/ecmwf/WeatherGenerator/blob/develop/packages/evaluate/src/weathergen/evaluate/score.py#L258-L260

You can fill gt1 and p1 only when needed :)

packages/evaluate/src/weathergen/evaluate/utils.py
jpolz: Yes this it was my intention to do so. The question is where we want the label, i.e. at forecast step x is froct the difference between x and x+1 or x-1 and x?

iluise: I'd say x vs x+1 as we don't store the initial time fstep = 0

jpolz: Ok. Then it is implemented correctly and I just need to make sure the last step is set to NaN

packages/evaluate/src/weathergen/evaluate/utils.py
jpolz: Thanks for the suggestion. I implemented it that way now, but could not avoid an if/else statement depending on the metric used. See my comment at the respective location

packages/evaluate/src/weathergen/evaluate/score.py
jpolz: done

packages/evaluate/src/weathergen/evaluate/utils.py
jpolz: Done. the score now shows NaN for the last step. The only issue is that the plotting still plots it as 0, but that is not the scope of this PR

packages/evaluate/src/weathergen/evaluate/score.py
```diff
@@ -257,7 +263,15 @@ def get_score(
...
-        args = {"p": data.prediction, "gt": data.ground_truth}
+        if score_name in ["froct", "troct"]:
+            args = {
+                "p": data.prediction,
+                "gt": data.ground_truth,
+                "p_next": data.prediction_next,
+                "gt_next": data.ground_truth_next,
+            }
+        else:
+            args = {"p": data.prediction, "gt": data.ground_truth}
```
jpolz, line 274: @iluise  This is where I moved the cases for different metrics. IMO this is better than adding all args to all metrics. The only way to get around this is to pass the VerifiedData object to the scores as a whole, but then it will not be obvious from the docstrings of scores what they actually use which is not ideal.

PR #907, opened 2025-09-15
[906] Bug fix in tokenizer

tjhunter:
## Type of Change

-   [ ] Documentation update

## Code Compatibility

### Dependencies

### Documentation

### DIFF
--- a/src/weathergen/datasets/tokenizer.py
+++ b/src/weathergen/datasets/tokenizer.py
@@ -76,27 +76,46 @@ class Tokenizer:
             vertsmm_rots.to(torch.float32),
         ]
 
+        transforms = [
+            ([verts10, verts11, verts01, vertsmm], verts00_rots),
+            ([verts00, verts11, verts01, vertsmm], verts10_rots),
+            ([verts00, verts10, verts01, vertsmm], verts11_rots),
+            ([verts00, verts11, verts10, vertsmm], verts01_rots),
+            ([verts00, verts10, verts11, verts01], vertsmm_rots),
+        ]
+
         self.verts_local = []
-        verts = torch.stack([verts10, verts11, verts01, vertsmm])
-        verts_trans: list[torch.Tensor] = [t for t in verts.transpose(0, 1)]
-        temp = ref - locs_to_cell_coords_ctrs(verts00_rots, verts_trans)
-        self.verts_local.append(temp.flatten(1, 2))
+        for (_verts, rot) in transforms:
+            # Compute local coordinates
+            verts = torch.stack(_verts)
+            verts = verts.transpose(0, 1)
+            # Perform the rotation:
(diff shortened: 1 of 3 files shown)


src/weathergen/datasets/tokenizer.py
```diff
@@ -76,26 +75,24 @@ def __init__(self, healpix_level: int):
...
+        ]
 
-        verts = torch.stack([verts00, verts10, verts11, verts01])
-        temp = ref - torch.stack(locs_to_cell_coords_ctrs(vertsmm_rots, verts.transpose(0, 1)))
-        self.verts_local.append(temp.flatten(1, 2))
+        self.verts_local = []
+        for _verts, rot in transforms:
+            # Compute local coordinates
+            verts = torch.stack(_verts)
+            # <healpix, 4, 3>
```
shmh40: This comment is just the dimensions of verts?

src/weathergen/datasets/tokenizer.py
```diff
@@ -76,26 +75,24 @@ def __init__(self, healpix_level: int):
             vertsmm_rots.to(torch.float32),
         ]
 
-        self.verts_local = []
-        verts = torch.stack([verts10, verts11, verts01, vertsmm])
-        temp = ref - torch.stack(locs_to_cell_coords_ctrs(verts00_rots, verts.transpose(0, 1)))
```
shmh40, line 81: This function locs_to_cell_coords_ctrs still exists in utils.py (rewritten from before), but now we don't use it? We can remove it from utils.py or no?

src/weathergen/datasets/tokenizer.py
```diff
@@ -76,26 +75,24 @@ def __init__(self, healpix_level: int):
...
-        temp = ref - torch.stack(locs_to_cell_coords_ctrs(vertsmm_rots, verts.transpose(0, 1)))
-        self.verts_local.append(temp.flatten(1, 2))
+        self.verts_local = []
+        for _verts, rot in transforms:
+            # Compute local coordinates
+            verts = torch.stack(_verts)
+            # <healpix, 4, 3>
+            verts = verts.transpose(0, 1)
+            # Perform the rotation using batch matrix multiplication
+            t1 = torch.bmm(rot, verts.transpose(-1, -2)).transpose(-2, -1)
```
shmh40, line 95: Why do we transpose verts and then transpose again without ? I see this was what we had before so happy to leave it as is for now.

src/weathergen/datasets/tokenizer.py
tjhunter: locs_to_cell_coords_ctrs is used also in utils.py for grouping the points together. the tokenizer actually does not need the full functionality of it, hence not calling it.

src/weathergen/datasets/tokenizer.py
tjhunter: I updated the comment to reflect that, good point.

PR #912, opened 2025-09-16
[554] Updates the PR template

tjhunter:
Makes the PR template shorter and more action oriented. We do not respect the current checklist, which is more appropriate for large projects anyway.


grassesi (approved):
looks much better :), maybe it is worth also keeping a check for docstrings/comments?

PR #914, opened 2025-09-16
Add the calculation of 10ff for ERA5 and CERRA 

### LINKED ISSUE #567: 10ff score

One standard headline score is 10ff, wind speed at nominally 10m. We do not compute it currently.

Compute 10ff when 10u,10v or equivalent (u10m, u10v; obsvalue_u10m, obsvalue_v10m) are available.

CC @iluise @mlangguth89 

_No response_

### Additional context

_No response_

### Organisation

_No response_

SavvasMel:
## Description

One of important scores relates to the `10ff`, the mean total wind speed. This is not included in ERA5. It is included in CERRA but not as a total mean, but evaluated for each grid point under the name  `10si`.

This PR takes the 10u and 10v from ERA5 and the `10si` from CERRA and calculates the mean total wind speed `10ff`.

I believe that a new feature should be added in the near future, if more such  scores need to be calculated, in the form of a class for such cases. If 10ff needs to be included for even more datasets the current code/PR is easily adaptable. 

This PR can be tested with this config below. The only thing I changed from the current template is that I added a `calc_channels` key which can have all the channels we want to calculate from existing ones:

[Link to config](https://gitlab.jsc.fz-juelich.de/esde/WeatherGenerator-private/-/blob/086c37b6a297afa9e081e00c83f51b98a09bf962/evaluate/plot_config_template.yml)

## Issue Number

Closes #567

-   [X] I have informed and aligned with people impacted by my change:

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/derived_channels.py
+++ b/packages/evaluate/src/weathergen/evaluate/derived_channels.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from dataclasses import dataclass
 
 import numpy as np
@@ -16,41 +17,62 @@ class DeriveChannels:
         channels: list,
         stream_cfg: dict,
     ):
+        """
+        Initializes the DeriveChannels class with necessary configurations for channel derivation.
+
+        Args:
+            available_channels (np.array): an array of all available channel names in the datasets (target or pred).
+            channels (list): A list of channels of interest to be evaluated and/or plotted.
+            stream_cfg (dict): A dictionary containing the stream configuration settings for evaluation and plottings.
+
+        Returns:
+            None
+        """
         self.available_channels = available_channels
         self.channels = channels
         self.stream_cfg = stream_cfg
 
-    def calc_10ff_channel(self, da: xr.DataArray) -> xr.DataArray | None:
+    def calc_xxff_channel(self, da: xr.DataArray, level: str) -> xr.DataArray | None:
         """
-        Calculate 10m wind speed ('10ff') from wind components or directly.
+        Calculate wind speed at xx level ('xxff') from wind components or directly.
         Args:
             da: xarray DataArray with data
         Returns:
-            xarray: Calculated 10ff value, or None if calculation is not possible
+            xarray: Calculated xxff value, or None if calculation is not possible
         """
 
-        stream = da.stream.values
         channels = da.channel.values
 
-        if stream == "ERA5" and "10si" not in channels:
-            if "10u" in channels and "10v" in channels:
-                u_component = da.sel(channel="10u")
-                v_component = da.sel(channel="10v")
+        if f"{level}si" not in channels:
+            for suffix in ["u", "v"]:
+                for name in [
+                    f"{level}{suffix}",
+                    f"{suffix}_{level}",
+                    f"obsvalue_{suffix}{level}m_0",
+                ]:
+                    component = da.sel(channel=name) if name in channels else None
+                    if component is not None:
+                        break
+                if suffix == "u":
+                    u_component = component if component is not None else None
+                else:
+                    v_component = component if component is not None else None
+            if not (u_component is None or v_component is None):
                 ff = np.sqrt(u_component**2 + v_component**2)
                 return ff
             else:
-                _logger.debug("10u or 10v not found - skipping 10ff calculation")
+                _logger.debug(
+                    f"u or v not found for level {level} - skipping {level}ff calculation"
+                )
                 return None
-
-        elif stream in ["CERRA", "ERA5"] and "10si" in channels:
-            ff = da.sel(channel="10si")
+        elif f"{level}si" in channels:
+            ff = da.sel(channel=f"{level}si")
             return ff
-
         else:
-            _logger.debug("Skipping 10ff calculation - unsupported data format")
+            _logger.debug(f"Skipping {level}ff calculation - unsupported data format")
             return None
 
-    def get_channel(self, tag, calc_func) -> None:
+    def get_channel(self, data_tars, data_preds, tag, level, calc_func) -> None:
         """
         Add a new channel data to both target and prediction datasets.
 
@@ -59,8 +81,8 @@ class DeriveChannels:
         If the calculation returns None, the original datasets are preserved unchanged.
 
         The method updates:
-        - self.data_tars: Target dataset with added 10ff channel
-        - self.data_preds: Prediction dataset with added 10ff channel
+        - data_tars: Target dataset with added 10ff channel
+        - data_preds: Prediction dataset with added 10ff channel
         - self.channels: Channel list with '10ff' added
 
         Returns:
@@ -69,8 +91,8 @@ class DeriveChannels:
 
         data_updated = []
 
-        for data in [self.data_tars, self.data_preds]:
-            new_channel = calc_func(data)
+        for data in [data_tars, data_preds]:
+            new_channel = calc_func(data, level)
 
             if new_channel is not None:
                 conc = xr.concat(
@@ -90,7 +112,8 @@ class DeriveChannels:
             else:
                 data_updated.append(data)
 
-        self.data_tars, self.data_preds = data_updated
+        data_tars, data_preds = data_updated
+        return data_tars, data_preds
 
     def get_derived_channels(
         self,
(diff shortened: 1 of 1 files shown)


packages/evaluate/src/weathergen/evaluate/io_reader.py
```diff
@@ -361,6 +369,58 @@ def get_channels(self, stream: str) -> list[str]:
...
 
+    def calc_channels(
+        self,
+        data_tars: dict[xr.DataArray],
+        data_preds: dict[xr.DataArray],
+        available_channels: np.array,
+        existing_channels: list,
+        stream_cfg: dict,
+    ) -> dict[xr.DataArray]:
+        def calc_10ff(da):
```
iluise: Can we move `calc_10ff` to a new file called `derived_channels.py` and set a getter like `get_10ff` that takes care of the ERA5/CERRA/others renaming?

I would put there also `calc_channels`, and call it smth like "get_derived_channels" or smth like that

packages/evaluate/src/weathergen/evaluate/io_reader.py
```diff
@@ -306,6 +306,14 @@ def get_data(
                                 f"The following channels were not found: {list(set(channels) - set(existing_channels))}. Skipping them."
```
iluise: for "10ff" ERA5 this will be triggered because it is checked before we calculate it, right?

packages/evaluate/src/weathergen/evaluate/io_reader.py
```diff
@@ -306,6 +306,14 @@ def get_data(
...
                             )
 
+                        da_tars_fs, da_preds_fs, existing_channels = self.calc_channels(
```
iluise: maybe we can avoid returning a new `existing_channels` if we get the derive quantities before calculating the available_channels?

packages/evaluate/src/weathergen/evaluate/io_reader.py
SavvasMel: No it will not because I am not including the `10ff ` into the `channels` in the config. I have another key for example:

```
    streams:
      ERA5:
        channels: ["2t"]
        calc_channels: ["10ff"]    <-------
        evaluation:
          forecast_step: "all"
          sample: [0, 1]
        plotting:
          sample: [0]
          forecast_step: [1, 3, 4, 5]
          plot_maps: true
          plot_histograms: false
          plot_animations: false

```

So this one does not go into the `check_availability`. 

On another point, irrelevant to "10ff", now that we have the `check_availability` functionality, I am not sure if the below is even needed anymore:

```
  if len(existing_channels) < len(channels):
      _logger.warning(
          f"The following channels were not found: {list(set(channels) - set(existing_channels))}. Skipping them.")
```

Or to say it differently, it is always `len(existing_channels) = len(channels)`.

packages/evaluate/src/weathergen/evaluate/io_reader.py
SavvasMel: Perfect, I will work on it!

packages/evaluate/src/weathergen/evaluate/io_reader.py
SavvasMel: I think I see your point ok!

packages/evaluate/src/weathergen/evaluate/io_reader.py
iluise: you are right, the check is not needed anymore here we can remove it!
about the 10ff: IMO we should compute it only if it is listed among the variables and it fails if 10u and 10v are not present in zarrIO.channels, but it is not necessary to list them in the plot_config. What do you think?

SavvasMel: I was about to do that, but then the variable is deleted because of the `check_availability ` functionality. The data reading takes place after so I am not sure how we can list it among the variables. In some way the calculation then needs to take place before `check_availability `  if we are about to list it in one of the variables. Do I miss something here?

iluise: Ah yes, you're right. Then let's keep it as you said, with `calc_channels` and if it's annoying for the users then we will change it

SavvasMel: I will change it to `derive_channels`. Much more intuitional.

packages/evaluate/src/weathergen/evaluate/derived_channels.py
```diff
@@ -0,0 +1,130 @@
...
+
+        Returns:
+        --------
+        - data_tars: Updated target dataset (if channel can be added)
+        - data_preds:  Updated prediction dataset (if channel can be added)
+        - self.channels: all the channels of interest
+
+        """
+
+        self.data_tars = data_tars
```
iluise: Is there a reason why you needed to add self.data_tars and self.data_preds instead of passing  data_tars and data_preds directly to get_channel? 
If we pass them directly we always know which dataset is being processed, while with self it can still come from the previous iteration and it can be tricky to track.

packages/evaluate/src/weathergen/evaluate/derived_channels.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+@dataclass
+class DeriveChannels:
+    def __init__(
+        self,
+        available_channels: np.array,
+        channels: list,
+        stream_cfg: dict,
+    ):
```
iluise, line 19: can you add a description of what these entries are? (especially to clarify the distinction between available_channels and channels)

packages/evaluate/src/weathergen/evaluate/derived_channels.py
```diff
@@ -0,0 +1,130 @@
...
+        Args:
+            da: xarray DataArray with data
+        Returns:
+            xarray: Calculated 10ff value, or None if calculation is not possible
+        """
+
+        stream = da.stream.values
+        channels = da.channel.values
+
+        if stream == "ERA5" and "10si" not in channels:
```
iluise: Adding it here so it doesn't get lost from Mattermost: use a regex to be as general as possible as suggested by Christian

packages/evaluate/src/weathergen/evaluate/derived_channels.py
```diff
@@ -0,0 +1,130 @@
...
+    def get_channel(self, tag, calc_func) -> None:
+        """
+        Add a new channel data to both target and prediction datasets.
+
+        This method computes new channel values using given calculations methods
+        and appends them as a new channel to both self.data_tars and self.data_preds.
+        If the calculation returns None, the original datasets are preserved unchanged.
+
+        The method updates:
+        - self.data_tars: Target dataset with added 10ff channel
```
iluise: Not sure I understand why we need self here instead of just returning `data_tars` and `data_preds`. See below

packages/evaluate/src/weathergen/evaluate/io_reader.py
```diff
@@ -465,25 +472,13 @@ def get_data(
...
-                            existing_channels,
-                            stream_cfg,
+
+                        da_tars_fs, da_preds_fs, channels = dc.get_derived_channels(
+                            da_tars_fs, da_preds_fs
                         )
 
-                        da_tars_fs = da_tars_fs.sel(channel=existing_channels)
-                        da_preds_fs = da_preds_fs.sel(channel=existing_channels)
+                        da_tars_fs = da_tars_fs.sel(channel=channels)
```
iluise, line 480: What happens now if we have a weird channel like e.g. "blah"? Does it get removed without crashing as before?

packages/evaluate/src/weathergen/evaluate/derived_channels.py
SavvasMel: Fixed

packages/evaluate/src/weathergen/evaluate/derived_channels.py
SavvasMel: Fixed

packages/evaluate/src/weathergen/evaluate/derived_channels.py
SavvasMel: Fixed I added a regex

packages/evaluate/src/weathergen/evaluate/derived_channels.py
SavvasMel: Fixed

packages/evaluate/src/weathergen/evaluate/io_reader.py
SavvasMel: `INFO:weathergen.evaluate.io_reader:Requested channel(s) {'blah'} do(es) not exist in Zarr. Removing missing channel(s) for evaluation.`

PR #959, opened 2025-09-25
Sophiex/dev/fsdp2 fix

### LINKED ISSUE #669: FSDP2

Trying out FSDP2 after initial exploration of DeepSpeed showed some potential issues.

_No response_

_No response_

### Additional context

_No response_

### Organisation

_No response_

sophie-xhonneux:
## Description

Work-around for FSDP 2 issue detailed here github.com/pytorch/pytorch/issues/158719

## Issue Number

Closes #669 


sophie-xhonneux:
Convergence on pretraining mixed streams looks excellent (FSDP2 at the top, develop at the bottom)
image

packages/common/src/weathergen/common/io.py
```diff
@@ -68,6 +71,45 @@ def create(cls, other: typing.Any) -> typing.Self:
 
         return cls(**dataclasses.asdict(other))
 
+    @classmethod
+    def spoof(
+        cls, other: typing.Any, nchannels, datetime, geoinfo_size, mean_of_data
+    ) -> typing.Self:
+        """
+        Spoof an instance from data_reader_base.ReaderData instance.
```
tjhunter: could you explain why we need to do it and point to the torch issue? 
without context, no one except you will remember why we need that in 6 mothns (and the bug might get fixed by the pytorch team, who knows)

src/weathergen/datasets/data_reader_obs.py
```diff
@@ -86,7 +88,7 @@ def length(self) -> int:
 
     def select_channels(
         self, colnames: list[str], cols_select: list[str] | None, cols_exclude: list[str] | None
-    ) -> None:
+    ) -> list:
```
tjhunter: even `) -> list[str]:`

packages/common/src/weathergen/common/io.py
```diff
@@ -68,6 +71,45 @@ def create(cls, other: typing.Any) -> typing.Self:
...
+        other should be such an instance.
+        """
+
+        hl = 5
```
tjhunter: this one should be a parameter. It is important because other people have been experimenting with the healpix level, and it would fail in mysterious shape errors. If it does not matter, then we should fix healpix level = 1 with a comment.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -334,23 +339,29 @@ def __iter__(self):
...
+                            torch.from_numpy(rdata.geoinfos),
+                            torch.from_numpy(rdata.data),
+                            rdata.datetimes,
+                            (time_win1.start, time_win1.end),
+                            ds,
+                        )
 
-                            stream_data.add_source(rdata_wrapped, ss_lens, ss_cells, ss_centroids)
+                        stream_data.add_source(rdata_wrapped, ss_lens, ss_cells, ss_centroids)
+                        stream_data.is_spoof = sample_is_empty
```
tjhunter, line 363: this field should be declared in `streamdata`

```py
class StreamData:
    """
    StreamData object that encapsulates all data the model ingests for one batch item
    for one stream.
    """

    # TODO: comment
    is_spoof: bool = False

```

Adding fields on objects and relying on them existing or not breaks interface contracts (and type checkers. vscode complains about it for valid reasons).

src/weathergen/model/model.py
```diff
@@ -34,17 +36,82 @@
...
+        bs = cf.batch_size_per_gpu
+        nqs = 9
+        s = [bs, self.num_healpix_cells, cf.ae_local_num_queries, cf.ae_global_dim_embed]
+
+        ### POSITIONAL EMBEDDINGS ###
+        len_token_seq = 1024
+        self.pe_embed = torch.nn.Parameter(
+            torch.zeros(len_token_seq, cf.ae_local_dim_embed, dtype=self.dtype), requires_grad=False
+        )
+        # self.pe_embed = torch.zeros(len_token_seq, cf.ae_local_dim_embed, dtype=self.dtype)
```
tjhunter: remove or justify keeping with a comment

src/weathergen/model/model.py
```diff
@@ -34,17 +36,82 @@
...
+        self.healpix_level = cf.healpix_level
+        self.num_healpix_cells = 12 * 4**cf.healpix_level
+        self.dtype = get_dtype(cf.attention_dtype)
+
+        bs = cf.batch_size_per_gpu
+        nqs = 9
+        s = [bs, self.num_healpix_cells, cf.ae_local_num_queries, cf.ae_global_dim_embed]
+
+        ### POSITIONAL EMBEDDINGS ###
+        len_token_seq = 1024
```
tjhunter, line 62: this is a parameter I think?

src/weathergen/model/model.py
```diff
@@ -34,17 +36,82 @@
...
+        )
+        self.q_cells_lens.data[0] = 0
+        # self.q_cells_lens = torch.ones(self.num_healpix_cells + 1, dtype=torch.int32)
+        # self.q_cells_lens[0] = 0
+
     def create(self, cf: Config) -> "ModelParams":
+        self.reset_parameters(cf)
+        return self
+
+    def reset_parameters(self, cf: Config) -> "ModelParams":
```
tjhunter, line 108: `def reset_parameters(self, cf: Config) -> None:`
Also you already have self.cf, no need to pass it

src/weathergen/model/model.py
```diff
@@ -788,29 +868,30 @@ def predict(
...
-                            tc_embed,
-                            streams_data[i_b][ii].target_coords[fstep],
-                            use_reentrant=False,
-                        )
-                        if len(streams_data[i_b][ii].target_coords[fstep].shape) > 1
-                        else streams_data[i_b][ii].target_coords[fstep]
-                        for i_b in range(len(streams_data))
-                    ]
-                )
+            # arguably we should to the mixed precision policy when creating the model in FSDP
```
tjhunter, line 850: is it a TODO?

src/weathergen/train/loss_calculator.py
```diff
@@ -232,6 +232,11 @@ def compute_loss(
 
             loss_fsteps = torch.tensor(0.0, device=self.device, requires_grad=True)
             ctr_fsteps = 0
+            stream_is_spoof = streams_data[i_batch][i_stream_info].is_spoof
```
tjhunter, line 235: such a hack... but it works

tjhunter: nice solution given all the constraints!

src/weathergen/train/trainer_base.py
```diff
@@ -68,90 +66,84 @@ def init_torch(use_cuda=True, num_accs_per_task=1, multiprocessing_method="fork"
...
+            if is_root():
+                print("DDP initialized: root.")
+            # Wait for all ranks to reach this point
+
+            dist.barrier()
+            # communicate run id to all nodes
+            len_run_id = len(cf.run_id)
+            run_id_int = torch.zeros(len_run_id, dtype=torch.int32).to(device)
+            if is_root():
+                print(f"Communicating run_id to all nodes: {cf.run_id}")
```
tjhunter, line 125: I understand the initialization of the loggers is tricky, but it will bite us in the future. can you create an issue to fix the logging here back to using python logging?

src/weathergen/utils/logger.py
```diff
@@ -43,48 +95,8 @@ def format(self, record, *args, **kwargs):
...
-
-        handler.setLevel(stream_handle)
-        handler.setFormatter(formatter)
-        logger.addHandler(handler)
-
-    return logger
-
-
 @cache
-def init_loggers(
```
tjhunter, line 80: I see the @cache is kept in the end -> there should not be the need to remove the other calls to `init_loggers()` elsewhere? Fine if we do that as a cleanup followup

src/weathergen/utils/logger.py
```diff
@@ -103,32 +115,35 @@ def init_loggers(
...
+    # timestamp = now.strftime("%Y-%m-%d-%H%M")
+
+    # output_dir = f"./output/{timestamp}-{run_id}"
+    output_dir = f"./output/{run_id}"
+
+    # load the structure for logging config
+    if logging_config is None:
+        logging_config = json.loads(LOGGING_CONFIG)
+
+    for _, handler in logging_config["handlers"].items():
```
tjhunter, line 130: fyi: you can do `for v in my_dict.values()`

src/weathergen/utils/logger.py
```diff
@@ -103,32 +115,35 @@ def init_loggers(
...
+        for k, v in handler.items():
+            if k == "formatter":
+                handler[k] = v
+            elif k == "filename":
+                filename = f"{output_dir}/{run_id}-{v}"
+                ofile = pathlib.Path(filename)
+                # make sure the path is independent of path where job is launched
```
tjhunter, line 137: given the amount of postprocessing that still happens to convert on the json dict to convert it to the right format, I would personally have simply a function that does all that on dictionaries already. it is just as verbose (chatgpt output). This is a suggestion, your call.

```py
def _make_config(debug_filename, error_filename) -> dict:
    return dict(
        version=1,
        disable_existing_loggers=False,
        formatters={
            "custom": {
                "class": "weathergen.utils.logger.ColoredRelPathFormatter",
                "format": \
                    "%(asctime)s %(process)d %(filename)s:%(lineno)d : %(levelname)-8s : %(message)s"
            }
        },
        handlers={
            "stdout": {
                "class": "logging.StreamHandler",
                "level": "INFO",
                "formatter": "custom",
                "stream": "ext://sys.stdout"
            },
            "stderr": {
                "class": "logging.StreamHandler",
                "level": "ERROR",
                "formatter": "custom",
                "stream": "ext://sys.stderr"
            },
            "logfile": {
                "class": "logging.FileHandler",
                "level": "DEBUG",
                "formatter": "custom",
                "filename": debug_filename,
                "mode": "w"
            },
            "errorfile": {
                "class": "logging.FileHandler",
                "level": "ERROR",
                "formatter": "custom",
                "filename": error_filename,
                "mode": "w"
            }
        },
        root={
            "level": "DEBUG",
            "handlers": [
                "stderr",
                "stdout",
                "logfile",
                "errorfile"
            ]
        }
    )

```

src/weathergen/model/model.py
```diff
@@ -34,17 +36,82 @@
...
+            self.num_healpix_cells,
+            cf.ae_local_num_queries,
+            cf.ae_global_dim_embed,
+            dtype=self.dtype,
+        )
+        self.pe_global = torch.nn.Parameter(pe, requires_grad=False)
+        # self.pe_global = pe
+
+        ### HEALPIX NEIGHBOURS ###
+        hlc = self.healpix_level
```
tjhunter, line 76: I am confused where this code is coming from: is it a new feature?

src/weathergen/datasets/stream_data.py
```diff
@@ -106,8 +106,9 @@ def add_empty_source(self, source: IOReaderData) -> None:
         None
         """
 
+        source = IOReaderData.spoof(source)
```
tjhunter: this code was probably not tried out since it would immediately generate a runtime error. I wonder if we need it here? That looks more like debugging.

src/weathergen/train/trainer.py
```diff
@@ -627,24 +679,151 @@ def validate(self, epoch):
...
+        if not is_model_sharded:
+            self.model = self.model.to(self.device)
+
+        if len(mkeys) > 0:
+            logger.warning(f"Missing keys when loading model: {mkeys}")
+
+        if len(ukeys) > 0:
+            logger.warning(f"Unused keys when loading model: {mkeys}")
+
+    def load_optim(self):
```
tjhunter, line 726: we should just remove this piece for the time being, this is not ready yet

src/weathergen/train/trainer.py
```diff
@@ -187,18 +194,8 @@ def run(self, cf, run_id_contd=None, epoch_contd=None):
...
-            logger.info(f"Continuing run with id={run_id_contd} at epoch {epoch_contd}.")
-            self.model.load(run_id_contd, epoch_contd)
-            if cf.with_ddp and cf.with_fsdp:
-                FSDP.set_state_dict_type(
-                    self.model,
-                    StateDictType.FULL_STATE_DICT,
-                    FullStateDictConfig(rank0_only=False),
-                )
-            logger.info(f"Loaded model id={run_id_contd}.")
+        with torch.device("meta"):
```
tjhunter, line 197: a bit confused here about this device. can you leave a comment if we keep it?

src/weathergen/train/trainer.py
```diff
@@ -227,25 +218,80 @@ def run(self, cf, run_id_contd=None, epoch_contd=None):
...
+            for module in self.model.target_token_engines.modules():
+                if isinstance(module, modules_to_shard):
+                    fully_shard(module, **full_precision_fsdp_kwargs)
+
+        self.model_params = ModelParams(cf).create(cf)  # .to(device)
+
+        if cf.with_ddp and cf.with_fsdp:
+            fully_shard(self.model)
+            for tensor in itertools.chain(self.model.parameters(), self.model.buffers()):
+                assert tensor.device == torch.device("meta")
```
tjhunter, line 275: assert tensor.device == torch.device("meta"), tensor.device

config/default_config.yml
```diff
@@ -127,7 +127,7 @@ weight_decay: 0.1
 norm_type: "LayerNorm"
 nn_module: "te"
 
-start_date: 197901010000
+start_date: 199001010000
```
tjhunter: do you mean to commit this change?

src/weathergen/model/embeddings.py
```diff
@@ -124,7 +124,10 @@ def __init__(
             assert self.unembed_mode == "block"  # only supported mode at the moment
             # padding needed if the unembedded columns cannot be concatenated to dim_out (e.g GPSRO)
             self.pad = self.dim_out % token_size
-            self.out_pad = torch.nn.Parameter(torch.zeros(self.pad))
+            self.out_pad = torch.nn.Parameter(
```
clessig: This should not get gradients. Let's please fix it here.

src/weathergen/model/model.py
clessig: This is `cf.ae_local_dim_embed `

src/weathergen/model/model.py
```diff
@@ -34,17 +36,82 @@
...
+        # self.pe_embed = torch.zeros(len_token_seq, cf.ae_local_dim_embed, dtype=self.dtype)
+
+        pe = torch.zeros(
+            self.num_healpix_cells,
+            cf.ae_local_num_queries,
+            cf.ae_global_dim_embed,
+            dtype=self.dtype,
+        )
+        self.pe_global = torch.nn.Parameter(pe, requires_grad=False)
+        # self.pe_global = pe
```
clessig: Remove

src/weathergen/model/model.py
clessig: This code was there before. But why is it flagged here as a change?

src/weathergen/model/model.py
clessig: In which precision is this running now? We changed this because of serious issues in April/May.

src/weathergen/run_train.py
```diff
@@ -59,13 +61,17 @@ def inference_from_args(argl: list[str]):
     )
     cf = config.set_run_id(cf, args.run_id, args.reuse_run_id)
 
-    fname_debug_logging = f"./logs/debug_log_{cf.run_id}.txt"
-    init_loggers(logging_level=logging.DEBUG, debug_output_streams=fname_debug_logging)
+    devices = Trainer.init_torch()
```
clessig, line 64: I still quite strongly disagree that we should have this explicitly here. Why don't we call `init_loggers` in `Trainer.init()`. Then we would have 1 line instead of 4 or 5 that are repeated in all functions in this file.

src/weathergen/utils/logger.py
```diff
@@ -7,14 +7,66 @@
...
 import logging
+import logging.config
 import os
 import pathlib
-import sys
 from functools import cache
 
 from weathergen.common.config import _load_private_conf
 
+LOGGING_CONFIG = """
```
clessig, line 19: This should go to a separate file, not in here verbatim.

src/weathergen/train/trainer.py
```diff
@@ -227,25 +218,80 @@ def run(self, cf, run_id_contd=None, epoch_contd=None):
...
-                    param_dtype=self.mixed_precision_dtype, cast_forward_inputs=True
+            fsdp_kwargs = {
+                "mp_policy": MixedPrecisionPolicy(
+                    param_dtype=self.mixed_precision_dtype,
+                    reduce_dtype=torch.float32,
                 )
+                if cf.with_mixed_precision
+                else None,
+            }
+            modules_to_shard = (
```
clessig, line 226: This needs to go out of this function. And I think it should actually go to model.py or a some helper function/class.

src/weathergen/train/trainer.py
clessig: Yes, remove.

src/weathergen/train/trainer.py
```diff
@@ -627,24 +679,151 @@ def validate(self, epoch):
...
         return (
-            [[d.to_device() for d in db] for db in batch[0]],
-            batch[1].to("cuda"),
-            [[b.to("cuda") for b in bf] for bf in batch[2]],
+            [[d.to_device(self.device) for d in db] for db in batch[0]],
+            batch[1].to(self.device),
+            [[b.to(self.device) for b in bf] for bf in batch[2]],
         )
 
+    def load_model(self, run_id: str, epoch=-1):
```
clessig, line 684: Does this need to be in Trainer or can it go back to model.py. fsdp1 made it necessary to move it to Trainer (can't remember the details but could look it up) but I don't think it's necessary, and logically it doesn't belong here.

src/weathergen/train/trainer.py
```diff
@@ -627,24 +679,151 @@ def validate(self, epoch):
...
-        if self.cf.with_ddp and self.cf.with_fsdp:
-            with FSDP.state_dict_type(
-                self.ddp_model,
-                StateDictType.FULL_STATE_DICT,
-                FullStateDictConfig(rank0_only=True, offload_to_cpu=True),
-            ):
-                state = self.ddp_model.state_dict()
-        else:
-            state = self.ddp_model.state_dict()
+        model_state_dict = self._get_full_model_state_dict()
```
clessig, line 822: Can we move save_model back to model,py?

packages/common/src/weathergen/common/io.py
```diff
@@ -68,6 +70,45 @@ def create(cls, other: typing.Any) -> typing.Self:
...
+        dx = 0.5
+        dy = 0.5
+        other_copy = deepcopy(other)
+        num_healpix_cells = 12 * 4**hl
+        lons, lats = hp.healpix_to_lonlat(
```
clessig: Why do we use all cells and not just one random. From ERA5 global it's perhaps natural but from the point of view of SYNOP, CERRA etc it is better to have one random cell (and ideally from the domain). And one random cell would be identical in expectation but have far less computational overhead.

clessig (commented):
Thanks for the huge effort. There are various details that need to be cleaned up. I don't think it's realistic to do it in the PR but can we make sure there are follow up PRs and these get implemented ASAP.

src/weathergen/model/model.py
tjhunter: then it should be `len_token_seq = cf.ae_local_dim_embed ` ? Otherwise there will be confusing errors when someone changes the size of the embeddings

src/weathergen/run_train.py
tjhunter: I recall the issue is how to distribute the run_id to the worker processes (they need it for logging into the right file). 

I forgot is there a more elegant to pass it, for example using env variables? At least for slurm, the RUN_ID env should be defined everywhere, unsure for the interactive sessions

tjhunter: For that part, I am open to a later cleanup - as long as we have a GH issue and we do it :-)

src/weathergen/train/trainer_base.py
sophie-xhonneux: as discussed there is a chicken and egg problems because the loggers need the same run id, which is only communicated after `init_ddp`

src/weathergen/utils/logger.py
sophie-xhonneux: let's make it a follow-up

src/weathergen/train/trainer.py
sophie-xhonneux: meta device is fairly standard to prevent very large CPU allocations that then have to be moved to GPU, this is standard, only thing we add as a comment is to point to the docs

config/default_config.yml
sophie-xhonneux: no

packages/common/src/weathergen/common/io.py
sophie-xhonneux: This worked in terms of convergence and is akin to a uniform prior on all cells, which except for local re-analysis and the GEOs individually is quite good

src/weathergen/utils/logger.py
sophie-xhonneux: I know I prefer this, because items() has very clear mechanics in terms of order that I am not sure if values has them as well

src/weathergen/utils/logger.py
sophie-xhonneux: Should be a future logging PR

packages/common/src/weathergen/common/io.py
sophie-xhonneux: this is already done in the very top of the issue ;)

src/weathergen/train/trainer.py
sophie-xhonneux: It should be in the trainer conceptually, this is the norm

src/weathergen/train/trainer.py
sophie-xhonneux: I don't think so because it needs to be aware of the device mesh among other things

clessig (approved):
The PR is functional as far as we can tell currently. There are some remaining software engineering issues but it's better to get it merged and address them in follow up PRs.

PR #961, opened 2025-09-25
encoding targets into latent space

### LINKED ISSUE #941: encoding target variales in the latent space

For various reasons we want to have the targets also encoded in the latent space the same way the sources are encoded. This is implemented for the latent loss (https://github.com/ecmwf/WeatherGenerator/issues/706) but the target encoding part needs to be cleaned  up to be merged to develop.

_No response_

_No response_

### Additional context

_No response_

### Organisation

_No response_

kctezcan:
## Description

Encoding target variables into the latent space, similar to sources.

The changes are made for the forecasting mode, tested for both training and inference.

Some open points:
1. How to name the variables and functions? Now everything is called "..._srclk" as abbreviation to "source like", indicating that the targets are being processed as sources.
2. Indeed in some cases it is possible to reuse some code for processing sources by looping through targets fsteps and processing each seperately. But this requires significant rewriting of the existing functions since the funcitons do not operate on simple input variables but input objects and access their fields.
3. Is it also relevant in the MTM mode? This is not tested.
4. A test with multiple datasets still needs to be done.
5. The saving of the latent variables is left for future work

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/941
Refs https://github.com/ecmwf/WeatherGenerator/issues/941
Refs #941


src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -386,15 +386,15 @@ def __iter__(self):
                                     np.concatenate((rdata.coords, rdata.geoinfos, rdata.data), 1)
                                 )
                                 (tt_cells_srclk, tt_lens_srclk, tt_centroids_srclk) = (
-                                    self.tokenizer.batchify_source( # TODO: KCT, check if anything source related is happening in the function
+                                    self.tokenizer.batchify_source(  # TODO: KCT, check if anything source related is happening in the function
```
clessig: Can we please remove all the KCT's. We missed a few last time ...

Also, what's the question here?

src/weathergen/datasets/tokenizer_forecast.py
```diff
@@ -42,14 +42,18 @@ def batchify_source(
         times: np.array,
         time_win: tuple,
         normalizer,  # dataset,
-        use_normalizer: str, # "source" or "target"
+        use_normalizer: str,  # "source" or "target"
```
clessig: Remove whitespace

src/weathergen/datasets/tokenizer_masking.py
```diff
@@ -48,14 +48,18 @@ def batchify_source(
         times: np.array,
         time_win: tuple,
         normalizer,  # dataset
-        use_normalizer: str, # "source" or "target"
+        use_normalizer: str,  # "source" or "target"
```
clessig: Remove whitespace

src/weathergen/datasets/utils.py
```diff
@@ -694,48 +695,54 @@ def compute_offsets_scatter_embed_target_srclk(batch: StreamData) -> StreamData:
...
-                            torch.stack(
-                                [
-                                    s.target_srclk_tokens_lens[fstep]
-                                    if len(s.target_srclk_tokens_lens[fstep]) > 0
-                                    else torch.tensor([])
-                                    for fstep in range(len(s.target_srclk_tokens_lens))
-                                ]
-                            )
-                            for s in stl_b
+                            s.target_srclk_tokens_lens[fstep]
```
clessig: Why did this change? Or is this from a different PR?

src/weathergen/datasets/utils.py
```diff
@@ -694,48 +695,54 @@ def compute_offsets_scatter_embed_target_srclk(batch: StreamData) -> StreamData:
...
+        (offsets_base.shape[0], 1), dtype=offsets_base.dtype, device=offsets_base.device
+    )
+    offsets = torch.cat([zeros_col, offsets_base[:, :-1]], dim=1)
     offsets_pe = torch.zeros_like(offsets)
 
     for ib, sb in enumerate(batch):
         for itype, s in enumerate(sb):
             for fstep in range(offsets.shape[0]):
-                if not (target_srclk_tokens_lens[ib, itype, fstep].sum() == 0): # if not empty
+                if not (target_srclk_tokens_lens[ib, itype, fstep].sum() == 0):  # if not empty
```
clessig: remove whitespace

src/weathergen/datasets/utils.py
```diff
@@ -694,48 +695,54 @@ def compute_offsets_scatter_embed_target_srclk(batch: StreamData) -> StreamData:
...
+        ]
+    )
 
     # precompute index sets for scatter operation after embed
     offsets_base = target_srclk_tokens_lens.sum(1).sum(0).cumsum(1)
-    #take offset_base up to last col and append a 0 in the beginning per fstep
-    zeros_col = torch.zeros((offsets_base.shape[0], 1), dtype=offsets_base.dtype, device=offsets_base.device)
-    offsets = torch.cat([zeros_col, offsets_base[:,:-1]], dim=1)
+    # take offset_base up to last col and append a 0 in the beginning per fstep
+    zeros_col = torch.zeros(
```
clessig, line 720: Can you expand on the comment? It's not clear to me why this is necessary

src/weathergen/model/model.py
```diff
@@ -557,15 +557,14 @@ def forward(self, model_params: ModelParams, batch, forecast_offset: int, foreca
                         model_params, tokens_targets_srclk[fstep], source_cell_lens
                     )
                     tokens_target = self.assimilate_global(model_params, tokens_target)
-                    tokens_target_det = tokens_target.detach() # explicitly detach as well
+                    tokens_target_det = tokens_target.detach()  # explicitly detach as well
```
clessig: Remove whitespace

src/weathergen/model/model.py
```diff
@@ -649,7 +648,9 @@ def embed_cells_targets_srclk(self, model_params: ModelParams, streams_data) ->
                 ]
             )
             offsets_base = target_srclk_tokens_lens.sum(1).sum(0).cumsum(1)
-            num_fsteps = target_srclk_tokens_lens.shape[2]  # TODO: KCT, if there are diff no of tokens per fstep, this may fail
+            num_fsteps = target_srclk_tokens_lens.shape[
+                2
+            ]  # TODO: KCT, if there are diff no of tokens per fstep, this may fail
```
clessig: Can we handle this special case if things might break. When would it be triggered.

src/weathergen/run_evaluate.py
```diff
@@ -0,0 +1,192 @@
+# (C) Copyright 2025 WeatherGenerator contributors.
```
clessig: Why all these changes here?

clessig (commented):
Thanks for the contribution. Maybe best to have a quick call to go through the questions.

src/weathergen/datasets/multi_stream_data_sampler.py
kctezcan: That was a reminder for me, all good, I removed it.

src/weathergen/datasets/tokenizer_forecast.py
kctezcan: unfortunately ruff is adding that whitespace back :)

src/weathergen/datasets/tokenizer_masking.py
kctezcan: unfortunately ruff is adding that whitespace back :)

src/weathergen/datasets/utils.py
kctezcan: again ruff...

src/weathergen/datasets/utils.py
kctezcan: I rephrased it to: 
```python
    # shift the offsets for each fstep by one to the right, add a zero to the beginning the first token starts at 0
```

src/weathergen/model/model.py
kctezcan: ruff :(

src/weathergen/model/model.py
kctezcan: Actually this is OK, it does not break, I checked.

src/weathergen/run_evaluate.py
kctezcan: ah, i need this to start the code in debugging mode in VScode. It slipped into a commit, untracked it again.

src/weathergen/datasets/utils.py
kctezcan: Not sure what your question is here

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -367,6 +369,7 @@ def __iter__(self):
 
                             if rdata.is_empty():
                                 stream_data.add_empty_target(fstep)
+                                stream_data.add_empty_target_srclk(fstep)
```
sophie-xhonneux: please change the name to `add_empty_target_source_like` or something more readable as a variable/function name

src/weathergen/datasets/stream_data.py
```diff
@@ -318,6 +389,37 @@ def merge_inputs(self) -> None:
             self.source_tokens_cells = torch.tensor([])
             self.source_centroids = torch.tensor([])
 
+        # >>>>>>>
```
sophie-xhonneux: remove

src/weathergen/model/model.py
```diff
@@ -600,6 +620,73 @@ def embed_cells(self, model_params: ModelParams, streams_data) -> torch.Tensor:
 
         return tokens_all
 
+    def embed_cells_targets_srclk(self, model_params: ModelParams, streams_data) -> torch.Tensor:
```
sophie-xhonneux: why is it necessary to duplicate the function `embed_cells`? I think ideally we avoid that, because each them the embedding engine changes this function also needs to change, ie it is quite prone to code rot

kctezcan: This is one of the discussion points for me as well. if we have a function that we can use both for sources and targets, then we could also use it n times for each fstep as well, i.e. taking the loop over fsteps out of the funciton, as you and Christian have suggested earlier.

As the code is written at this point, this function does not take a variable inside but accesses the variables directly from the `streams_data` object. So my suggestion would be:
1. rewrite the function to take not the object as input but the related variables
2. call it in a for loop over fsteps for the targets.


I will implement it so we can see how it looks and decide what is better.

src/weathergen/datasets/multi_stream_data_sampler.py
kctezcan: "source_like" sounds good. It is a bit long, but des not bother me, if @clessig you are also ok, I can change it.

src/weathergen/model/model.py
sophie-xhonneux: Maybe ideally we can do that change in a separate PR? that way we can merge this if it looks good independently of the latent loss, e.g. faster ?

src/weathergen/datasets/multi_stream_data_sampler.py
clessig: Yes, `srclk` wasn't clear to me. Rather long and explicit.

src/weathergen/model/model.py
clessig: Yes, this should go into a separate PR.

src/weathergen/model/model.py
```diff
@@ -600,6 +622,76 @@ def embed_cells(self, model_params: ModelParams, streams_data) -> torch.Tensor:
...
+                        dtype=self.dtype,
+                        device="cuda",
+                    )
+                )
+
+            tokens_all_scattered = []
+            for _, sb in enumerate(streams_data):
+                for _, (s, embed) in enumerate(zip(sb, self.embeds, strict=False)):
+                    for fstep in range(num_fsteps):
+                        if s.target_source_like_tokens_lens[fstep].sum() != 0:
```
kctezcan, line 675: @clessig @sophie-xhonneux 
what do you think? should we skip empty fsteps or return an empty tensor for those?

clessig: Empty tensor

kctezcan: implemented

clessig:
Merged in revised form.

PR #1017, opened 2025-09-30
New templates for issues

### LINKED ISSUE #1002: Simplify the issue templates
We have a number of issue templates that we do not really use for the time being:
- many tasks have the useless 'enhancement' label
- the bug form is longer than necessary
- we do not need maintenance, user support forms

Based on how we have used issues in the past, I would suggest 3 templates:

- bug report: with a subset of the current options, everything optional.
- task: a blank GH issue with pointers on how to write a good task. This is for work meant to take less than a week.
- initiative: longer piece of work. It would have the initiative label. Maybe also a field for link to a hedgedoc, since this seems popular.

A consequence would be to remove 'enhancement' as a label. 90% of our work is enhancing

tjhunter:
## Description

Creates just 3 issues:
- bug (simpler than current form)
- task / issue: small piece of work
- initiative: big piece

Adds a few fields such as links to hedgedoc etc. all optional.

## Issue Number

Closes #1002

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,32 @@
+name: Bug Report
+description: Report a bug related to WeatherGenerator.
+labels: ["bug"]
+body:
+  - type: textarea
+    id: what-happened
+    attributes:
+      label: What happened?
+      description: |
+        Actual Results: What happened when you got the bug?
+        Expected Results: What was supposed to happen?
+        Additional Information: Add anything else you feel we need to know e.g. The exact error message printed
+    validations:
+      required: true
+  - type: textarea
+    id: reproduce
+    attributes:
+      label: What are the steps to reproduce the bug?
+      description: |
+        Minimal steps to reproduce the behavior:
+        - code branch with ALL configuration files
+        - HPC
+        - run_id
+        - command run
+        - if applicable: node configuration
+    validations:
+      required: "false"
+  - type: input
+    id: data
+    attributes:
+      label: Hedgedoc link to logs and more information. This ticket is public, do not attach files directly.
+      description: Please put all relevant information (logs, plots, etc.) in the Hedgedoc and link it here.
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,32 @@
+name: Bug Report
+description: Report a bug related to WeatherGenerator.
+labels: ["bug"]
+body:
+  - type: textarea
+    id: what-happened
+    attributes:
+      label: What happened?
+      description: |
+        Actual Results: What happened when you got the bug?
```
clessig: I would switch l10 and 11. For "actual results" I would also talk about "behaviour" and not "results" and "What happened when you got the bug?" -> "What happened instead?|

.github/ISSUE_TEMPLATE/initiative.yml
```diff
@@ -0,0 +1,52 @@
+name: Initiative
+description: A piece of work that will likely take more than a week to complete.
+title: "[ISSUE] "
+labels: ["initiative"]
+
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Describe the task. It can be a feature, documentation, etc.
```
clessig: "It can be a feature, documentation, etc." -> "It can be a feature, a set of experiments, documentation, etc."

.github/ISSUE_TEMPLATE/initiative.yml
```diff
@@ -0,0 +1,52 @@
...
+
+
+  - type: checkboxes
+    id: area
+    attributes:
+      label: Area
+      description: The general area this task relates to.
+      options:
+        - label: datasets, data readers, data preparation and transfer
+        - label: model
```
clessig, line 44: - label: science (e.g. experiments)

.github/ISSUE_TEMPLATE/task.yml
```diff
@@ -0,0 +1,37 @@
...
+      required: false
+
+  - type: checkboxes
+    id: area
+    attributes:
+      label: Area
+      description: The general area this task relates to.
+      options:
+        - label: datasets, data readers, data preparation and transfer
+        - label: model
```
clessig, line 29: We could again add science here although I have less strong feeling here than for initative

Issue #1036, opened 2025-10-06
evaluation crashes with empty CERRA samples


iluise:
### What happened?

One application reported the following problem with CERRA output:
```
 all_metrics, points_per_sample = calc_scores_per_stream(
                                     ^^^^^^^^^^^^^^^^^^^^^^^
  File "WeatherGenerator/packages/evaluate/src/weathergen/evaluate/utils.py", line 151, in calc_scores_per_stream
    metric_stream.loc[{"forecast_step": int(fstep)}] = combined_metrics
...
    raise IndexError(
IndexError: dimension coordinate 'sample' conflicts between indexed and indexing objects:
<xarray.DataArray 'sample' (sample: 1)> Size: 8B
array([0])
Coordinates:
  * sample         (sample) int64 8B 0
    stream         <U5 20B 'CERRA'
    forecast_step  int64 8B 2
vs.
<xarray.IndexVariable 'sample' (sample: 1)> Size: 8B
array([1])
```

### What are the steps to reproduce the bug?

run on develop branch at JSC with run id `t7tdo54f`

### Version

develop

### Platform (OS and architecture)

JSC 

### Relevant log output

```shell

```

PR #1039, opened 2025-10-06
[evaluation] fix score computation with empty cerra samples

iluise:
## Description
Cerra has empty samples and the scores metrics are not computed for those samples so for each forecast step we have e.g. sample 0 _or_ sample 1. When trying to allocate the values the xarray sees different samples for the same index and the code crashes. 

## Issue Number

closes https://github.com/ecmwf/WeatherGenerator/issues/1036

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/utils.py
+++ b/packages/evaluate/src/weathergen/evaluate/utils.py
@@ -140,13 +140,18 @@ def calc_scores_per_stream(
                 f"No data available for stream {stream} at forecast step {fstep} in region {region}. Skipping metrics calculation."
             )
             continue
-        
+
         metric_list.append(combined_metrics)
 
-        assert int(combined_metrics.forecast_step) == int(fstep), "Different steps in data and metrics. Please check."
+        assert int(combined_metrics.forecast_step) == int(fstep), (
+            "Different steps in data and metrics. Please check."
+        )
 
         metric_stream.loc[
-            {"forecast_step": int(combined_metrics.forecast_step), "sample": combined_metrics.sample}
+            {
+                "forecast_step": int(combined_metrics.forecast_step),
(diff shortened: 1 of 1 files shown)


packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -147,8 +142,9 @@ def calc_scores_per_stream(
             continue
 
         metric_list.append(combined_metrics)
-
-        metric_stream.loc[{"forecast_step": int(fstep)}] = combined_metrics
+        metric_stream.loc[
+            {"forecast_step": int(fstep), "sample": combined_metrics.sample}
```
clessig: This is the actual fix? The changes above seem to be just reformatting?

iluise: yes correct

SavvasMel: I am just wondering, if it's about making the `metric_stream ` "fit" the `combined_metrics`, wouldn't it be better to have:

`metric_stream.loc[{"forecast_step": combined_metrics.forecast_step, "sample": combined_metrics.sample}]`

I know that my comment is not changing anything - it might be assumed unimportant- but it might be more intuitive this way?

SavvasMel: Other than it works fine for me too, I checked for ERA5 and CERRA by having breakpoints.

iluise: Hi Savvas, I added your suggestions. is it good to go for you?

SavvasMel: Yes, definitely actually is very good to have assertion there, those 2 should be equal, and we can remove it in the future if nobody gets an error.

PR #1042, opened 2025-10-06
pyproject.toml checks

### LINKED ISSUE #883: Synchronize configuration of the sub-packages
We have multiple `pyproject.toml` files, one for the package and one for each sub-package. Each of them should have their own set of dependencies, but we need to ensure they are harmonized on some core aspects:

- project.version/authors/urs
- build-system
- tool.hatch/ruff/
- tool.uv ? 

The goal of this task is to write a small script that checks that certain sections of `pyproject.toml` files are all in sync. We cannot import one from another: the developers of `uv` have explicitly rejected this use case.

If we do not do that, each sub-package will evolve its own set of rules and versions, making it harder to develop and release all the project in lockstep and eventually causing incompatibility between the subpackages. One of our basic rules is to use the same development practices across all parts of the code.

enssow:
## Description

Adds a check into `actions.sh` so developers can check that the pyproject toml files are consistent across subpacakges and main pacakge

CHanges:
- added `toml-check` to actions.sh and corresponding `check_tomls.py` script
- have disabled lint check for print for the script so the results can be read from the terminal
- **Open to suggestions on alternative ways to run the test e.g. in intergration or as a pytest, I chose to do it like this as it is less invasive and can be run by those specifically maintaining the package repos**

## Issue Number

Closes #883 

      - **I have fixed formatting issues with `./scripts/actions.sh lint`** have disabled lint check for print for the script so the results can be read from the terminal

### DIFF
--- a/scripts/check_tomls.py
+++ b/scripts/check_tomls.py
@@ -1,4 +1,6 @@
 #!/usr/bin/env -S uv run
+
+# ruff: noqa: T201
 """
 Checks that all pyproject.toml files are consistent for select sections
 USAGE EXAMPLE: ./scripts/actions.sh toml-check from the root of the repo
@@ -24,7 +26,7 @@ def check_toml_key(main_toml_dict, other_toml_dict, list_keys, name):
             f"{list_keys} mismatch with main pyproject.toml and {name} pyproject.toml"
         )
     except Exception as e:
-        if type(e) == KeyError:
+        if type(e) is KeyError:
             print(
                 f"""KeyError: '{list_keys}' not found in {name} pyproject.toml, 
                 please populate this field"""
--- a/scripts/check_tomls.py
+++ b/scripts/check_tomls.py
@@ -7,50 +7,56 @@ USAGE EXAMPLE: ./scripts/actions.sh toml-check from the root of the repo
 import tomllib
 from pathlib import Path
 
-_REPO_ROOT = Path(
-    __file__
-).parent.parent
+_REPO_ROOT = Path(__file__).parent.parent
+
 
 def loop_keys(toml_dict, list_keys):
     for i in list_keys:
         toml_dict = toml_dict[i]
     return toml_dict
 
+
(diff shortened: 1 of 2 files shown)


scripts/check_tomls.py
```diff
@@ -0,0 +1,68 @@
...
+        with open(toml, "rb") as toml_file:
+            toml_dict = tomllib.load(toml_file)
+        all_tomls[str(toml)] = toml_dict
+    for name, toml_dict in all_tomls.items():
+        # shorten name to package path
+        name = name.split("/")[-2]
+        # check build system is the same
+        check_toml_key(main_toml_dict, toml_dict, ["build-system"], name)
+        # check python version is the same
+        check_toml_key(main_toml_dict, toml_dict, ["requires-python"], name)
```
tjhunter: this goes along with `version` etc.

scripts/check_tomls.py
```diff
@@ -0,0 +1,68 @@
...
+        # check project.version/authors/urls are the same
+        for key in ["version", "authors", "urls"]:
```
tjhunter: let's remove authors and urls, they are different right now

scripts/check_tomls.py
```diff
@@ -0,0 +1,68 @@
...
+        main_toml_dict = tomllib.load(toml_file)
+    all_tomls = {}
+    for toml in tomls:
+        toml_dict = {}
+        with open(toml, "rb") as toml_file:
+            toml_dict = tomllib.load(toml_file)
+        all_tomls[str(toml)] = toml_dict
+    for name, toml_dict in all_tomls.items():
+        # shorten name to package path
+        name = name.split("/")[-2]
```
grassesi: you can use pathlib here for more reliable path manipulation:
1. store path instead of str in `all_tomls`
2. `name = name.parent.name` (a but nonsensical, but with a bit of renaming it gets better) => `name = toml_path.parent.name`

enssow: sounds good - thank you!

Issue #911, opened 2025-09-16
Enable FesomDataReader to have different source and target datasets


kacpnowak:
### Is your feature request related to a problem? Please describe.

Currently all data readers use the same dataset for source and target. This limitation makes fine-tuning for some downstream application impossible. 

### Describe the solution you'd like

For a start, just FesomDataReader will use two filenames if they are provided, first as source and second as target 

### Organisation

AWI

PR #1046, opened 2025-10-06
Enable FesomDataReader to have different source and target datasets

kacpnowak:
## Description

This PR implements separate source and target datasets for FesomDataReader. 

It was also necessary modify masking strategies, as healpix cell can empty for the source but not for the target. Additionally there can be different number of tokens per cell. Different adjustments had to be made for each strategy:

- random: when mask is missing for each token, it's assumed to be False. 
- healpix: when mask is missing it's for the entire healpix cell it's assumed to be False. 
- casual: If mask is missing for an entire cell, it's assumed to be True for every token inside. If number of tokens in cell is different for source and target, source masked ratio is calculated and used for masking the same fraction of target tokens. 
- channels: unsupported 

Huge thanks for @shmh40 for all the help. 

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/911

### DIFF
--- a/.github/ISSUE_TEMPLATE/initiative.yml
+++ b/.github/ISSUE_TEMPLATE/initiative.yml
@@ -1,6 +1,6 @@
 name: Initiative
 description: A piece of work that will likely take more than a week to complete.
-title: "[ISSUE] "
+title: ""
 labels: ["initiative"]
 
 body:
--- a/.github/ISSUE_TEMPLATE/task.yml
+++ b/.github/ISSUE_TEMPLATE/task.yml
@@ -1,6 +1,6 @@
 name: Task / Issue
 description: A task or issue that should take less than a week to complete.
-title: "[ISSUE] "
+title: ""
 
 body:
   - type: textarea
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -133,6 +133,7 @@ grad_clip: 1.0
 weight_decay: 0.1
 norm_type: "LayerNorm"
 nn_module: "te"
+log_grad_norms: False
 
 start_date: 197901010000
 end_date: 202012310000
--- a/config/evaluate/config_zarr2cf.yaml
+++ b/config/evaluate/config_zarr2cf.yaml
@@ -98,6 +98,7 @@ coordinates:
         stream: stream
         forecast_step: forecast_period
         forecast_ref_time: forecast_ref_time
+        ncells: ncells
     pl:
         pressure_level: pressure
         valid_time: valid_time
@@ -106,6 +107,8 @@ coordinates:
         stream: stream
         forecast_step: forecast_period
         forecast_ref_time: forecast_ref_time
+        ncells: ncells
+
 dimensions:
     valid_time:
       wg: valid_time
@@ -131,6 +134,9 @@ dimensions:
     stream:
       wg: stream
       std: stream
+    ncells:
+      wg: ncells
+      std: ncells
     #TODO maybe absorb stream as an attribute
     # forecast_step: forecast_step
     # don't want forecast_step anyway
--- a/integration_tests/small1_test.py
+++ b/integration_tests/small1_test.py
@@ -172,8 +172,9 @@ def assert_train_loss_below_threshold(run_id):
     )
     # Check that the loss does not explode in a single epoch
     # This is meant to be a quick test, not a convergence test
-    assert loss_metric < 1.25, (
-        f"'stream.ERA5.loss_mse.loss_avg' is {loss_metric}, expected to be below 0.25"
+    target = 1.5
+    assert loss_metric < target, (
+        f"'stream.ERA5.loss_mse.loss_avg' is {loss_metric}, expected to be below {target}"
     )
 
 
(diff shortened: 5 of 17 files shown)


config/streams/fesom/fesom.yml
```diff
@@ -10,6 +10,7 @@
 FESOM_NODE :
   type : fesom
   filenames : ['ocean_node']
```
shmh40, line 12: Separately, I think the paths in WG-private need to be updated for FESOM data

src/weathergen/datasets/masking.py
```diff
@@ -293,11 +297,28 @@ def mask_target(
 
             elif self.current_strategy == "causal":
                 # select only the target times where mask is True
-                selected_tensors = [c for i, c in enumerate(cc) if pp[i]]
-
+                if len(cc) == len(pp):
```
shmh40, line 300: Do we want this if statement for the other strategies to be safer?

src/weathergen/datasets/masking.py
```diff
@@ -293,11 +297,28 @@ def mask_target(
...
+                elif len(pp) == 0:
+                    selected_tensors = cc
+                else:  # If length of target and mask doesn't match, create new mask
+                    ratio = np.sum(cc) / len(pp)  # Ratio of masked tokens in source
+                    indx = max(1, int(ratio * len(cc)))  # Get the same for target
+                    selected_tensors = cc[:indx]
+
+            elif self.current_strategy == "healpix":
+                selected_tensors = (
+                    cc if len(pp) > 0 and pp[0] else []
```
shmh40, line 311: I need to double check this

src/weathergen/datasets/masking.py
```diff
@@ -293,11 +297,28 @@ def mask_target(
...
+                )  # All tokens inside healpix cell have the same mask
+
+            elif self.current_strategy == "random":
+                # For random masking, we simply select the tensors where the mask is True.
+                # When there's no mask it's assumed to be False. This is done via strict=False
+                selected_tensors = [c for c, p in zip(cc, pp, strict=False) if p]
```
shmh40, line 317: In the case where cc is longer that pp this just **drops** the "extra tokens" right? Do we want instead to set them to be true/select them, so they are part of the target?

src/weathergen/datasets/masking.py
```diff
@@ -293,11 +297,28 @@ def mask_target(
...
             elif self.current_strategy == "causal":
                 # select only the target times where mask is True
-                selected_tensors = [c for i, c in enumerate(cc) if pp[i]]
-
+                if len(cc) == len(pp):
+                    selected_tensors = [c for i, c in enumerate(cc) if pp[i]]
+                elif len(pp) == 0:
+                    selected_tensors = cc
+                else:  # If length of target and mask doesn't match, create new mask
+                    ratio = np.sum(cc) / len(pp)  # Ratio of masked tokens in source
```
shmh40: Shouldn't this be np.sum(pp) or similar? You are trying to look at the ratio of masking in the source right? I am not exactly sure what the right code should be here

shmh40 (commented):
Thank you for the PR @kacpnowak! Added some comments. It runs for me for random, healpix for the fesom data but it doesn't work for causal, np.sum(cc) I don't think is right here.

Will may try to test again properly including for existing data that we want to make sure still works safely. There are some plans to rewrite some of this code for upcoming work, so we will try to see how it fits in.

src/weathergen/datasets/masking.py
kacpnowak: That's right, they are dropped. I'm not sure what's the best choice, but dropping is definitely easier

src/weathergen/datasets/masking.py
kacpnowak: I'm not sure if this makes sens for others. In case of random the same code handles both cases

src/weathergen/datasets/masking.py
shmh40: There might be problems here if we have a token_size that is insufficient. If for example, token size is small for source, we generate many more tokens than 12288. Then in target, if we have an appropriate token size, then we will end up with 12288 tokens. Now, the mask of the source will not appropriately map on to the target in any spatially meaningful sense.

I suggest we open an issue documenting this, and in the meantime add a comment stating this problem, and that we want this done spatially, we should use healpix masking at level 5. What do you think?

src/weathergen/datasets/masking.py
shmh40: Looks good, thanks for your patience.

src/weathergen/datasets/masking.py
```diff
@@ -293,11 +297,28 @@ def mask_target(
...
                 # select only the target times where mask is True
-                selected_tensors = [c for i, c in enumerate(cc) if pp[i]]
-
+                if len(cc) == len(pp):
+                    selected_tensors = [c for i, c in enumerate(cc) if pp[i]]
+                elif len(pp) == 0:
+                    selected_tensors = cc
+                else:  # If length of target and mask doesn't match, create new mask
+                    ratio = np.sum(pp) / len(pp)  # Ratio of masked tokens in source
+                    indx = max(1, int(ratio * len(cc)))  # Get the same for target
```
shmh40, line 306: So if we have e.g. one timestep in source, so it is False, we will have indx of 1. Then if target has more timesteps, then we have selected_tensors cc[:1] so then the first element of the target data is the actual target? Then it is just first element of source is predicting first element of target? Don't you want cc[-indx:] for some semblance to causal masking? The timesteps might be all over the place anyway but that is at least somewhat better?

kacpnowak: Yes, I think it should be `cc[-indx:]`

Issue #1055, opened 2025-10-08
CERRA lat lon mismatch across forecast steps


iluise:
### What happened?

I was investigating CERRA and the values seem to change from one forecast step to another:
 
sort_lats = forecast_step 1
ref_lats = forecast_step 2
 
```
np.amin(sort_lats), np.amax(sort_lats), np.amin(sort_lons), np.amax(sort_lons)
20.292282, 75.346344, -57.89835,74.03632

np.amin(ref_lats), np.amax(ref_lats), np.amin(ref_lons), np.amax(ref_lons)
20.321579,75.34659,-58.02143,74.10508
```

### What are the steps to reproduce the bug?

run evaluation on develop for a run containing CERRA and check the values inside this function https://github.com/ecmwf/WeatherGenerator/blob/develop/packages/evaluate/src/weathergen/evaluate/utils.py#L29 

### Version

develop - run_id `ar40mckx`

### Platform (OS and architecture)

ATOS

### Relevant log output

```shell

```

vhertel:
assign me

vhertel:
On commit `94b035a8` with run_id `r8u5vxn9`, lat/lon match across forecast steps (checked full arrays):

```
float(preds_next.lat.min()), float(preds_next.lat.min()), float(preds_next.lon.min()), float(preds_next.lon.min())
20.3653564453125, 20.3653564453125, -57.41053009033203, -57.41053009033203

float(tars_next.lat.min()), float(tars_next.lat.min()), float(tars_next.lon.min()), float(tars_next.lon.min())
20.3653564453125, 20.3653564453125, -57.41053009033203, -57.41053009033203
```

vhertel:
The corrected output still shows the mismatch across forecast steps:

```
float(preds.lat.min()), float(preds.lat.max()), float(preds.lon.min()), float(preds.lon.max())
20.3653564453125, 75.30535125732422, -57.6883544921875, 73.25624084472656

float(preds_next.lat.min()), float(preds_next.lat.max()), float(preds_next.lon.min()), float(preds_next.lon.max())
20.3653564453125, 75.34278869628906, -57.41053009033203, 73.82913208007812
```

clessig:
> The corrected output still shows the mismatch across forecast steps:
> 
> ```
> float(preds.lat.min()), float(preds.lat.max()), float(preds.lon.min()), float(preds.lon.max())
[...]

PR #1057, opened 2025-10-08
Fix CERRA eval breaking with coord sorting in `froct`

### LINKED ISSUE #1056: CERRA eval breaks with coord sorting for froct
### What happened?

When sorting the coordinates to compute the `froct` metric for ERA5 (as fixed in #1007), the evaluation pipeline crashes when trying to sort CERRA. The reason is explaine in #1055, that is, the coordinates in CERRA change from time-step to time-step and thus cannot be alligned between subsequent time steps.

Run evaluation including the `froct` metric with CERRA data.

### Version

develop

All HPCs

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

ECMWF

MatKbauer:
## Description

Move the coordinate (lat/lon) sorting required for a clean `froct` calculation such that it is only executed when `froct` is calculated. This prevents crashes with CERRA data, which cannot be sorted by coords (see #1055).

## Issue Number

Closes #1056 

      - I have run the integration tests with `./scripts/actions.sh integration-test` -> did not run

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/score.py
+++ b/packages/evaluate/src/weathergen/evaluate/score.py
@@ -619,7 +619,10 @@ class Scores:
 
         return rmse
 
-    def sort_by_coords(self, da_to_sort, da_reference):
+    @staticmethod
+    def sort_by_coords(
+        da_to_sort: xr.DataArray, da_reference: xr.DataArray
+    ) -> xr.DataArray:
         """
         Sorts one xarray.DataArray's coordinate ordering to match a reference array using KDTree.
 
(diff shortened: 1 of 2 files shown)


packages/evaluate/src/weathergen/evaluate/score.py
```diff
@@ -618,6 +619,66 @@ def calc_rmse(
 
         return rmse
 
+    def sort_by_coords(self, da_to_sort, da_reference):
```
iluise: since we are here, can you put the types? `da_to_sort: xr.DataArray` and the return `-> xr.DataArray`

packages/evaluate/src/weathergen/evaluate/score.py
```diff
@@ -618,6 +619,66 @@ def calc_rmse(
...
+
+        # Find nearest neighbors for reference coordinates
+        ref_coords = np.column_stack((ref_lats, ref_lons))
+        dist, indices = tree.query(ref_coords, distance_upper_bound=1e-5)
+
+        # Check for unmatched coordinates
+        unmatched_mask = ~np.isfinite(dist)
+        if np.any(unmatched_mask):
+            n_unmatched = np.sum(unmatched_mask)
+            raise ValueError(
```
iluise: I was wondering, instead of crashing should we just add an error message and return `da_to_sort` and `calc_change_rate` as all nans as in case of line 704 when s1 is none?

PR #1060, opened 2025-10-08
[1059][eval] Fix eval crash of inference models from other HPCs

### LINKED ISSUE #1059: Crash when evaluating model from other HPC
### What happened?

When `uv run evaluate` with an inference model from another HPC (different from code execution) the code breaks while trying to read the model path from the inference config.

The reason is that the inference config stores the model path from the HPC where the inference has been executed. When the inference model is transitioned to another HPC, the model path does not fit the current HPC's directories.

Copy an inference model from HPC-B to HPC-A and run evaluate on HPC-A.

### Version

develop

All

### Relevant log output

```shell
File "/lus/h2resw01/hpcperm/ecm1575/repos/WeatherGenerator/.venv/bin/evaluate", line 10, in <module>
    sys.exit(evaluate())
             ^^^^^^^^^^
  File "/lus/h2resw01/hpcperm/ecm1575/repos/WeatherGenerator/packages/evaluate/src/weathergen/evaluate/run_evaluation.py", line 36, in evaluate
    evaluate_from_args(sys.argv[1:])
  File "/lus/h2resw01/hpcperm/ecm1575/repos/WeatherGenerator/packages/evaluate/src/weathergen/evaluate/run_evaluation.py", line 58, in evaluate_from_args
    evaluate_from_config(OmegaConf.load(config))
  File "/lus/h2resw01/hpcperm/ecm1575/repos/WeatherGenerator/packages/evaluate/src/weathergen/evaluate/run_evaluation.py", line 88, in evaluate_from_config
    reader = WeatherGenReader(run, run_id, private_paths)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lus/h2resw01/hpcperm/ecm1575/repos/WeatherGenerator/packages/evaluate/src/weathergen/evaluate/io_reader.py", line 320, in __init__
    f"Zarr file {self.fname_zarr} does not exist or is not a directory."
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: Zarr file /leonardo_scratch/large/userexternal/thunter0/weathergen/shared_work/results/d0ko4ej5/validation_epoch00000_rank0000.zarr does not exist or is not a directory.
```

### Accompanying data

_No response_

### Organisation

ECMWF

MatKbauer:
## Description

Instead of reading the model path from the inference config (which invalidates when the inference model has changed HPC), it is read from the private config of the current HPC.

## Issue Number

Closes #1059 

      - I have run the integration tests with `./scripts/actions.sh integration-test` -> Did not run

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/io_reader.py
+++ b/packages/evaluate/src/weathergen/evaluate/io_reader.py
@@ -16,7 +16,7 @@ import omegaconf as oc
 import xarray as xr
 from tqdm import tqdm
 
-from weathergen.common.config import _load_private_conf, load_config, load_model_config
+from weathergen.common.config import get_shared_wg_path, load_config, load_model_config
 from weathergen.common.io import ZarrIO
 from weathergen.evaluate.derived_channels import DeriveChannels
 from weathergen.evaluate.score_utils import RegionBoundingBox, to_list
@@ -67,7 +67,9 @@ class DataAvailability:
 
 
 class Reader:
-    def __init__(self, eval_cfg: dict, run_id: str, private_paths: dict | None = None):
+    def __init__(
+        self, eval_cfg: dict, run_id: str, private_paths: dict[str, str] | None = None
+    ):
         """
         Generic data reader class.
 
@@ -77,7 +79,7 @@ class Reader:
             config with plotting and evaluation options for that run id
         run_id : str
             run id of the model
-        private_paths: dict
+        private_paths: dict[srt, str]
             dictionary of private paths for the supported HPC
         """
         self.eval_cfg = eval_cfg
@@ -283,10 +285,7 @@ class WeatherGenReader(Reader):
         self.inference_cfg = self.get_inference_config()
 
         if not self.results_base_dir:
-            pcfg = _load_private_conf()
-            self.results_base_dir = (
-                Path(pcfg.get("path_shared_working_dir")) / "results"
-            )
+            self.results_base_dir = Path(get_shared_wg_path("results"))
             _logger.info(
                 f"Results directory obtained from private config: {self.results_base_dir}"
             )
--- a/packages/evaluate/src/weathergen/evaluate/io_reader.py
+++ b/packages/evaluate/src/weathergen/evaluate/io_reader.py
@@ -15,6 +15,7 @@ import numpy as np
 import omegaconf as oc
 import xarray as xr
 from tqdm import tqdm
+
 from weathergen.common.config import _load_private_conf, load_config, load_model_config
 from weathergen.common.io import ZarrIO
 from weathergen.evaluate.derived_channels import DeriveChannels
--- a/packages/evaluate/src/weathergen/evaluate/io_reader.py
+++ b/packages/evaluate/src/weathergen/evaluate/io_reader.py
@@ -15,8 +15,7 @@ import numpy as np
 import omegaconf as oc
 import xarray as xr
 from tqdm import tqdm
-
-from weathergen.common.config import load_config, load_model_config
+from weathergen.common.config import _load_private_conf, load_config, load_model_config
 from weathergen.common.io import ZarrIO
 from weathergen.evaluate.derived_channels import DeriveChannels
 from weathergen.evaluate.score_utils import RegionBoundingBox, to_list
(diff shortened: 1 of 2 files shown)


packages/evaluate/src/weathergen/evaluate/io_reader.py
```diff
@@ -283,9 +283,12 @@ def __init__(self, eval_cfg: dict, run_id: str, private_paths: dict | None = Non
         self.inference_cfg = self.get_inference_config()
 
         if not self.results_base_dir:
-            self.results_base_dir = Path(self.inference_cfg["run_path"])
+            pcfg = _load_private_conf()
```
tjhunter: we need to find the path `results` in quite a few places. Instead, could we have this logic in `config.py`? I would suggest something such as:

```py
def path_results() -> Path:
     pcfg = _load_private_conf()
     return Path(pcfg.get("path_shared_working_dir")) / "results"

or an alternative for all paths such as output, results, ...

def path_in_wg(local_path: str | Path) -> Path:
     pcfg = _load_private_conf()
     return Path(pcfg.get("path_shared_working_dir")) / local_path

```

packages/evaluate/src/weathergen/evaluate/io_reader.py
```diff
@@ -77,8 +77,8 @@ def __init__(self, eval_cfg: dict, run_id: str, private_paths: dict | None = Non
             config with plotting and evaluation options for that run id
         run_id : str
             run id of the model
-        private_paths: lists
-            list of private paths for the supported HPC
+        private_paths: dict
```
tjhunter: this is not helpful, it it a dict but of what? I guess it is `dict[str, str]` ? What matters most is above in `def __init__(self, eval_cfg: dict, run_id: str, private_paths: dict | None = None):` then it should be `def __init__(self, eval_cfg: dict, run_id: str, private_paths: dict[str, str] | None = None):`

tjhunter (approved):
I have a suggestion to not break the interface

packages/evaluate/src/weathergen/evaluate/io_reader.py
```diff
@@ -283,9 +282,12 @@ def __init__(self, eval_cfg: dict, run_id: str, private_paths: dict | None = Non
         self.inference_cfg = self.get_inference_config()
 
         if not self.results_base_dir:
-            self.results_base_dir = Path(self.inference_cfg["run_path"])
+            pcfg = _load_private_conf()
```
iluise: Hi Matze, thanks for this!
actually `self.inference_cfg` already contains `"path_shared_working_dir"` (I checked it with a print) so no need to re-load the private config:
```
if not self.results_base_dir:
       self.results_base_dir = Path(self.inference_cfg["run_path"])

       shared_base_dir = (
                Path(self.inference_cfg.get("path_shared_working_dir")) / "results"
            )
        if (self.results_base_dir != shared_base_dir) : 
              _logger.info("Detected a result directory from a different HPC. Changing it.")
              self.results_base_dir = shared_base_dir
        _logger.info(
                    f"Results directory obtained from config: {self.results_base_dir}"
                )
```
I tested it on a run that was done and evaluated on the same machine though. Can you quickly re-test it with your run?

packages/evaluate/src/weathergen/evaluate/io_reader.py
MatKbauer: I like the idea and opened an issue (#1061). According to @iluise's solution, we have `path_shared_working_dir` already available here and don't have to load it from the private conf.

Edit: Following my answer below, we have to load the private config here. I implement your solution now.

packages/evaluate/src/weathergen/evaluate/io_reader.py
MatKbauer: Aaah, great catch! I tested it, though, and found that `self.inference_cfg.get("path_shared_working_dir") / "results"` is the same as `self.inference_cfg["run_path"]` (both pointing to the other HPC's results dir). I have thus implemented @tjhunter's solution above.

PR #1157, opened 2025-10-27
LayerNorm after last skip connection

SavvasMel:
## Description

This PR changes the MLP of the last block of the forecast engine. Specifically, it includes a LayerNorm with scale and bias turned off after the last skip connection of the last block.

## Issue Number

This is a draft PR

### DIFF
--- a/src/weathergen/model/engines.py
+++ b/src/weathergen/model/engines.py
@@ -24,7 +24,7 @@ from weathergen.model.embeddings import (
     StreamEmbedLinear,
     StreamEmbedTransformer,
 )
-from weathergen.model.layers import MLP, LayerNormBlock
+from weathergen.model.layers import FEMLP, MLP
 from weathergen.model.utils import ActivationFactory
 from weathergen.utils.utils import get_dtype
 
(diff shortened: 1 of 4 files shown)


src/weathergen/model/engines.py
```diff
@@ -317,18 +317,31 @@ def create(self) -> torch.nn.ModuleList:
...
+                    self.fe_blocks.append(
+                        MLP(
+                            self.cf.ae_global_dim_embed,
+                            self.cf.ae_global_dim_embed,
+                            with_residual=True,
+                            dropout_rate=self.cf.fe_dropout_rate,
+                            norm_type=self.cf.norm_type,
+                            dim_aux=1,
+                            norm_eps=self.cf.mlp_norm_eps,
+                        )
```
MatKbauer, line 343: For the sake of having less redundant code, can we modify the MLP to receive an additional argument `with_residual_layer_norm=False` instead of introducing FEMLP? When calling MLP here, we can set `with_residual_layer_norm=(i + 1) == self.cf.fe_num_blocks` to add the residual layer norm in the last MLP layer.

src/weathergen/model/layers.py
```diff
@@ -93,3 +93,80 @@ def forward(self, *args):
...
+
+        for _ in range(num_layers - 2):
+            self.layers.append(torch.nn.Linear(dim_hidden, dim_hidden))
+            self.layers.append(nonlin())
+            self.layers.append(torch.nn.Dropout(p=dropout_rate))
+
+        self.layers.append(torch.nn.Linear(dim_hidden, dim_out))
+
+        # Add LayerNorm after skip connection if residuals are used
+        if self.with_residual:
```
MatKbauer, line 150: Move this to the MLP class and add a second condition to the if statement: `if self.with_residual and self.with_residual_layer_norm:`

src/weathergen/model/layers.py
```diff
@@ -93,3 +93,80 @@ def forward(self, *args):
...
+
+        if self.with_residual:
+            if x.shape[-1] == x_in.shape[-1]:
+                x = x_in + x
+            else:
+                assert x.shape[-1] % x_in.shape[-1] == 0
+                x = x + x_in.repeat([*[1 for _ in x.shape[:-1]], x.shape[-1] // x_in.shape[-1]])
+
+            # Apply LayerNorm to the residual connection
+            x = self.residual_norm(x)
```
MatKbauer, line 170: Move to MLP and add `if self.with_residual_layer_norm:`

MatKbauer (changes requested):
Thanks for sharing your code, Savvas. Let's please modify it slightly to minimize redundant lines. See my suggestions below.

src/weathergen/model/engines.py
clessig: Why would we need to modify the MLP? It can--and in my opinion definitely should--be implemented in the forecast engine. Where we can just have the LayerNorm as the last block.

SavvasMel: Can you check now? Is that better?

clessig:
@MatKbauer : This has been merged via the forecast branch and the PR can be closed?

PR #1151, opened 2025-10-27
Make gradient checkpointing configurable in the embed transformer

javak87:
## Description
In order to save memory and improve computational efficiency, and since gradient checkpointing is used in several parts of the code, having the ability to enable or disable gradient checkpointing in specific sections can provide flexibility to maximize performance while also conserving memory.

## Issue Number
Refs #1141 

## Checklist before asking for review

-   [x] I have performed a self-review of my code
-   [ ] My changes comply with basic sanity checks:
      - I have fixed formatting issues with `./scripts/actions.sh lint`
      - I have run unit tests with `./scripts/actions.sh unit-test`
      - I have documented my code and I have updated the docstrings.
      - I have added unit tests, if relevant
-   [ ] I have tried my changes with data and code:
      - I have run the integration tests with `./scripts/actions.sh integration-test`
      - (bigger changes) I have run a full training and I have written in the comment the run_id(s): `launch-slurm.py --time 60`
      - (bigger changes and experiments) I have shared a hegdedoc in the github issue with all the configurations and runs for this experiments
-   [ ] I have informed and aligned with people impacted by my change:
    - for config changes: the MatterMost channels and/or a design doc
    - for changes of dependencies: the MatterMost software development channel

## Comparing the baseline and current PR performance

When embed_gradient_checkpoint_mode is set to false, the performance and GPU memory peak are as follows:
#### default_config.yml:
`../WeatherGenerator-private/hpc/launch-slurm.py --time 60
`
embed_gradient_checkpoint_mode

#### mixed.yml:
`../WeatherGenerator-private/hpc/launch-slurm.py --time 60 --config ./config/mixed.yml
`

embed_gradient_checkpoint_mode_mixed_dataset

For those GPUs with memory more than `25 GiB`, it is recommended to set `embed_gradient_checkpoint_mode` to False

PR #1168, opened 2025-10-28
Test turning all grad checkpointings off

### LINKED ISSUE #1141: Balance Memory and Compute in Gradient Checkpointing

Gradient checkpointing helps save memory, but it also slows down performance. While the lower memory usage is useful, the extra recomputation reduces training efficiency. It would be helpful to have tunable options to choose which layers use gradient checkpointing, so we can balance memory savings and performance.

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation

javak87:
## Description
This is a merge branch related to the following PRs:
https://github.com/ecmwf/WeatherGenerator/pull/1151  
https://github.com/ecmwf/WeatherGenerator/pull/1152
https://github.com/ecmwf/WeatherGenerator/pull/1153
https://github.com/ecmwf/WeatherGenerator/pull/1155
https://github.com/ecmwf/WeatherGenerator/pull/1156

## Issue Number
Closes #1141 

## Comparing the baseline and current PR performance

When `pred_gradient_checkpoint_mode` is set to false, the performance and GPU memory peak are as follows:
#### default_config.yml:
`../WeatherGenerator-private/hpc/launch-slurm.py --time 60
`
all_checkpoint_off

#### mixed.yml:
`../WeatherGenerator-private/hpc/launch-slurm.py --time 60 --config ./config/mixed.yml
`

all_checkpoint_off_mixed

### DIFF
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -33,6 +33,7 @@ ae_global_with_qk_lnorm: True
 ae_global_att_dense_rate: 1.0
 ae_global_block_factor: 64
 ae_global_mlp_hidden_factor: 2
+assimilate_global_gradient_checkpoint_mode: False 
 
 decoder_type: PerceiverIOCoordConditioning # CrossAttentionAdaNormConditioning
 pred_adapter_kv: False
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -6,6 +6,7 @@ embed_centroids_local_coords: False
 embed_size_centroids: 0
 embed_unembed_mode: "block"
 embed_dropout_rate: 0.1
+embed_gradient_checkpoint_mode: False
 
 target_cell_local_prediction: True
 
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -17,6 +17,7 @@ ae_local_with_qk_lnorm: True
 
 ae_local_num_queries: 1
 ae_local_queries_per_cell: False
+ae_local_blocks_grdient_checkpoint_mode: False
 ae_adapter_num_heads: 16
 ae_adapter_embed: 128
 ae_adapter_with_qk_lnorm: True
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -39,6 +39,7 @@ pred_adapter_kv: False
 pred_self_attention: True
 pred_dyadic_dims: False
 pred_mlp_adaln: True
+pred_gradient_checkpoint_mode: False
 
 # number of steps offset applied to first target window; if set to zero and forecast_steps=0 then
 # one is training an auto-encoder
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -22,6 +22,7 @@ ae_adapter_embed: 128
 ae_adapter_with_qk_lnorm: True
 ae_adapter_with_residual: True
 ae_adapter_dropout_rate: 0.1
+ae_adapter_grdient_checkpoint_mode: False
 
 ae_global_dim_embed: 2048
 ae_global_num_blocks: 8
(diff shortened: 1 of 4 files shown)


config/default_config.yml
```diff
@@ -6,6 +6,7 @@ embed_centroids_local_coords: False
 embed_size_centroids: 0
 embed_unembed_mode: "block"
 embed_dropout_rate: 0.1
+embed_gradient_checkpoint_mode: False
```
tjhunter, line 9: Ok for leaving the flags next to the modules. I would not call it `_mode` though, unsure what it means in this context.

src/weathergen/model/embeddings.py
```diff
@@ -148,23 +150,47 @@ def __init__(
     def forward_channels(self, x_in, centroids):
         peh = positional_encoding_harmonic
 
-        # embed provided input data
-        x = peh(checkpoint(self.embed, x_in.transpose(-2, -1), use_reentrant=False))
+        if self.embed_gradient_checkpoint_mode:
```
tjhunter, line 153: This is not the right place for branching, because you are then forced to copy the same logic across each branch.

Here is what I would suggest: we write a small conditional checkpoint function in one of the utility files (the signature is based on looking at the pytorch checkpoint function):

```py
    def cond_checkpoint(
            enable_checkpoint: bool,
            function,
            *args,
            use_reentrant: Optional[bool] = None,
            context_fn: Callable[[], Tuple[ContextManager, ContextManager]] = noop_context_fn,
            determinism_check: str = _DEFAULT_DETERMINISM_MODE,
            debug: bool = False,
            **kwargs
    ):
        if enable_checkpoint:
            checkpoint(function, ...)
        else:
            function(*args, **kwargs)
```

and then, the only change required is to convert:

```py
x = peh(checkpoint(self.embed, x_in.transpose(-2, -1), use_reentrant=False))
```

into:

```py
x = peh(cond_checkpoint(self.embed_gradient_checkpoint_mode, self.embed, x_in.transpose(-2, -1), use_reentrant=False))
```

How does that sound?

tjhunter: This way, we can also build this rule in the linter: do not use checkpoint directly but use cond_checkpoint

src/weathergen/model/model.py
```diff
@@ -754,15 +758,24 @@ def assimilate_local(
...
-            for block in self.ae_adapter:
-                tokens_global_c = checkpoint(
-                    block,
-                    tokens_global_c,
-                    tokens_c,
-                    q_cells_lens_c,
-                    cell_lens_c,
-                    use_reentrant=False,
-                )
+            if self.cf.ae_adapter_grdient_checkpoint_mode:
```
tjhunter, line 761: always assume the parameter may be missing:

```py
self.cf.get("ae_adapter_grdient_checkpoint_mode", True)
```

tjhunter (changes requested):
@javak87  thanks for the assembly of this PR. I suggest a way to abstract all the checkpoint changes, happy to talk about it if unclear.

config/default_config.yml
javak87: I changed `_mode` to `_enabled`

src/weathergen/model/embeddings.py
javak87: I’m okay with this helper function. However, I should put it in `__init__` for every checkpoint, because having these conditional helper functions in `forward` is degrading performance.

javak87:
Because of multiple conflicts, I opened a new PR.
New PR #1564

PR #1190, opened 2025-10-31
Rename epoch to mini_epoch

### LINKED ISSUE #515: Use steps instead of epochs as a basic progress unit

The current codebase is organized around both epochs and steps, but there are a couple of issues: 
- cf.istep is injected in the config as an increment of the batch_size_per_gpu . Why batch_size_per_gpu and not just +1?
- epoch is defined in the literature as a full pass over the dataset, but we hardcode the data length as 4096
- the metrics pipeline does not log the steps
- restarting a run should pick up from the right step (verify that it is the case)

The goal of this task is:
- alignment on how we measure the training progress: steps or epochs? steps seem to be the default given the size of the dataset. Should steps be invariant to the number of GPUs + number of nodes? I think so, but this is not how the code is implemented
- implement these decisinos

_No response_

_No response_

### Additional context

_No response_

### Organisation

_No response_

TillHae:
## Description

I renamed epoch to mini_epoch and used chkpt for file names.
Epoch didn't represent what was happening correctly. We never passed over the entire dataset, but just 4096 samples. Therefore we decided to rename it to mini_epoch.

## Issue Number

Closes #515

partially addresses issue but doesn't fix it entirely

### DIFF
--- a/packages/common/src/weathergen/common/config.py
+++ b/packages/common/src/weathergen/common/config.py
@@ -61,7 +61,7 @@ def save(config: Config, mini_epoch: int | None):
     dirname = path_models / config.run_id
     dirname.mkdir(exist_ok=True, parents=True)
 
-    fname = _get_model_config_file_name(path_models, config.run_id, mini_epoch)
+    fname = _get_model_config_file_write_name(path_models, config.run_id, mini_epoch)
 
     json_str = json.dumps(OmegaConf.to_container(config))
     with fname.open("w") as f:
@@ -84,7 +84,7 @@ def load_model_config(run_id: str, mini_epoch: int | None, model_path: str | Non
                 config=pconf, attribute_name="model_path", fallback="models"
             )
         path = Path(model_path)
-        fname = _get_model_config_file_name(path, run_id, mini_epoch)
+        fname = _get_model_config_file_read_name(path, run_id, mini_epoch)
         assert fname.exists(), (
             "The fallback path to the model does not exist. Please provide a `model_path`.",
             fname,
@@ -100,7 +100,18 @@ def load_model_config(run_id: str, mini_epoch: int | None, model_path: str | Non
     return _apply_fixes(config)
 
 
-def _get_model_config_file_name(path: Path, run_id: str, mini_epoch: int | None):
+def _get_model_config_file_write_name(path: Path, run_id: str, mini_epoch: int | None):
+    if mini_epoch is None:
+        mini_epoch_str = ""
+    elif mini_epoch == -1:
+        mini_epoch_str = "_latest"
+    else:
+        mini_epoch_str = f"_chkpt{mini_epoch:05d}"
+
+    return path / run_id / f"model_{run_id}{mini_epoch_str}.json"
+
+
+def _get_model_config_file_read_name(path: Path, run_id: str, mini_epoch: int | None):
     if mini_epoch is None:
         mini_epoch_str = ""
     elif mini_epoch == -1:


packages/common/src/weathergen/common/config.py
```diff
@@ -100,22 +100,22 @@ def load_model_config(run_id: str, epoch: int | None, model_path: str | None) ->
...
 
-def get_model_results(run_id: str, epoch: int, rank: int) -> Path:
+def get_model_results(run_id: str, mini_epoch: int, rank: int) -> Path:
     """
-    Get the path to the model results zarr store from a given run_id and epoch.
+    Get the path to the model results zarr store from a given run_id and mini_epoch.
     """
     run_results = Path(_load_private_conf(None)["path_shared_working_dir"]) / f"results/{run_id}"
-    zarr_path = run_results / f"validation_epoch{epoch:05d}_rank{rank:04d}.zarr"
+    zarr_path = run_results / f"validation_chkpt{mini_epoch:05d}_rank{rank:04d}.zarr"
```
MatKbauer: Can we have this backwards-compatible? Something like (not tested)
```python
zarr_path_old = run_results / f"validation_epoch{mini_epoch:05d}_rank{rank:04d}.zarr"
zarr_path_new = run_results / f"validation_chkpt{mini_epoch:05d}_rank{rank:04d}.zarr"
zarr_path = zarr_path_new if zarr_path_new.exists() else zarr_path_old
```

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -380,7 +380,7 @@ def retrieve_metric_from_json(reader: Reader, stream: str, region: str, metric:
     """
     score_path = (
         Path(reader.metrics_dir)
-        / f"{reader.run_id}_{stream}_{region}_{metric}_epoch{reader.epoch:05d}.json"
+        / f"{reader.run_id}_{stream}_{region}_{metric}_chkpt{reader.mini_epoch:05d}.json"
```
MatKbauer: Backward compatibility

packages/evaluate/src/weathergen/evaluate/io_reader.py
```diff
@@ -334,7 +334,7 @@ def __init__(self, eval_cfg: dict, run_id: str, private_paths: dict | None = Non
         )
 
         self.fname_zarr = self.results_dir.joinpath(
-            f"validation_epoch{self.epoch:05d}_rank{self.rank:04d}.zarr"
+            f"validation_chkpt{self.mini_epoch:05d}_rank{self.rank:04d}.zarr"
```
MatKbauer, line 502: Backward compatibility

packages/common/src/weathergen/common/config.py
```diff
@@ -456,9 +458,9 @@ def get_path_model(config: Config) -> Path:
     return Path(config.model_path) / config.run_id
 
 
-def get_path_output(config: Config, epoch: int) -> Path:
+def get_path_output(config: Config, mini_epoch: int) -> Path:
     base_path = get_path_run(config)
-    fname = f"validation_epoch{epoch:05d}_rank{config.rank:04d}.zarr"
+    fname = f"validation_chkpt{mini_epoch:05d}_rank{config.rank:04d}.zarr"
```
MatKbauer, line 494: Backward compatibility?

src/weathergen/model/model.py
```diff
@@ -549,16 +549,18 @@ def rename_old_state_dict(self, params: dict) -> dict:
...
+            mini_epoch : The mini_epoch to load. Default (-1) is the latest mini_epoch
         """
 
         path_run = Path(self.cf.model_path) / run_id
-        epoch_id = f"epoch{epoch:05d}" if epoch != -1 and epoch is not None else "latest"
-        filename = f"{run_id}_{epoch_id}.chkpt"
+        mini_epoch_id = (
+            f"chkpt{mini_epoch:05d}" if mini_epoch != -1 and mini_epoch is not None else "latest"
+        )
+        filename = f"{run_id}_{mini_epoch_id}.chkpt"
```
MatKbauer, line 563: Backward compatibility

src/weathergen/train/trainer.py
```diff
@@ -745,16 +749,18 @@ def batch_to_device(self, batch):
...
+            mini_epoch : The mini_epoch to load. Default (-1) is the latest mini_epoch
         """
 
         path_run = Path(self.cf.model_path) / run_id
-        epoch_id = f"epoch{epoch:05d}" if epoch != -1 and epoch is not None else "latest"
-        filename = f"{run_id}_{epoch_id}.chkpt"
+        mini_epoch_id = (
+            f"chkpt{mini_epoch:05d}" if mini_epoch != -1 and mini_epoch is not None else "latest"
+        )
+        filename = f"{run_id}_{mini_epoch_id}.chkpt"
```
MatKbauer, line 771: Backward compatibility

MatKbauer (commented):
Looks good, thanks Till for taking this on. I'd appreciate if we (temporarily) could make the code backward compatible, such that old models can still be loaded. @clessig, do you think backward compatibility would be reasonable here? I have left comments at code-parts where we load `.json` or `.ckpts` files. Hope I cought all occurances.

If we agree against backward compatibility, I will go forward and test the code. Have only inspected it so far, not tested.

packages/common/src/weathergen/common/config.py
grassesi: Can you check if this method is still in use? Ideally the knowledge of the specifics of the path should be encapsulated in `get_output_path` and anywhere this information is required `get_output_path` should be used instead. A simple backward compatibility mechanism such as @MatKbauer describes can be then implemented there.

clessig:
> I am inclined (weak yes) to break compatibility with this PR:
> 
> * we are changing inputs (config) and outputs (checkpoints)
> * we are still at a stage when it is easy to do, later stages will be harder/more painful
> 
> For the outputs, rather than the mini-epoch index, could we store the value of the mini-epoch? This way, we are more flexible in the future for having different increments such as every 500 optimizer steps etc. rather than being epoch based. Sorry if this was discussed in the design doc.

I don't think it's difficult to ensure backward compatability if we introduce a function that abstracts the checkpoint file name, and similar if we have a problem elsewhere. For the CI, we can just switch to the new name and properly publize it.

packages/common/src/weathergen/common/config.py
TillHae: The method is still in use in `evaluate/export`. I think `get_path_output` is only use for writing files, while `get_model_results` is only used for reading files

packages/evaluate/src/weathergen/evaluate/io_reader.py
```diff
@@ -498,9 +498,17 @@ def __init__(self, eval_cfg: dict, run_id: str, private_paths: dict | None = Non
...
         )
+        fname_zarr_old = self.results_dir.joinpath(
+            f"validation_epoch{self.mini_epoch:05d}_rank{self.rank:04d}.zarr"
+        )
+
+        if fname_zarr_new.exists():
+            self.fname_zarr = fname_zarr_new
+
+        self.fname_zarr = fname_zarr_old
```
MatKbauer: Is this one missing an else? That is
```python
if fname_zarr_new.exists():
    fname_zarr = fname_zarr_new
else:
    fname_zarr = fname_zarr_old
```

packages/common/src/weathergen/common/config.py
MatKbauer, line 1: For backward compatibility (evaluating models with old config), we need to add the following to the bottom of the `load_config()` function:
```python
    assert isinstance(c, Config)
    
    # Ensure the config has mini-epoch notation
    if hasattr(c, "samples_per_epoch"):
        c.samples_per_mini_epoch = c.samples_per_epoch
        c.num_mini_epochs = c.num_epochs

    return c
```

packages/common/src/weathergen/common/config.py
```diff
@@ -100,24 +100,38 @@ def load_model_config(run_id: str, epoch: int | None, model_path: str | None) ->
...
+    if mini_epoch is None:
+        mini_epoch_str = ""
+    elif mini_epoch == -1:
+        mini_epoch_str = "_latest"
+    elif (path / run_id / f"model_{run_id}_chkpt{mini_epoch:05d}.json").exists():
+        mini_epoch_str = f"_chkpt{mini_epoch:05d}"
     else:
-        epoch_str = f"_epoch{epoch:05d}"
-    return f"model_{run_id}{epoch_str}.json"
+        mini_epoch_str = f"_epoch{mini_epoch:05d}"
```
MatKbauer, line 120: The `.chkpt` files are named correctly, e.g., `a7ngybjp_chkpt000001.chkpt`. However, the `.json` files still have the `epoch` in their name.

In my new run directory, I had `model_a7ngybjp_epoch00001.json` but it should be `model_a7ngybjp_chkpt00001.json` instead, right?

TillHae: I think the problem is that `_get_model_config_file_name` is used for reading files as well as creating files. I think this should be fixed by just putting `epoch` in the `elif` check and `chkpt` in the `else`.
But in the case of both versions existing and trying to read one file, it would lead to the `epoch` being read instead of the new `chkpt` one.
We could fix this by using two different functions. One for reading and one for writing.

MatKbauer: I see, thanks for identifying the problem, Till. It's not very pretty, but let's create two separate functions for reading and writing, then. This will only be required for the transition period and we should revert back in two months or so to using a single function.

PR #1279, opened 2025-11-17
[1276][model] Implement noise conditioning for the diffusion model

### LINKED ISSUE #1276: Implement noise embedding and conditioning in the forecasting engine

Following the EDM structure, the diffusion denoiser needs to have an additional layer that computes scale and offset parameters as an encoding of the noise level applied at the current stage of the denoiser. As the EDM codebase does not implement this for transformer based architectures, we peek into the GenCast implementation [here](https://github.com/google-deepmind/graphcast/blob/c0376564b94581c973cd085df7c515237b77cc6b/graphcast/sparse_transformer.py#L411C1-L477C13). This prescribes that the norm conditioning layer is added before the attention and before the feed-forward operation. To stay close to the EDM codebase however, we choose the same embedding as chosen in the EBM codebase. We may replace this embedding method by the one used in Gencast.

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation

moritzhauschulz:
## Description
Incorporate the noise conditioning (aka time step embedding) into our setup of the diffusion model and the forecasting engine. We introduce new layers into the local and global attention blocks, as well as the MLP. We also enable the forecasting engine forward method to take the noise embedding and pass it to those blocks. All other changes are in the `diffusion.py`. 

Note that it was proposed by @clessig to align this with the code from the [DiT paper](https://github.com/facebookresearch/DiT/blob/main/models.py#L27C1-L64C21). This is taken into consideration in the most recent version, though the code combines elements from DiT, GenCast and EDM.

Some things still missing to undraft this PR:
- need to update functions that freeze the additional layers (e.g. for encoding of the noise)
- [DONE] update noise conditioning to align with [[DiT](https://github.com/facebookresearch/DiT/tree/main)](https://github.com/facebookresearch/DiT/tree/main) conditioning
    - see in particular the [timestep embedding class](https://github.com/facebookresearch/DiT/blob/ed81ce2229091fd4ecc9a223645f95cf379d582b/models.py#L27C1-L64C21) from DiT
- code currently draws on DiT, EDM and GenCast. We must check that all dimensionalities are correct, which may not be easily seen unless the code is run...
- Add copyright notices to DiT, GenCast as appropriate

## Issue Number

Closes #1276

### DIFF
--- a/src/weathergen/model/attention.py
+++ b/src/weathergen/model/attention.py
@@ -197,7 +197,7 @@ class MultiSelfAttentionHeadLocal(torch.nn.Module):
         dim_aux=None,
         norm_eps=1e-5,
         attention_dtype=torch.bfloat16,
-        with_noise_conditioning=False, #should only be True for diffusion model
+        with_noise_conditioning=False,  # should only be True for diffusion model
     ):
         super(MultiSelfAttentionHeadLocal, self).__init__()
 
@@ -244,7 +244,7 @@ class MultiSelfAttentionHeadLocal(torch.nn.Module):
         self.flex_attention = torch.compile(flex_attention, dynamic=False)
 
         if with_noise_conditioning:
-            self.noise_conditioning_q = LinearNormConditioning(dim_embed, dtype=self.dtype)
+            self.noise_conditioning = LinearNormConditioning(dim_embed, dtype=self.dtype)
 
 
     def forward(self, x, ada_ln_aux=None, emb=None):
@@ -255,7 +255,7 @@ class MultiSelfAttentionHeadLocal(torch.nn.Module):
         #NOTE: this is currently based on GenCast, not on DiT
         if self.noise_conditioning:
             assert emb is not None, "Need noise embedding if using noise conditioning"
-            x = self.noise_conditioning(x, emb)
+            x, gate = self.noise_conditioning(x, emb)
 
         # project onto heads
         s = [x.shape[0], x.shape[1], self.num_heads, -1]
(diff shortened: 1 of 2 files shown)


src/weathergen/model/engines.py
```diff
@@ -336,6 +336,7 @@ def __init__(self, cf: Config, num_healpix_cells: int) -> None:
                             dim_aux=1,
                             norm_eps=self.cf.norm_eps,
                             attention_dtype=get_dtype(self.cf.attention_dtype),
+                            with_noise_conditioning=self.cf.fe_diffusion
```
MatKbauer, line 339: We probably will not have a particular flag in the config to specify the use of the diffusion model. We will have to take care how to parameterize the engines correctly when intending to use diffusion. Let's keep this in mind.

MatKbauer:
> Thanks @MatKbauer. Not sure if I am understanding correctly. This conditioning is not related to conditioning on a previous state/label/etc. We will need this even in the most basic version as far as I am aware. Regardless, I am happy to proceed with the remaining bullet points (mainly functionality for freezing/unfreezing the new layers).

Aaah, you're right, @moritzhauschulz, we need the noise conditioning already for our very first experiments. I got confused with the conditioning types. Thanks for clarifying.

Can you pick me up concerning the freezing? I do not see where we depend on freezing. Say, we start with training a "simple" encoder/decoder (without forecast engine). Subsequently, we can freeze the pre-trained enc/dec modules, using something like `'freeze_modules=".*global.*|.*local.*|.*adapter.*|.*ERA5.*"'`, and train the latent diffusion model.

Or would you like to support the freezing of particular diffusion model components? That would be good to have but not urgent, as far as I can see.

PR #1293, opened 2025-11-18
Improve efficiency of score calculations when more than one regions are used

### LINKED ISSUE #659: Increase data processing efficiency in regional evaluation

Currently, the function `get_data` is called separately when evalutaion for several regions is requested.
Albeit the data is handled lazily with dask, this produces a computational over-head since especially the concatenation over samples requires time - it scales with the number of samples in the inference output.

Instead of lazyily loading the data repeatedly, the `get_data` should only be called once. Any subsequent slicing of the data to the domain of interest will then happen before feeding the metric-calculation. With this the region-loop moves from the `plot_inference.py`-script (see [here](https://github.com/ecmwf/WeatherGenerator/blob/0bc173f9c216695576c77f8a1f3551a9e36f6f9e/packages/evaluate/src/weathergen/evaluate/plot_inference.py#L80)) to the `calc_scores_per_stream`-method (see [here](https://github.com/ecmwf/WeatherGenerator/blob/0bc173f9c216695576c77f8a1f3551a9e36f6f9e/packages/evaluate/src/weathergen/evaluate/utils.py#L178)).

A more performant way would to avoid concatenation of xarray DataArrays, but more natively concatenate the dask-arrays in the zio-object. Likewise, regional filtering could be realized there.

### Additional context

N/A

### Organisation

JSC

SavvasMel:
## Description

This PR improves efficiency of score calculations when more than one regions are used. Previously for each region the data would be reloaded. Now the data are loaded only once. For more details see the relevant issue.

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/659#issuecomment-3348062283


packages/evaluate/src/weathergen/evaluate/io_reader.py
```diff
@@ -889,7 +876,7 @@ def load_scores(self, stream: str, region: str, metric: str) -> xr.DataArray | N
         """
         score_path = (
             Path(self.metrics_dir)
-            / f"{self.run_id}_{stream}_{region}_{metric}_epoch{self.epoch:05d}.json"
+            / f"{self.run_id}_{stream}_{region}_{metric}_epoch{self.mini_epoch:05d}.json"
```
SavvasMel: This one is a correction not related to this PR but it seems it is a wrong reference anyway.

SavvasMel: Actually, even my correction is not correct, it should be:

`f"{self.run_id}_{stream}_{region}_{metric}_chkpt{self.mini_epoch:05d}.json"`

iluise (approved):

Issue #1172, opened 2025-10-28
Add baseline tag for comparison in evaluation package


SavvasMel:
### What happened?

Currently, if someone wants to create score cards or bar plots to compare some models, the baseline is selected as the first `run_id ` in the config. This might be desirable, but not optimal. We would like to have the possibility for the user to select their baseline-run-id.

PR #1309, opened 2025-11-20
Add possibility for choosing a run_id as baseline for bar plots and score cards

SavvasMel:
## Description

Currently, if someone wants to create score cards or bar plots to compare some models, the baseline is selected as the first run_id  in the config. This might be desirable, but not optimal. We would like to have the possibility for the user to select their baseline-run-id. This possible by adding:

```
evaluation:
  metrics  : ["rmse","froct"] #, "mse"] # supported metrics: rmse, mae, l1, l2.. (see score.py)
  regions: ["global", "nhem"] 
  summary_plots : true
  summary_dir: "./plots/"
  print_summary: false #print out score values on screen. it can be verbose
  plot_ensemble: false #supported: false, "std", "minmax", "members"
  log_scale: false
  add_grid: false
  plot_score_maps: false
  score_cards: true
  bar_plots: true
  baseline: "o8b60tgh"    <---------------
```

The run_id `o8b60tgh ` should be one of the `run_ids ` in the evaluation config. Otherwise, the first `run_id ` in order will be taken as a baseline.

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/1172

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/plotter.py
+++ b/packages/evaluate/src/weathergen/evaluate/plotter.py
@@ -14,8 +14,8 @@ import xarray as xr
 from matplotlib.lines import Line2D
 from PIL import Image
 from scipy.stats import wilcoxon
-from weathergen.common.config import _load_private_conf
 
+from weathergen.common.config import _load_private_conf
 from weathergen.evaluate.plot_utils import (
     DefaultMarkerSize,
 )


packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -107,6 +106,7 @@ def evaluate_from_config(cfg, mlflow_client: MlflowClient | None) -> None:
 
     # to get a structure like: scores_dict[metric][region][stream][run_id] = plot
     scores_dict = defaultdict(lambda: defaultdict(lambda: defaultdict(dict)))
+    needs_to_be_computed = defaultdict(lambda: defaultdict(lambda: defaultdict(set)))
```
iluise: if I understand the logic, there's no need to have this global as it is used "per `run_id` and per 'stream'" anyway. 
I'd put it within the `stream` loop so it is just a simple dictionary and we avoid the double loop over streams.

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -165,23 +163,22 @@ def evaluate_from_config(cfg, mlflow_client: MlflowClient | None) -> None:
                                 forecast_step=available_data.fsteps,
                             )
 
-                    if metrics_to_compute:
-                        all_metrics, points_per_sample = calc_scores_per_stream(
-                            reader, stream, region, metrics_to_compute, plot_score_maps
-                        )
+        if run_id in needs_to_be_computed:
```
iluise: I'd rename it as "scores_to_be_computed" or smth similar. Also, if you move it within the `stream` loop you don't need this if statement and the second loop over `streams` right? you just need the lines below:

 missing_regions = list(scores_to_be_computed.keys())
missing_metrics = sorted({m for metrics in scores_to_be_computed.values() for m in metrics})

packages/evaluate/src/weathergen/evaluate/utils.py
```diff
@@ -112,88 +115,116 @@ def calc_scores_per_stream(
...
 
-        climatology = aligned_clim_data[fstep] if aligned_clim_data else None
-        score_data = VerifiedData(preds, tars, preds_next, tars_next, climatology)
-        # Build up computation graphs for all metrics
-        _logger.debug(f"Build computation graphs for metrics for stream {stream}...")
+            if region != "global":
+                _logger.debug(
+                    f"Applying bounding box mask for region '{region}' to targets and predictions."
+                )
+                tars = bbox.apply_mask(tars)
```
iluise: I was a bit confused a you are masking and overwriting tars and preds so I was afraid that the next step would have been calculated on the masked data from the step before, but I checked and it doesn't seem to be the case. 

minor, if you want you can use a more compact style: 
```
tars, preds, tars_next, preds_next = [
    bbox.apply_mask(x) if x is not None else None
    for x in (tars, preds, tars_next, preds_next)
]
```

src/weathergen/utils/validation_io.py
```diff
@@ -30,7 +30,7 @@ def write_output(
     sample_idxs,
 ):
     stream_names = [stream.name for stream in cf.streams]
-    analysis_streams_output = cf.get( 'analysis_streams_output', None)
+    analysis_streams_output = cf.get("analysis_streams_output", None)
```
iluise, line 33: can you restore the files that are not related to this PR?

SavvasMel: This is sub-branch from this PR https://github.com/ecmwf/WeatherGenerator/pull/1293, so I will just rebase when we merge the #1293.

iluise: oh sorry I added the review of the other PR here then. Should I copy the comments there or it's okay for you to have them here?

SavvasMel: It is totally ok. Let's finish with the other one, because this PR is few lines of changes but it is not clear now. Hopefully this will be clear when I will rebase.

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
SavvasMel: Done!

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
SavvasMel: Done!

packages/evaluate/src/weathergen/evaluate/utils.py
SavvasMel: Done!

PR #1336, opened 2025-11-21
[450] Integration test with multiple streams

### LINKED ISSUE #450: Integration test with multiple streams and obs

Currently, various code paths are not tested because the existing integration test uses a single source and only a gridded dataset

Integration test with multiple streams and a variety of data sources.

_No response_

### Additional context

_No response_

### Organisation

_No response_

simone99n:
## Description
Integration test with multiple streams:
- ERA5
- NPP-ATMS
- SYNOP

How to run: `./scripts/actions.sh integration-test-multi`
Test duration: ~8 minutes on Juwels-Booster

## Issue Number
* Closes #450 

### DIFF
--- a/integration_tests/small_multi_stream_test.py
+++ b/integration_tests/small_multi_stream_test.py
@@ -71,7 +71,7 @@ def infer_multi_stream(run_id):
     """Run inference for multi-stream model."""
     logger.info("run multi-stream inference")
     inference_from_args(
-        ["-start", "2022-10-10", "-end", "2022-10-11", "--samples", "10", "--mini_epoch", "0"]
+        ["-start", "2021-10-10", "-end", "2022-10-11", "--samples", "10", "--mini_epoch", "0"]
         + [
             "--from_run_id",
             run_id,


integration_tests/small_multi_stream.yaml
```diff
@@ -0,0 +1,21 @@
...
+loader_num_workers: 1
+
+forecast_offset: 0
+
+train_log:
+  log_interval: 1
+
+start_date: 201210010000
+# need to customize the starting date for NPPATMS
+# otherwise Integration Test can fails (nan values in the losses)
```
grassesi, line 35: Is this not rather an indication that something is wrong with the code? I assumed the WeatherGenerator data sampling should handle any combination of streams on any interval.

integration_tests/small_multi_stream_test.py
```diff
@@ -0,0 +1,227 @@
...
+            "ERA5": 2.0,
+            "NPPATMS": 2.0,
+            "SurfaceCombined": 2.0,
+        },
+        "val": {
+            "ERA5": 1.5,
+            "NPPATMS": 1.5,
+            "SurfaceCombined": 1.5,
+        },
+    }
```
grassesi, line 195: It would be great if you can make this dict a module level constant. That way a reader can identify (and manipulate) those critical values right away.

integration_tests/small_multi_stream_test.py
```diff
@@ -0,0 +1,227 @@
...
+    logger.info(f"{stage_label} losses – " + ", ".join(f"{k}: {v:.4f}" for k, v in losses.items()))
+
+
+def assert_train_losses_below_threshold(run_id):
+    assert_stream_losses_below_threshold(run_id, stage="train")
+
+
+def assert_val_losses_below_threshold(run_id):
+    assert_stream_losses_below_threshold(run_id, stage="val")
+
```
grassesi: I think it is easier to just call `assert_stream_losses_below_threshold` with the correct `stage` argument whereever it is required.

integration_tests/small_multi_stream_test.py
```diff
@@ -0,0 +1,227 @@
...
+            },
+        }
+    )
+    evaluate_from_config(cfg, None, None)
+
+
+def load_metrics(run_id):
+    """Helper function to load metrics"""
+    file_path = get_train_metrics_path(base_path=WEATHERGEN_HOME / "results", run_id=run_id)
+    if not os.path.exists(file_path):
```
grassesi: `get_train_metrics_path` returns a `pathlib.Path` so please use it here

integration_tests/small_multi_stream_test.py
```diff
@@ -0,0 +1,227 @@
...
+        raise FileNotFoundError(f"Metrics file not found for run_id: {run_id}")
+    with open(file_path) as f:
+        json_str = f.readlines()
+    return json.loads("[" + "".join([s.replace("\n", ",") for s in json_str])[:-1] + "]")
+
+
+def assert_metrics_file_exists(run_id):
+    """Test that the metrics file exists and can be loaded."""
+    file_path = get_train_metrics_path(base_path=WEATHERGEN_HOME / "results", run_id=run_id)
+    assert os.path.exists(file_path), f"Metrics file does not exist for run_id: {run_id}"
```
grassesi: `file_path.is_file()` (see above)

integration_tests/small_multi_stream.yaml
simone99n: Yes, it is.
This happens because of the few-sample training: if, by chance, the dataloader doesn’t pick any samples from after 2012, the NPP-ATMS losses still contain NaN values (and an error will arise during the threshold check).
Workaround: check whether the values are NaN and skip the threshold check for that specific stream — though in that case, we won’t be able to see whether the model is correctly learning the NPP-ATMS stream.

grassesi: can you open an issue for this?

clessig: > can you open an issue for this?

This should be handled at the level of the integration test.

scripts/actions.sh
```diff
@@ -101,6 +101,12 @@ case "$1" in
       uv run --offline pytest ./integration_tests/small1_test.py --verbose -s
     )
     ;;
+    integration-test-multi)
```
tjhunter: let's keep all the integration tests together, rather than having separate tasks. We should split it only when we have very long running tests.

scripts/actions.sh
```diff
@@ -155,7 +161,9 @@ case "$1" in
     )
     ;;
   *)
-    echo "Usage: $0 {sync|lint|lint-check|type-check|unit-test|toml-check|integration-test|create-links|create-jupyter-kernel|jupytext-sync}"
+    # Automatically extract all options from the case statement
+    options=$(grep -oP '^\s*\K[\w-]+(?=\))' "$0" | tr '\n' '|' | sed 's/|$//')
```
tjhunter, line 165: I cannot read it, but it seems to work.

integration_tests/small_multi_stream.yaml
tjhunter: The goal of this test right now is to check we can handle multiple streams. Let's keep this date for the time being and investigate later why nan values in npp-atms cause issue with the data reader.

integration_tests/small_multi_stream.yaml
```diff
@@ -0,0 +1,21 @@
...
+loss_fcts_val: [["mse", 1.0]]
+num_mini_epochs: 1
+samples_per_mini_epoch: 16
+samples_per_validation: 4
+lr_steps: 4
+lr_steps_warmup: 2
+lr_steps_cooldown: 2
+loader_num_workers: 1
+
+forecast_offset: 0
```
tjhunter: Would you mind changing the test setup to run the model in forecast mode? This code path is not tested yet, and it seems to be causing issues.

```yaml
forecast_offset : 1
forecast_steps: 2
forecast_policy: "fixed"
forecast_freeze_model: False
forecast_att_dense_rate: 1.0
fe_num_blocks: 2
fe_num_heads: 16
fe_dropout_rate: 0.1
fe_with_qk_lnorm: True
fe_layer_norm_after_blocks: [] 
impute_latent_noise_std: 0.0

healpix_level: 4
training_mode: "forecast"
```

tjhunter: Tell me if this does not work for you

grassesi:
Training Runns fine when rebased onto newest develop, however I get a `AssertionError: dataset size too small for forecast range` when starting the inference run.

grassesi:
> Training Runns fine when rebased onto newest develop, however I get a `AssertionError: dataset size too small for forecast range` when starting the inference run.
 
My branch was stale, runns fine :)

PR #1337, opened 2025-11-23
Issue1279 noise conditioning

### LINKED ISSUE #1276: Implement noise embedding and conditioning in the forecasting engine

moritzhauschulz:
## Description

Merge the noise embedding and conditioning functionality into the branch which collects all the diffusion PRs before merging into develop. Main new components are NoiseEmbedder and LinearNormConditioning. I tried to keep the changes as separate from existing code as possible, however some changes in the attention.py and engine.py were necessary to allow passing the noise embedding to the attention and MLP blocks. The implementation in large parts follows the DiT paper, but some adaptations have been made which are drawn form the EDM paper or GenCast. The code runs with the given default config, meaning that all shapes are aligned. The code has not been tested to ensure all new functionality works as intended. The code also runs when the fe_diffusion_model is set to False, but code should still be reviewed to ensure old functionality is not impacted. 

Copyright notices still to be added (can anyone point me to a resource on when it is necessary and how to do it?)

## Issue Number

Closes #1279 
Closes #1276 

-   [x] I have tried my changes with data and code
      **got Driver Not Loaded error on JuwelsBooster, but it worked with `uv run train`**

### DIFF
--- a/src/weathergen/model/attention.py
+++ b/src/weathergen/model/attention.py
@@ -13,8 +13,8 @@ import torch
 from flash_attn import flash_attn_func, flash_attn_varlen_func
 from torch.nn.attention.flex_attention import create_block_mask, flex_attention
 
-from weathergen.model.norms import AdaLayerNorm, RMSNorm
 from weathergen.model.layers import LinearNormConditioning
+from weathergen.model.norms import AdaLayerNorm, RMSNorm
 
 
 class MultiSelfAttentionHeadVarlen(torch.nn.Module):
@@ -248,10 +248,8 @@ class MultiSelfAttentionHeadLocal(torch.nn.Module):
         if with_noise_conditioning:
             self.noise_conditioning = LinearNormConditioning(dim_embed, dtype=self.dtype)
 
-
     def forward(self, *args):
-
-        # NOTE: Hotfix to accomodate TargetPredictionEngineClassic forward pass for both attention block and MLP...
+        # NOTE: Hotfix to accomodate TargetPredictionEngineClassic forward pass for attn. block, MLP...
         x = args[0]
         if len(args) == 2:
             ada_ln_aux = args[1]
@@ -549,16 +547,16 @@ class MultiSelfAttentionHead(torch.nn.Module):
         else:
             self.att = self.attention
             self.softmax = torch.nn.Softmax(dim=-1)
-            
+
         self.noise_conditioning = None
         if with_noise_conditioning:
-             #NOTE: noise_emb_dim currently hard-coded
-            self.noise_conditioning = LinearNormConditioning(latent_space_dim=dim_embed, noise_emb_dim=512, dtype=self.dtype)
-
+            # NOTE: noise_emb_dim currently hard-coded
+            self.noise_conditioning = LinearNormConditioning(
+                latent_space_dim=dim_embed, noise_emb_dim=512, dtype=self.dtype
+            )
 
     def forward(self, *args):
-
-        # NOTE: Hotfix to accomodate TargetPredictionEngineClassic forward pass for both attention block and MLP...
+        # NOTE: Hotfix to accomodate TargetPredictionEngineClassic forward pass for attn. block, MLP...
         x = args[0]
         if len(args) == 2:
             ada_ln_aux = args[1]
@@ -576,7 +574,6 @@ class MultiSelfAttentionHead(torch.nn.Module):
         if self.noise_conditioning:
             assert emb is not None, "Need noise embedding if using noise conditioning"
             x, gate = self.noise_conditioning(x, emb)
-
 
         # project onto heads and q,k,v and
         # ensure these are 4D tensors as required for flash attention
@@ -674,5 +671,3 @@ class MultiCrossAttentionHead(torch.nn.Module):
             outs = x_q_in + outs
 
         return outs
-
-
--- a/src/weathergen/model/attention.py
+++ b/src/weathergen/model/attention.py
@@ -14,7 +14,7 @@ from flash_attn import flash_attn_func, flash_attn_varlen_func
 from torch.nn.attention.flex_attention import create_block_mask, flex_attention
 
 from weathergen.model.norms import AdaLayerNorm, RMSNorm
-
+from weathergen.model.diffusion import LinearNormConditioning
 
 class MultiSelfAttentionHeadVarlen(torch.nn.Module):
     def __init__(
@@ -197,6 +197,7 @@ class MultiSelfAttentionHeadLocal(torch.nn.Module):
         dim_aux=None,
         norm_eps=1e-5,
         attention_dtype=torch.bfloat16,
+        with_noise_conditioning=False, #should only be True for diffusion model
     ):
         super(MultiSelfAttentionHeadLocal, self).__init__()
 
@@ -242,10 +243,19 @@ class MultiSelfAttentionHeadLocal(torch.nn.Module):
         # compile for efficiency
         self.flex_attention = torch.compile(flex_attention, dynamic=False)
 
-    def forward(self, x, ada_ln_aux=None):
+        if with_noise_conditioning:
+            self.noise_conditioning_q = LinearNormConditioning(dim_embed, dtype=self.dtype)
+
+
+    def forward(self, x, ada_ln_aux=None, emb=None):
         if self.with_residual:
             x_in = x
         x = self.lnorm(x) if ada_ln_aux is None else self.lnorm(x, ada_ln_aux)
+
+        #NOTE: this is currently based on GenCast, not on DiT
+        if self.noise_conditioning:
+            assert emb is not None, "Need noise embedding if using noise conditioning"
+            x = self.noise_conditioning(x, emb)
 
         # project onto heads
         s = [x.shape[0], x.shape[1], self.num_heads, -1]
@@ -320,11 +330,17 @@ class MultiCrossAttentionHeadVarlen(torch.nn.Module):
         self.dtype = attention_dtype
         assert with_flash, "Only flash attention supported at the moment"
 
-    def forward(self, x_q, x_kv, x_q_lens=None, x_kv_lens=None, ada_ln_aux=None):
+    def forward(self, x_q, x_kv, emb=None, x_q_lens=None, x_kv_lens=None, ada_ln_aux=None):
         if self.with_residual:
             x_q_in = x_q
         x_q = self.lnorm_in_q(x_q) if ada_ln_aux is None else self.lnorm_in_q(x_q, ada_ln_aux)
         x_kv = self.lnorm_in_kv(x_kv)
+
+        #should only apply to diffusion model
+        if self.noise_conditioning:
+            assert emb is not None, "Need noise embedding if using noise conditioning"
+            x_q = self.noise_conditioning_q(x_q, emb)
+            x_kv = self.noise_conditioning_kv(x_kv, emb)
 
         # project onto heads and q,k,v and
         # ensure these are 4D tensors as required for flash attention
@@ -487,6 +503,7 @@ class MultiSelfAttentionHead(torch.nn.Module):
         dim_aux=None,
         norm_eps=1e-5,
         attention_dtype=torch.bfloat16,
+        with_noise_conditioning=False, #should only be True for diffusion model
     ):
         super(MultiSelfAttentionHead, self).__init__()
 
@@ -526,11 +543,21 @@ class MultiSelfAttentionHead(torch.nn.Module):
         else:
             self.att = self.attention
             self.softmax = torch.nn.Softmax(dim=-1)
+            
+        if with_noise_conditioning:
+            self.noise_conditioning = LinearNormConditioning(dim_embed, dtype=self.dtype)
 
-    def forward(self, x, ada_ln_aux=None):
+
+    def forward(self, x, ada_ln_aux=None, emb=None):
         if self.with_residual:
             x_in = x
         x = self.lnorm(x) if ada_ln_aux is None else self.lnorm(x, ada_ln_aux)
+
+        #NOTE: this is currently based on GenCast, not on DiT
+        if self.noise_conditioning:
+            assert emb is not None, "Need noise embedding if using noise conditioning"
+            x = self.noise_conditioning(x, emb, dtype=self.dtype)
+
 
         # project onto heads and q,k,v and
         # ensure these are 4D tensors as required for flash attention
--- a/src/weathergen/model/attention.py
+++ b/src/weathergen/model/attention.py
@@ -14,7 +14,8 @@ from flash_attn import flash_attn_func, flash_attn_varlen_func
 from torch.nn.attention.flex_attention import create_block_mask, flex_attention
 
 from weathergen.model.norms import AdaLayerNorm, RMSNorm
-from weathergen.model.diffusion import LinearNormConditioning
+from weathergen.model.layers import LinearNormConditioning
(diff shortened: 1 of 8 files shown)


src/weathergen/model/diffusion.py
```diff
@@ -93,13 +99,13 @@ def forward(self, tokens: torch.Tensor, fstep: int) -> torch.Tensor:
         # noise = torch.randn(y.shape, device=y.device)  # now eta from MultiStreamDataSampler
         sigma = (eta * self.p_std + self.p_mean).exp()
         n = torch.randn_like(y) * sigma
-        return self.denoise(x=y + n, c=c, sigma=sigma)
+        return self.denoise(x=y + n, fstep=fstep, c=c, sigma=sigma)
```
Jubeku: Does the conditioning on the fstep make sense here? Our forecast engine certainly works with the fstep here (which can be used in the adaptive layer norm), but is it also the correct conditioning for denoising from a diffusion perspective? If yes, then it's great :)

src/weathergen/model/attention.py
```diff
@@ -242,11 +244,29 @@ def mask_block_local(batch, head, idx_q, idx_kv):
...
+            self.noise_conditioning = LinearNormConditioning(dim_embed, dtype=self.dtype)
+
+    def forward(self, *args):
+        # NOTE: Hotfix to accomodate TargetPredictionEngineClassic forward pass for attn. block, MLP...
+        x = args[0]
+        if len(args) == 2:
+            ada_ln_aux = args[1]
+        elif len(args) > 2:
+            ada_ln_aux = args[-1]
+            emb = args[1] if self.noise_conditioning else None
```
MatKbauer, line 258: I'd rather be specific here and use `elif len(args) == 3 and self.noise_conditioning` instead of `> 2`

src/weathergen/model/attention.py
```diff
@@ -527,11 +548,33 @@ def __init__(
...
+            self.noise_conditioning = LinearNormConditioning(
+                latent_space_dim=dim_embed, noise_emb_dim=512, dtype=self.dtype
+            )
+
+    def forward(self, *args):
+        # NOTE: Hotfix to accomodate TargetPredictionEngineClassic forward pass for attn. block, MLP...
+        x = args[0]
+        if len(args) == 2:
+            ada_ln_aux = args[1]
+        elif len(args) > 2:
```
MatKbauer, line 563: Specific check (see above)

src/weathergen/model/diffusion.py
MatKbauer: Let's preserve the ordering in the `denoise()` function and put `fstep` as last argument

src/weathergen/model/diffusion.py
```diff
@@ -111,13 +117,17 @@ def denoise(self, x: torch.Tensor, c: torch.Tensor, sigma: float) -> torch.Tenso
         c_in = 1 / (sigma**2 + self.sigma_data**2).sqrt()
         c_noise = sigma.log() / 4
 
+        # Embed noise level
+        emb = self.noise_embedder(c_noise)
```
MatKbauer: Let's call this var `noise_emb` instead of `emb`

src/weathergen/model/diffusion.py
```diff
@@ -111,13 +117,17 @@ def denoise(self, x: torch.Tensor, c: torch.Tensor, sigma: float) -> torch.Tenso
...
+
         # Precondition input and feed through network
         x = self.preconditioner.precondition(x, c)
-        return c_skip * x + c_out * self.net(c_in * x, c_noise)  # Eq. (7) in EDM paper
+        return c_skip * x + c_out * self.net(c_in * x, fstep=fstep, emb=emb)  # Eq. (7) in EDM paper
```
MatKbauer: `noise_emb` here as well :)

src/weathergen/model/diffusion.py
```diff
@@ -170,3 +180,44 @@ def __init__(self):
...
+        super().__init__()
+        self.dtype = dtype
+        self.mlp = torch.nn.Sequential(
+            torch.nn.Linear(frequency_embedding_dim, embedding_dim, bias=True),
+            torch.nn.SiLU(),
+            torch.nn.Linear(embedding_dim, embedding_dim, bias=True),
+        )
+        self.frequency_embedding_dim = frequency_embedding_dim
+
+    def timestep_embedding(self, t, max_period=10000):
```
MatKbauer: `t: float` and `max_perios: int=10000`

src/weathergen/model/diffusion.py
```diff
@@ -170,3 +180,44 @@ def __init__(self):
...
+        freqs = torch.exp(
+            -math.log(max_period) * torch.arange(start=0, end=half, dtype=self.dtype) / half
+        ).to(device=t.device)
+        args = t[:, None].float() * freqs[None]
+        embedding = torch.cat([torch.cos(args), torch.sin(args)], dim=-1)
+        if self.frequency_embedding_dim % 2:
+            embedding = torch.cat([embedding, torch.zeros_like(embedding[:, :1])], dim=-1)
+        return embedding
+
+    def forward(self, t):
```
MatKbauer: `t: float`

src/weathergen/model/engines.py
```diff
@@ -376,10 +379,17 @@ def init_weights_final(m):
         for block in self.fe_blocks:
             block.apply(init_weights_final)
 
-    def forward(self, tokens, fstep):
+    def forward(self, tokens, fstep, emb=None):
```
MatKbauer: `noise_emb` for specificity

src/weathergen/model/engines.py
```diff
@@ -376,10 +379,17 @@ def init_weights_final(m):
...
         aux_info = torch.tensor([fstep], dtype=torch.float32, device="cuda")
-        for block in self.fe_blocks:
-            tokens = checkpoint(block, tokens, aux_info, use_reentrant=False)
+        if self.cf.fe_diffusion_model:
+            assert emb is not None, (
+                "Noise embedding must be provided for diffusion forecasting engine"
```
MatKbauer: "forecast engine" instead of "forecasting engine". Sorry for being so picky

src/weathergen/model/layers.py
```diff
@@ -79,11 +86,20 @@ def __init__(
 
         self.layers.append(torch.nn.Linear(dim_hidden, dim_out))
 
+    # TODO: expanded args, must check dependencies (previously aux = args[-1])
     def forward(self, *args):
-        x, x_in, aux = args[0], args[0], args[-1]
+        x, x_in = args[0], args[0]
+        if len(args) == 2:
+            aux = args[1]
+        elif len(args) > 2:
```
MatKbauer, line 109: More specific check here again (see above)

src/weathergen/model/layers.py
```diff
@@ -79,11 +86,20 @@ def __init__(
...
+            aux = args[-1]
+            emb = args[1] if self.with_noise_conditioning else None
```
MatKbauer: Also `noise_emb` here

src/weathergen/model/layers.py
```diff
@@ -79,11 +86,20 @@ def __init__(
...
 
         for i, layer in enumerate(self.layers):
-            x = layer(x, aux) if (i == 0 and self.with_aux) else layer(x)
+            if isinstance(layer, LinearNormConditioning):
+                x = layer(x, emb)  # noise embedding
```
MatKbauer: and here

MatKbauer (commented):
From a visual inspection the code looks good. I have added some comments to rename variables for clarity and do more specific `elif` checks. I will test the code functionality tomorrow.

Concerning the copyright notes: Please add a comments similar to those in the `diffusion.py` class at the top to all classes that contain code from respective repositories. Moreover, we should add notes to the `NOTES` document in the main directory of the repo. Let me know if you need guidance on what to put there.

src/weathergen/model/attention.py
MatKbauer: Edit: @Jubeku and I had some thoughts about refactoring the `forward` functions. Instead of passing `*args`, let's use the same notation as with `ada_ln_aux` that we pass as `None` by default. The `forward()` function would then look like this:
```python
    def forward(
        self, 
        x: torch.Tensor, 
        ada_ln_aux: torch.Tensor = None, 
        noise_emb: torch.Tensor = None
    ) -> torch.Tensor:        
        if self.with_residual:
            x_in = x
        x = self.lnorm(x) if ada_ln_aux is None else self.lnorm(x, ada_ln_aux)

        if self.noise_conditioning:
            x, gate = self.noise_conditioning(x, noise_emb)

        # project onto heads
        s = [x.shape[0], x.shape[1], self.num_heads, -1]
        qs = self.lnorm_q(self.proj_heads_q(x).reshape(s)).to(self.dtype).permute([0, 2, 1, 3])
        ks = self.lnorm_k(self.proj_heads_k(x).reshape(s)).to(self.dtype).permute([0, 2, 1, 3])
        vs = self.proj_heads_v(x).reshape(s).permute([0, 2, 1, 3])

        outs = self.flex_attention(qs, ks, vs, block_mask=self.block_mask).transpose(1, 2)

        out = self.proj_out(self.dropout(outs.flatten(-2, -1)))
        if self.with_residual:
            out = x_in + out * gate if self.noise_conditioning else x_in + out

        return out
```
Can you please implement those changes or do you see issues with this notation?

src/weathergen/model/attention.py
MatKbauer: Edit: change to notation proposed above

src/weathergen/model/engines.py
```diff
@@ -376,10 +379,17 @@ def init_weights_final(m):
...
+            )
+            for block in self.fe_blocks:
+                tokens = checkpoint(block, tokens, emb, aux_info, use_reentrant=False)
+        else:
+            for block in self.fe_blocks:
+                tokens = checkpoint(block, tokens, aux_info, use_reentrant=False)
```
MatKbauer, line 392: With the default `noise_emb` introduced above, can we just pass `noise_emb` to the checkpoints and not depend on the case differentiation (remove the `if/else`)?

src/weathergen/model/layers.py
MatKbauer: Edit: Notation from above

MatKbauer (commented):
@Jubeku and I had some more thoughts on how to simplify the forward calls. The code runs and the loss does not explode. Next step is to approach the overfitting to a single sample.

src/weathergen/model/attention.py
moritzhauschulz: @MatKbauer Regarding copyright, is it okay to add the notice at the top of the file and then have #NOTEs referring to the papers? This is what I did for now.

moritzhauschulz: I could not find the `NOTES` document in the repo, do you have a link?

MatKbauer: Having the paper in the `NOTICE` document sounds good, yes! Sorry, the file name is [`NOTICE`](https://github.com/moritzhauschulz/WeatherGenerator/blob/issue1279_noise_conditioning/NOTICE) not `NOTES`

moritzhauschulz: > Edit: @Jubeku and I had some thoughts about refactoring the `forward` functions. Instead of passing `*args`, let's use the same notation as with `ada_ln_aux` that we pass as `None` by default. The `forward()` function would then look like this:
> 
> ```python
>     def forward(
>         self, 
>         x: torch.Tensor, 
>         ada_ln_aux: torch.Tensor = None, 
>         noise_emb: torch.Tensor = None
>     ) -> torch.Tensor:        
>         if self.with_residual:
>             x_in = x
>         x = self.lnorm(x) if ada_ln_aux is None else self.lnorm(x, ada_ln_aux)
> 
>         if self.noise_conditioning:
>             x, gate = self.noise_conditioning(x, noise_emb)
> 
>         # project onto heads
>         s = [x.shape[0], x.shape[1], self.num_heads, -1]
>         qs = self.lnorm_q(self.proj_heads_q(x).reshape(s)).to(self.dtype).permute([0, 2, 1, 3])
>         ks = self.lnorm_k(self.proj_heads_k(x).reshape(s)).to(self.dtype).permute([0, 2, 1, 3])
>         vs = self.proj_heads_v(x).reshape(s).permute([0, 2, 1, 3])
> 
>         outs = self.flex_attention(qs, ks, vs, block_mask=self.block_mask).transpose(1, 2)
> 
>         out = self.proj_out(self.dropout(outs.flatten(-2, -1)))
>         if self.with_residual:
>             out = x_in + out * gate if self.noise_conditioning else x_in + out
> 
>         return out
> ```
> 
> Can you please implement those changes or do you see issues with this notation?

@MatKbauer @Jubeku This is unfortunately a bit more complicated, or at least I could not find an easier solution. My solution is very ugly but making it prettier will require some broader refactoring. The issue is that forward is always called with the `checkpoint` function, e.g. [here](https://github.com/moritzhauschulz/WeatherGenerator/blob/10dfcdcaa2499e72df8ab671152bd36805933089/src/weathergen/model/engines.py#L388C13-L389C87). The checkpoint function does not allow passing arguments by name, so the order matters (or at least I haven't found out how). If all blocks were the same, and we could just pass noise as the last argument, then that would be fine, but currently the [MLP block](https://github.com/ecmwf/WeatherGenerator/blob/641f3ba6376e8c75c6f188b37585ebdbe495e634/src/weathergen/model/layers.py#L82C5-L84C1) uses `aux = args[-1]`. My first thought was to just change this to `aux = args[1]`, but then this causes problems elsewhere, as MLP is called also in other places. In particular, I ran into a problem [here](https://github.com/ecmwf/WeatherGenerator/blob/641f3ba6376e8c75c6f188b37585ebdbe495e634/src/weathergen/model/engines.py#L536), where `aux` needs to be passed as the last argument in the current set-up.

I fully agree that the current way is not ideal. Maybe the only way to get around this in the long run is to pass a `dict` to `checkpoint` and then each block unpacks it according to the input it needs. Happy to have a go at it but I thought it would be quite a big refactor to sneak into this PR. Let me know your thoughts – I hope I explained it well enough.

src/weathergen/model/engines.py
moritzhauschulz: No problem, but the class is also called ForecastingEngine. Am I missing something?

src/weathergen/model/attention.py
MatKbauer: Gotcha, thanks for explaining. Agree that we should leave as is for now -- your solution works well -- and we touch and clean it up later.

src/weathergen/model/engines.py
MatKbauer: True, I always referred to it as forecast engine for convenience. Let's also postpone this (optional) change to later.

Jubeku (approved):

PR #1363, opened 2025-11-26
Add support for using strings and floats for timedelta

### LINKED ISSUE #1360: Timesteps (len_hrs/step_hrs) smaller then 1 hour

Timesteps (len_hrs/step_hrs) are in hours in the config file. Convert so that it is also possible to enter timesteps in minutes or seconds.

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation

kacpnowak:
## Description

Add support for using float and string as step_hrs, len_hrs and forecast_delta_hrs config parameters. Their values can be smaller than 1h, but must be greater than 1ms.

Floats and ints are interpreted as hours. String must be compatible with pandas [`to_timedelta`](https://pandas.pydata.org/docs/reference/api/pandas.to_timedelta.html).

## Issue Number

closes https://github.com/ecmwf/WeatherGenerator/issues/1360


config/ifs_fesom_config.yml
```diff
@@ -17,4 +17,7 @@ samples_per_validation: 16
 shuffle: True
 loader_num_workers: 4
 
+len_hrs: "360 min"
+step_hrs: 6.0
```
clessig: We agreed that the supported time delta format is:

HH:MM:SS

so it should read:

```
len_hrs: 06:00:00
step_hrs: 06:00:00
```

I wouldn't have anything against a fallback that tries to parse other formats.

clessig (commented):
We should stick to the standard delta time format that we agreed on a while back.

We should also change the key names to:

```
time_window_len
time_window_step
```

but we can also do this in a separate PR.

config/ifs_fesom_config.yml
kacpnowak: Okay, I changed that. We rely on pandas for parsing so almost anything will work

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -100,20 +101,29 @@ def __init__(
...
-        )
-        assert self.forecast_delta_hrs == self.len_hrs, "Only supported option at the moment"
+
+        # Handle forecast_delta_hrs which might be int (hours) or string (timedelta)
+        f_delta_dt = parse_timedelta(cf.forecast_delta)
+
+        if f_delta_dt > np.timedelta64(0, "ms"):
+            self.forecast_delta_dt = f_delta_dt
+        else:
+            self.forecast_delta_dt = self.len_timedelta
```
grassesi, line 114: `forecast_delta_dt` is kind of redundant, if you simplify the name I would go for `forecast_dt`

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -100,20 +101,29 @@ def __init__(
...
             )
 
         self.forecast_offset = cf.forecast_offset
-        self.forecast_delta_hrs = (
-            cf.forecast_delta_hrs if cf.forecast_delta_hrs > 0 else self.len_hrs
-        )
-        assert self.forecast_delta_hrs == self.len_hrs, "Only supported option at the moment"
+
+        # Handle forecast_delta_hrs which might be int (hours) or string (timedelta)
+        f_delta_dt = parse_timedelta(cf.forecast_delta)
```
grassesi: this should be done before the config is used => during the parsing of the config

src/weathergen/datasets/data_reader_base.py
```diff
@@ -49,6 +49,17 @@ class TimeIndexRange:
...
+    """
+    Parse a value into a numpy timedelta64[ms].
+    Integers and floats are interpreted as hours.
+    Strings are parsed using pandas.to_timedelta.
+    """
+    if isinstance(val, int | float | np.number):
+        return np.timedelta64(pd.to_timedelta(val, unit="h")).astype("timedelta64[ms]")
+    return np.timedelta64(pd.to_timedelta(val)).astype("timedelta64[ms]")
+
+
```
grassesi: This parsing/conversion should be done during the creation of the config, so that there is no ambiguity with regard to the type/semantics of a config value. Therefor this function should also go to the `common.config` module.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -100,20 +101,29 @@ def __init__(
         self.mask_value = 0.0
         self._stage = stage
 
-        self.len_hrs: int = cf.len_hrs
-        self.step_hrs: int = cf.step_hrs
-        self.time_window_handler = TimeWindowHandler(start_date, end_date, cf.len_hrs, cf.step_hrs)
+        self.len_timedelta: np.timedelta64 = parse_timedelta(cf.time_window_len)
+        self.step_timedelta: np.timedelta64 = parse_timedelta(cf.time_window_step)
```
grassesi: Parse config values to a non-ambiguous type during creation of config.
Also I feel like `len_timedelta` and `step_timedelta` are not really good names. Since you are renaming these anyway, I would rather go for something like: `sampling_window_size` and `sampling_step_size` clearer :shrug:

config/default_config.yml
```diff
@@ -139,8 +139,8 @@ start_date: 197901010000
 end_date: 202012310000
 start_date_val: 202101010000
 end_date_val: 202201010000
-len_hrs: 6
-step_hrs: 6
+time_window_step: 06:00:00
+time_window_len: 06:00:00
```
grassesi, line 152: `len_hrs`, `step_hrs` are also used in `utils.validation_io` and will break. Please update the code there.

grassesi (changes requested):
I have some remarks:
- You have missed some references to `step_hrs` and `len_hrs`: in `data_reader_cams`, `data_reader_icon`, but most importantly in in `validation_io`
- With regard to the new variables names of `step_hrs` and `len_hrs`, I personally find them unintuitive, but that is just personal preference.
- the parsing to timedelta objects should definitely happen much earlier (during parsing of the config), since otherwise the sematics/type of these attributes are ambiguous (making it hard to understand/ leading to potential bugs). `parse_timedelta` should therefor go to `common.config`. Be aware you might need some additional logic to handle the (de)serialization of config objects (probably `_load_overwrite_conf`, `_load_default_conf`, `load_model_config`, `save`)

grassesi:
> Regarding data_reader_cams and data_reader_icon, they both have `step_hrs` as placeholders. I think changing that is beyond the scope of this PR

Yes, I just wanted to bring it up :+1:

kacpnowak:
It's ready for review. There's one big change: ints and floats are now interpreted as seconds.

YAML parser 24:00:00 matches the integer with pattern [1-9][0-9]*:[0-9]+.... The parser calculates 24*3600 + 0*60 + 0 = 86400 and stores it as an integer seconds. To work with this all floats and ints have to follow the same convention. However it shouldn't be a problem because one can just use "24h" instead. 

I ran tests with era5 and fesom datasets, but I'd be really grateful if you could run test in with something using obs data reader

packages/common/src/weathergen/common/config.py
```diff
@@ -27,12 +30,123 @@
...
+            conf[raw_key] = f"${{{key}}}"
+            conf[key] = f"${{{_TIMEDELTA_TYPE_NAME}:{conf[key]}}}"
+
+    for key in time_keys:
+        if key in conf:
+            raw_key = f"_{key}"
+            conf[raw_key] = f"${{{key}}}"
+            conf[key] = f"${{{_DATETIME_TYPE_NAME}:{conf[key]}}}"
+
+    return conf
```
grassesi, line 103: This produces an error for me when running the integration test:
```python
s = '1979-01-01T00:00:00.000000'

    def str_to_datetime64(s: str | int | np.datetime64) -> np.datetime64:
        """
        Convert a string to a numpy datetime64 object.
        """
        if isinstance(s, np.datetime64):
            return s
        s_str = str(s)
    
        supported_formats = [
            "%Y%m%d%H%M%S",
            "%Y-%m-%d %H:%M:%S",
            "%Y-%m-%d %H:%M",
            "%Y-%m-%dT%H:%M:%S",
            "%Y-%m-%dT%H:%M",
        ]
    
        for fmt in supported_formats:
            try:
                dt_obj = datetime.datetime.strptime(s_str, fmt)
                return np.datetime64(dt_obj)
            except ValueError:
                pass
    
>       raise ValueError(f"Unable to parse the date string '{s}'. Original string might be invalid.")
E       omegaconf.errors.InterpolationResolutionError: ValueError raised while resolving interpolation: Unable to parse the date string '1979-01-01T00:00:00.000000'. Original string might be invalid.
E           full_key: start_date
E           object_type=dict

packages/common/src/weathergen/common/config.py:92: InterpolationResolutionError

```

kacpnowak: I fixed that. Thank you for catching it

src/weathergen/datasets/data_reader_base.py
```diff
@@ -65,68 +63,17 @@ def __post_init__(self):
...
     def __init__(
         self,
-        t_start: str | int | NPDT64,
-        t_end: str | int | NPDT64,
-        t_window_len_hours: int,
-        t_window_step_hours: int,
+        t_start: NPDT64,
+        t_end: NPDT64,
+        t_window_len_hours: np.timedelta64,
+        t_window_step_hours: np.timedelta64,
```
grassesi: maybe use NPTDel64 here to be consistent?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -92,28 +92,38 @@ def __init__(
...
+        self.time_window_handler = TimeWindowHandler(
+            start_date, end_date, self.len_timedelta, self.step_timedelta
+        )
         if is_root():
             logger.info(
-                f"Time window handler: start={start_date}, end={end_date},"
-                f"len_hrs={cf.len_hrs}, step_hrs={cf.step_hrs}"
+                f"Time window handler: start={start_date}, end={end_date}, "
+                f"time_window_len={timedelta_to_str(cf.time_window_len)} "
+                f"time_window_step={timedelta_to_str(cf.time_window_step)}"
```
grassesi: Would be great if you implement this as the `__str__` method of TimeWindowHandler. Then you can just do `logger.info(self.time_window_handler)`

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -92,28 +92,38 @@ def __init__(
     ):
         super(MultiStreamDataSampler, self).__init__()
 
-        start_date = str_to_datetime64(start_date_)
-        end_date = str_to_datetime64(end_date_)
+        start_date = start_date_
+        end_date = end_date_
 
         assert end_date > start_date, (end_date, start_date)
```
grassesi: Can you do this check in the `TimeWindowHandler.__init__` since start/end time is really something the TimeWindowHandler handles.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -92,28 +92,38 @@ def __init__(
     ):
         super(MultiStreamDataSampler, self).__init__()
 
-        start_date = str_to_datetime64(start_date_)
-        end_date = str_to_datetime64(end_date_)
+        start_date = start_date_
+        end_date = end_date_
```
grassesi: You can drop those local variables if you implement my suggestions (see below), then we can also change the variable names `start_date_`/`end_date_` in the function signature to `start_date`/`end_date`, making it look much cleaner.

src/weathergen/datasets/data_reader_obs.py
grassesi, line 1: Unfortunately your change breaks the obervation datareader: The conversion of `np.datetime64` to `datetime.datetime` using `.item()` (see lines 155/156) is unstable and results in a conversion to int in your case, breaking lines 160/161 (`TypeError: unsupported operand type(s) for -: 'int' and 'datetime.datetime'`).
Via experimentation I found out the following rule:
- `np.datetime64('2012-10-01T00:00:00.000000').item()` (or less precise) => converted to `datetime.datetime`
- `np.datetime64('np.datetime64('2012-10-01T00:00:00.0000000').item()` (or more precise) => converted to `int`

grassesi (changes requested):

src/weathergen/datasets/data_reader_obs.py
grassesi: This can be tested using the new multistream integration test.

kacpnowak:
Hi @grassesi, thanks for the review and catching those errors. I think I fixed them however afer merging develop it seems to me that integration test fail too early for me to test data_reader_obs. I'd very much appreciate if you could test again 

```
__________________________________________________________ ERROR collecting integration_tests/small1_test.py ___________________________________________________________
integration_tests/small1_test.py:16: in <module>
    import omegaconf
.venv/lib/python3.12/site-packages/omegaconf/__init__.py:1: in <module>
    from .base import Container, DictKeyType, Node, SCMode, UnionNode
.venv/lib/python3.12/site-packages/omegaconf/base.py:11: in <module>
    from ._utils import (
.venv/lib/python3.12/site-packages/omegaconf/_utils.py:112: in <module>
    class OmegaConfDumper(yaml.Dumper):  # type: ignore
E   AttributeError: module 'yaml' has no attribute 'Dumper'
======================================================================= short test summary info ========================================================================
ERROR integration_tests/small1_test.py - AttributeError: module 'yaml' has no attribute 'Dumper'
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -82,21 +82,16 @@ class MultiStreamDataSampler(torch.utils.data.IterableDataset):
...
-        cf,
-        start_date_,
-        end_date_,
-        batch_size,
-        samples_per_mini_epoch,
+        cf: Config,
+        start_date: np.datetime64,
+        end_date: np.datetime64,
+        batch_size: int,
+        samples_per_mini_epoch: int,
```
grassesi, line 89: nice, thanks for the typehints :)

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -181,7 +182,9 @@ def __init__(
 
                 fsm = self.forecast_steps[0]
                 if len(ds) > 0:
-                    self.len = min(self.len, len(ds) - (self.len_hrs * (fsm + 1)) // self.step_hrs)
+                    self.len = min(
+                        self.len, len(ds) - (self.len_timedelta * (fsm + 1)) // self.step_timedelta
+                    )
```
grassesi, line 187: In MultiStreamDataSampler.reset a similiar quantitiy is computed using `self.forecast_delta_dt` instead of `self.len_timedelta`. For now both will be alway the same as far as I rembered, but we should keep it consistent as this assumption might break in the future.

grassesi (changes requested):
I can not reproduce the integration test failure that you got (On what system are you testing and how do you run your tests? I run `. scripts/actions integration-test` on a santis interactive session). But there is also a small reference you forgot to update when merging: In `MultiStreamDataSampler.__init__` could you change `forecast_len = (self.len_hrs * (fsm + 1)) // self.step_hrs` (l. 207) to  `forecast_len = (self.forecast_delta_dt * (fsm + 1)) // self.step_timedelta`.

grassesi (approved):
Everything works fine now, this is a very nice improvement

clessig (approved):
I didn't do extensive testing but it looks good. My only comment is that I think `def str_to_datetime64(s: str | int | NPDT64) -> NPDT64:` should be not in `data_reader_base.py` but in `datasets/utils.py` or in the common package.

grassesi:
> I didn't do extensive testing but it looks good. My only comment is that I think `def str_to_datetime64(s: str | int | NPDT64) -> NPDT64:` should be not in `data_reader_base.py` but in `datasets/utils.py` or in the common package.

It is already in the common package under `common.config.str_to_datetime`

PR #1450, opened 2025-12-11
Sorcha/dev/zarr3 compaction

### LINKED ISSUE #1384: zarr3 compaction

Investigation into the compaction capability of zarr3. Will implement [sharding](https://zarr-specs.readthedocs.io/en/latest/v3/codecs/sharding-indexed/) to compare the inode usage of the zarr storage created during an inference run.

https://gitlab.jsc.fz-juelich.de/hedgedoc/3X10vtdVQumQZy4qpBiHWg

### Area

- [ ] model
- [ ] science
- [ ] documentation

enssow:
## Description

Draft PR to showcase how to enable sharding (experimentation underway to optimise performance)

## Issue Number

**DRAFT**
Closes #1384 


enssow:
Latest version can read and write with the zipstore format but needs tidying up (produces lots of print statements to narrow down the issues when trying to run the evaluation from a .yml file)

enssow:
- Specify type of store to create during an inference with `uv run --offline inference --from_run_id zs581tqh --samples 1 --streams_output ERA5 --zarr-store local` or `uv run --offline inference --from_run_id zs581tqh --samples 1 --streams_output ERA5 --zarr-store zip`
- Code to run evaluation remains the same with no extra sections of config required (the type of store is detected using the extension) e.g. `uv run evaluate --config ../test_evaluate_zip.yml` (I tested the latest code with runs: `v6l27pog` (zip) and `bvu0y897` (local)
- Backwards Compatible: `/p/project1/weatherai/owens1/WeatherGenerator/.venv/lib/python3.12/site-packages/zarr/core/group.py:568: ZarrUserWarning: Both zarr.json (Zarr format 3) and .zgroup (Zarr format 2) metadata objects exist at file:///p/scratch/weatherai/shared_work/results/v8pqzh4y/validation_chkpt00000_rank0000.zarr. Zarr format 3 will be used.
  warnings.warn(msg, category=ZarrUserWarning, stacklevel=1)` get this warning but it does plot (tested with run `v8pqzh4y`)

enssow:
Integration test fails, need to investigate this further!

clessig:
> Latest version can read and write with the zipstore format but needs tidying up (produces lots of print statements to narrow down the issues when trying to run the evaluation from a .yml file)

What is the performance impact for running evaluation with using zipstore?

packages/common/src/weathergen/common/config.py
```diff
@@ -138,20 +130,21 @@ def get_model_results(run_id: str, mini_epoch: int, rank: int) -> Path:
...
-            f"Zarr file with run_id {run_id}, mini_epoch {mini_epoch} and rank {rank} does not "
-            f"exist or is not a directory."
-        )
+    for ext in ["zarr", "zip"]:
+        zarr_path_new = run_results / f"validation_chkpt{mini_epoch:05d}_rank{rank:04d}.{ext}"
+        zarr_path_old = run_results / f"validation_epoch{mini_epoch:05d}_rank{rank:04d}.{ext}"
 
-    return zarr_path
+        if zarr_path_new.exists() or zarr_path_new.is_dir():
+            zarr_path = zarr_path_new
```
tjhunter: return `zarr_path_new`?

packages/common/src/weathergen/common/config.py
```diff
@@ -138,20 +130,21 @@ def get_model_results(run_id: str, mini_epoch: int, rank: int) -> Path:
...
+            return zarr_path
+        elif zarr_path_old.exists() or zarr_path_old.is_dir():
+            zarr_path = zarr_path_old
```
tjhunter: same comment here

packages/common/src/weathergen/common/config.py
```diff
@@ -501,8 +491,13 @@ def get_path_model(config: Config) -> Path:
 
 
 def get_path_output(config: Config, mini_epoch: int) -> Path:
+    type = config.zarr_store
```
tjhunter: `type` is a reserved keyword => use `store_type`

packages/common/src/weathergen/common/config.py
```diff
@@ -501,8 +491,13 @@ def get_path_model(config: Config) -> Path:
...
+    if type == "zip":
```
tjhunter: this is the sort of logic to wrap in a small function (defensive programming principle):

```py
from typing import Literal

type StoreType = Literal["zip", "local"]

def _store_type(path: Path) -> StoreType:
    ext = path.suffix[1:]
    if ext == "zip":
        return "zip"
    elif ext == "zarr":
        return "local"
    raise ValueError(...)

```

packages/common/src/weathergen/common/io.py
```diff
@@ -247,12 +251,21 @@ def arrays(self) -> dict[str, ArrayType]:
...
         return np.arange(self.data.shape[0])
 
-    def as_xarray(self, chunk_nsamples=CHUNK_N_SAMPLES) -> xr.DataArray:
+    def as_xarray(
+        self, chunk_nsamples=CHUNK_N_SAMPLES, shard_nsamples=SHARD_N_SAMPLES
+    ) -> xr.DataArray:
         """Convert raw dask arrays into chunked dask-aware xarray dataset."""
-        chunks = (chunk_nsamples, *self.data.shape[1:])
-
+        chunks = (chunk_nsamples, *(max(x // 4, 1) for x in self.data.shape[1:]))
```
tjhunter: where is this formula coming from? Someone else should be able to update it later.

packages/common/src/weathergen/common/io.py
```diff
@@ -315,14 +328,50 @@ def _append_dataset(self, dataset: OutputDataset | None, name: str) -> None:
 class ZarrIO:
     """Manage zarr storage hierarchy."""
 
-    def __init__(self, store_path: pathlib.Path):
+    def __init__(self, store_path: pathlib.Path, create=False):
```
tjhunter: do you need the default? I would update the code everywhere to make it clear

packages/common/src/weathergen/common/io.py
```diff
@@ -315,14 +328,50 @@ def _append_dataset(self, dataset: OutputDataset | None, name: str) -> None:
...
         self._store_path = store_path
         self.data_root: zarr.Group | None = None
-        self._store: LocalStore | None = None
+        # determine type using extension
+        ext = self._store_path.suffix[1:]
```
tjhunter: you are repeating yourself

packages/common/src/weathergen/common/io.py
```diff
@@ -315,14 +328,50 @@ def _append_dataset(self, dataset: OutputDataset | None, name: str) -> None:
...
+            self.type = "local"
+        self.create = create
 
     def __enter__(self) -> typing.Self:
-        self._store = LocalStore(self._store_path)
-        self.data_root = zarr.group(store=self._store)
+        if self.type == "zip":
+            if self.create:
+                _logger.info("Creating zipstore")
+                self._store = ZipStore(self._store_path, mode="a")
```
tjhunter: append is more permissive, but there is a cost in opening and closing the store. I would rather have the store opened for the full duration of the write rather than for each chunk.

packages/common/src/weathergen/common/io.py
```diff
@@ -389,15 +439,25 @@ def _write_arrays(self, dataset_group: zarr.Group, dataset: OutputDataset):
...
             f"writing array: {name} with shape: {array.shape},chunks: {chunks}"
-            + "into group: {group}."
+            + f"into group: {group}."
         )
-        group.create_array(name, data=array, chunks=chunks)
+        start_time = timeit.default_timer()
+        if SHARDING_ENABLED and chunks != "auto":
+            shards = (SHARD_N_SAMPLES, *(5 * x for x in chunks[1:]))
+            group.create_array(name, data=array, chunks=chunks, shards=shards)
+            _logger.info(f"sharding enabled with shards: {shards} and chunks: {chunks}")
```
tjhunter: you should turn all the info messages here into debug, they are not relevant beyond debugging this feature

tjhunter (commented):
Some small style comments. Otherwise, it looks ready to try out on a larger scale.

When testing into the logic, I moved the creation of the zarr store straight into inference. Otherwise, I am personally confused when we write during inference and when we just calculate validation losses.

@grassesi : do you know why we don't write data when calling validation during training?

clessig:
> Some small style comments. Otherwise, it looks ready to try out on a larger scale.
> 
> When testing into the logic, I moved the creation of the zarr store straight into inference. Otherwise, I am personally confused when we write during inference and when we just calculate validation losses.
> 
> @grassesi : do you know why we don't write data when calling validation during training?

We don't write during validation since it's slows down the validation, generates additional files/storage, and is rarely used. There's a config parameter that controls this: https://github.com/ecmwf/WeatherGenerator/blob/1776b0ae830f7eb9cf5b942894624672eda3c56d/config/default_config.yml#L337 (in the process of being renamed).

packages/common/src/weathergen/common/io.py
enssow: extracted out into a scale factor parameter now for easier updating

enssow:
> > Latest version can read and write with the zipstore format but needs tidying up (produces lots of print statements to narrow down the issues when trying to run the evaluation from a .yml file)
> 
> What is the performance impact for running evaluation with using zipstore?

Hi Christian, sorry I missed this before the holidays - the latest tests are all in the hedge doc: https://gitlab.jsc.fz-juelich.de/hedgedoc/3X10vtdVQumQZy4qpBiHWg - there seems to be an small imporevemnt when using the ZipStore to evaluate

packages/common/src/weathergen/common/io.py
enssow: mode = "w" works with the new develop code, before I had some issues with overwriting so was using append - thanks for fixing this :)

packages/common/src/weathergen/common/io.py
```diff
@@ -377,6 +433,15 @@ def _get_group(self, item: ItemKey, create: bool = False) -> zarr.Array | zarr.G
...
     def _write_dataset(self, item_group: zarr.Group, dataset: OutputDataset):
+        # Constraint: the metadata has to be written at the same time as creating
+        # a group. When using zipstore, each update of the metadata creates a new
+        # entry in the zipstore with the same name, which is potentially unreadable
+        # or could be misinterpreted with a python version change.
+        # The current metadata also gets updated with each array update => requires
+        # more serious investigation
+        # TODO: fix these 2 issues:
+        # - metadata has to be correct from the start
+        # - write all metadata once at creation of the node
```
grassesi: The expectation is that _write_dataset is only called once per unique item key (and the associated group) in the store. Just to be sure is the current two step procedure a problem? (create group using `dataset_group = item_group.require_group` only then write metadata using `dataset_group.attrs["foo"] = "bar"` in `_write_metadata`)
If so `_write_metadata` could be changed into `_get_metadata` and return a dict that can be then used directly when creating the group: `dataset_group = require_group(dataset.name, attributes=_get_metadata())`

packages/common/src/weathergen/common/io.py
```diff
@@ -353,20 +408,21 @@ def load_zarr(self, key: ItemKey) -> OutputItem:
...
                 name, key, dict(dataset.arrays()), dict(dataset.attrs).copy()
             )
             for name, dataset in group.groups()
         }
 
-    def _get_group(self, item: ItemKey, create: bool = False) -> zarr.Array | zarr.Group:
+    def _get_group(self, item: ItemKey, create: bool) -> zarr.Array | zarr.Group:
         assert self.data_root is not None, "ZarrIO must be opened before accessing data."
         if create:
             group = self.data_root.create_group(item.path)
```
grassesi, line 443: we should assert here that no group under the same path already exists. This should prevent attempts to overwrite already written items.

packages/common/src/weathergen/common/io.py
```diff
@@ -392,15 +457,25 @@ def _write_arrays(self, dataset_group: zarr.Group, dataset: OutputDataset):
...
         _logger.debug(
             f"writing array: {name} with shape: {array.shape},chunks: {chunks}"
-            + "into group: {group}."
+            + f"into group: {group}."
         )
-        group.create_array(name, data=array, chunks=chunks)
+        start_time = timeit.default_timer()
+        if SHARDING_ENABLED and chunks != "auto":
+            shards = (SHARD_N_SAMPLES, *((SCALE_FACTOR + 1) * x for x in chunks[1:]))
+            group.create_array(name, data=array, chunks=chunks, shards=shards)
```
grassesi, line 479: would be nice if you can encapsulate this conditional:
```python
shards = _get_shards(chunks)
group.create_array(name, data=array, chunks=chunks, shards = shards)

def _get_shards(tuple[int]) -> dict[str, Any] | None:
   ...
```

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -224,6 +224,9 @@ def _process_stream(
         stream_computed_scores = calc_scores_per_stream(
             reader, stream, regions_to_compute, metrics_to_compute, plot_score_maps
         )
+        stream_dict = reader.get_stream(stream)
+        if not stream_dict:
+            return run_id, stream, {}
```
grassesi: Is this change required for the new zarr zip store? I am also surprised that this is needed. From the code I would have thought that `reader.get_stream` always returns at least an empty dict.

pyproject.toml
grassesi, line 1: Can you remove this file from your PR, since you did not make any real changes to `pyproject.toml`

src/weathergen/train/trainer.py
grassesi, line 1: Unfortunately I dont see the point of the ZarrIO context being opened in trainer. As far as I can tell your code is equivalent to how it was before (opening the ZarrIO context in `validation_io`) except that it completely disallows for output during training. But that we already have the `log_validation` parameter controlling if output is generated, that defaults to 0 during training (no output). I am happy to discuss this if there are other valid reasons.

src/weathergen/utils/validation_io.py
```diff
@@ -129,6 +131,8 @@ def write_output(cf, mini_epoch, batch_idx, dn_data, batch, model_output, target
...
     )
 
-    with io.ZarrIO(config.get_path_output(cf, mini_epoch)) as writer:
-        for subset in data.items():
-            writer.write_zarr(subset)
+    for subset in data.items():
+        start_time = timeit.default_timer()
+        writer.write_zarr(subset)
+        elapsed = timeit.default_timer() - start_time
+        _logger.info(f"writing subset: {subset} took {elapsed:.2f}")
```
grassesi: I think we want to have the timing logic in develop, what do you think @clessig ?

src/weathergen/utils/validation_io.py
grassesi, line 1: See my comment on `trainer.py`

src/weathergen/train/trainer_base.py
```diff
@@ -27,7 +27,7 @@ class TrainerBase:
     def __init__(self):
         self.device_handles = []
         self.device_names = []
-        self.cf: Config | None = None
+        self.cf: Config
```
grassesi: Please revert this, as otherwise self.cf is not properly initialized: Instead of getting `None` a `AttributeError: 'Trainer' object has no attribute 'cf'` would be raised. Even if this error would in practice never occure since self.cf is set right afterwards (in `self.init`), I think it is better to keep save.

src/weathergen/utils/cli.py
```diff
@@ -64,6 +64,14 @@ def get_inference_parser() -> argparse.ArgumentParser:
...
     )
 
+    parser.add_argument(
+        "--zarr-store",
+        default="zip",
+        help="Specify which type of Zarr storage to use. Default 'zip', which uses ZipStore.",
+        choices=["zip", "local"],
+        type=str,
+    )
+
```
grassesi: Since we have in general the capability to generate output during training this should be instead a general argument. I would also like to see it being added to `default_config.yaml` with a value of `zip`.

src/weathergen/run_train.py
```diff
@@ -48,6 +48,7 @@ def inference_from_args(argl: list[str]):
         samples_per_validation=args.samples,
         log_validation=args.samples if args.save_samples else 0,
         streams_output=args.streams_output,
+        zarr_store=args.zarr_store,
     )
```
grassesi: See my comment on `utils/cli.py`. I think it should be a proper option instead (with a default value in `default_config.yaml` etc.)

packages/common/src/weathergen/common/io.py
```diff
@@ -392,15 +457,25 @@ def _write_arrays(self, dataset_group: zarr.Group, dataset: OutputDataset):
...
+            _logger.debug(f"sharding enabled with shards: {shards} and chunks: {chunks}")
+        else:
+            group.create_array(name, data=array, chunks=chunks)
+            _logger.debug(f"sharding disabled, writing with chunks: {chunks}")
+        elapsed = timeit.default_timer() - start_time
+        _logger.debug(f"writing array: {name} took {elapsed:.2f}")
```
grassesi, line 485: I dont think we want that timing logic in develop.

packages/common/src/weathergen/common/io.py
```diff
@@ -250,12 +265,21 @@ def arrays(self) -> dict[str, ArrayType]:
...
+            shards = (shard_nsamples, *((SCALE_FACTOR + 1) * x for x in chunks[1:]))
+            _logger.info(f"sharding enabled with shards: {shards} and chunks: {chunks}")
+        else:
+            shards = None
+            _logger.info(f"sharding disabled, using chunks: {chunks}")
         # maybe do dask conversion earlier? => usefull for parallel writing?
-        data = da.from_zarr(self.data, chunks=chunks)  # dont call compute to lazy load
+        data = da.from_zarr(
+            self.data, chunks=chunks, shards=shards
+        )  # dont call compute to lazy load
```
grassesi, line 307: looks good :)

packages/common/src/weathergen/common/io.py
```diff
@@ -318,14 +342,45 @@ def _append_dataset(self, dataset: OutputDataset | None, name: str) -> None:
...
+                    (
+                        str(w.message)
+                        for w in reversed(caught)
+                        if issubclass(w.category, zarr.errors.ZarrUserWarning)
+                    ),
+                    "",
+                )
+                # warnings.warn(f"Zarr2 conflict: {last_msg}",
+                #                 DeprecationWarning
+                # )
```
grassesi: please remove if not important

packages/common/src/weathergen/common/io.py
```diff
@@ -318,14 +342,45 @@ def _append_dataset(self, dataset: OutputDataset | None, name: str) -> None:
...
+                _logger.info("Opening zipstore as read-only")
+                self._store = ZipStore(self._store_path, read_only=True)
+                self.data_root = zarr.open_group(store=self._store, mode="r")
+
+        elif self.type == "local":
+            # Capture warnings emitted during store creation/open
+            with warnings.catch_warnings(record=True) as caught:
+                self._store = LocalStore(self._store_path)
+                self.data_root = zarr.group(store=self._store)
+            # Raise DeprecationWarning only if a ZarrUserWarning was raised
```
grassesi: what is the cause of this warnings? Maybe add a short comment why we want to ignore it here.

grassesi (changes requested):
Great work, but I would like some structural changes: If possible `trainer.Trainer` should not be concerned with managing the ZarrIO context. In addition to the changes I requested below I would also like for the ZarrIO.__enter__ method to use polymorphism instead of switching behavior based on `self.type`. I would also like that number of times the literals indicating file extension/store_type (there is a one-to-one mapping) is reduced (eg. "zip" now occurs in 7 different places all over the code, similarly with "zarr"/"local"). I have implemented these two points in #1553 , feel free to just merge it if you like it.

packages/common/src/weathergen/common/io.py
enssow: For now it's okay as a 2 step process. However, the initial tests to try and get the intergration_test running were causing issues ([here](https://iffchat.fz-juelich.de/mlesm-dev/pl/1uj4fr8g3p89zmm1rokwiphkjw)). When this is merged i will open a new issue for this probelm. For now i'll move these comments out.

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
enssow: i was inherited from an older merge w/develop. I'll remove it

packages/common/src/weathergen/common/io.py
enssow: Result of discussion [here ](https://iffchat.fz-juelich.de/mlesm-dev/pl/t5urwr854pdrjgs91kesm6a1tc) - to summarise, because the opening old zarr2 stores can cause ZarrUserWarnings but in the future we should remove support for zarr2

packages/common/src/weathergen/common/io.py
enssow: yep agreed

src/weathergen/run_train.py
enssow: okay I've implemented this with these lines of [code](https://github.com/ecmwf/WeatherGenerator/blob/4fa4eb1ef11022494f1e6b61aeda20c12bc151a0/packages/common/src/weathergen/common/config.py#L336C1-L338C50) - it will take the default_config value of zarr_store if not specified as a command line argument

src/weathergen/utils/validation_io.py
enssow: removed for now (open to suggestions if there's a strong opinion either way)

packages/common/src/weathergen/common/io.py
grassesi: Please add a small comment explaining that :+1:

packages/common/src/weathergen/common/config.py
```diff
@@ -348,18 +348,22 @@ def load_merge_configs(
...
             for path in p:
                 c = _load_overwrite_conf(Path(path))
                 c = _load_streams_in_config(c)
                 overwrite_configs.append(c)
         else:
+            print("overwrite (dict or Config):", overwrite)
             # If it is a dict or DictConfig, we can directly use it
             c = _load_overwrite_conf(overwrite)
             c = _load_streams_in_config(c)
+            print("loaded overwrite config:", c)
```
grassesi: Please remove the print statements before merging.

packages/common/src/weathergen/common/config.py
```diff
@@ -369,14 +373,14 @@ def load_merge_configs(
...
+        from_run_id = base_config.general.run_id
 
     with open_dict(base_config):
         base_config.from_run_id = from_run_id
     # use OmegaConf.unsafe_merge if too slow
+    print("base_config:", base_config)
+    print("overwrite_configs:", overwrite_configs)
+    print("private_config:", private_config)
     c = OmegaConf.merge(base_config, private_config, *overwrite_configs)
     assert isinstance(c, Config)
```
grassesi, line 378: see above

packages/common/src/weathergen/common/config.py
```diff
@@ -385,7 +389,7 @@ def load_merge_configs(
     if hasattr(c, "samples_per_epoch"):
         c.samples_per_mini_epoch = c.samples_per_epoch
         c.num_mini_epochs = c.num_epochs
-
+    print(c.zarr_store)
     return c
```
grassesi, line 385: see above

packages/common/src/weathergen/common/config.py
```diff
@@ -654,6 +658,7 @@ def get_path_model(config: Config) -> Path:
 
 
 def get_path_output(config: Config, mini_epoch: int) -> Path:
+    print(config)
     ext = StoreType(config.zarr_store).value  # validate extension
```
grassesi: see above

src/weathergen/utils/validation_io.py
```diff
@@ -158,7 +158,7 @@ def write_output(
         sample_start,
         val_cfg.get("window_offset_prediction", 0),
     )
-
-    with writer(config.get_path_output(cf, mini_epoch)) as zio:
+    print(cf)
```
grassesi: see above

grassesi (approved):
I think it would be better if `zarr_store` would be under `general.zarr_store` but this is just nitpicking at this point. Please do remove the print statements tho.

enssow:
> I think it would be better if `zarr_store` would be under `general.zarr_store` but this is just nitpicking at this point. Please do remove the print statements tho.

Sorry, think the commit was not the most recent! prints all gone :) i did think about putting it under a different category but the `default_forecast_config` doesn't have a general section so it may be best just where it is - let me know what you think?

enssow:
`type-check` fails due to an import error with tqdm (this can be resolved by directly adding tqdm to the dependencies of common `pyproject.toml`. However, this then leads to further issues:
```
ERROR Type `object` is not iterable [not-iterable]
   --> packages/common/src/weathergen/common/io.py:412:24
    |
412 |         for dataset in tqdm(item.datasets):
    |                        ^^^^^^^^^^^^^^^^^^^
    |
 INFO 1 error (42 ignored)
 ```
 Which is a known issue with [pryefly](https://github.com/facebook/pyrefly/issues/1856)
 To discuss what is the best resolution :)

PR #1461, opened 2025-12-15
WeatherGenerator JSON reader

### LINKED ISSUE #1202: Evaluation fails if inference .zarr file not present
### What happened?

Currently, evaluation fails if the .zarr file is no longer present even though the metrics (or at least some) are already calculated. From the logic, would it make sense to be able to run the tool even if the .zarr file doesn't exist any longer? Use case would be for example if you want to plot against a baseline run for which you had calculated the metrics already previously. We often delete the zarr files now because of inodes. 
(If it is too complicated, MLflow might be solution in future because we can just select previous runs to compare against).

Error message: 

```
ERROR:weathergen.evaluate.io_reader:Zarr file /capstor/store/cscs/userlab/ch17/shared_work/results/jrpidu2s/validation_epoch00000_rank0000.zarr does not exist or is not a directory.
  File "/users/jkuehner/CODE/WeatherGenerator/.venv/bin/evaluate", line 10, in <module>
    sys.exit(evaluate())
             ^^^^^^^^^^
  File "/users/jkuehner/CODE/WeatherGenerator/packages/evaluate/src/weathergen/evaluate/run_evaluation.py", line 50, in evaluate
    evaluate_from_args(sys.argv[1:])
  File "/users/jkuehner/CODE/WeatherGenerator/packages/evaluate/src/weathergen/evaluate/run_evaluation.py", line 87, in evaluate_from_args
    evaluate_from_config(OmegaConf.load(config), mlflow_client)
  File "/users/jkuehner/CODE/WeatherGenerator/packages/evaluate/src/weathergen/evaluate/run_evaluation.py", line 114, in evaluate_from_config
    reader = WeatherGenReader(run, run_id, private_paths)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/jkuehner/CODE/WeatherGenerator/packages/evaluate/src/weathergen/evaluate/io_reader.py", line 353, in __init__
    raise FileNotFoundError(
FileNotFoundError: Zarr file /capstor/store/cscs/userlab/ch17/shared_work/results/jrpidu2s/validation_epoch00000_rank0000.zarr does not exist or is not a directory.
```

Run evaluate with an inference id that points to a results/ folder in which you have the evaluation/ folder but not the .zarr file. 

_No response_

s6sebusc:
## Description

Implements the option to read only the stored WeatherGen scores from JSON in the evaluation package, as requested in Issue [#1202](https://github.com/ecmwf/WeatherGenerator/issues/1202). To avoid duplication of code, the previous WeatherGenReader is split into a zarr reader and a json reader, which maybe wasn't the best idea. 

## Issue Number
Closes [#1202](https://github.com/ecmwf/WeatherGenerator/issues/1202)


packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -141,180 +81,132 @@ def evaluate_from_args(argl: list[str], log_queue: mp.Queue) -> None:
...
-                        reordered_dict[run_id][metric][region][stream] = data
+                    for run_id, final_dict in runs_dict.items():
+                        # Assign the final_dict to the new structure using the reordered keys
+                        reordered_dict[run_id][metric][region][stream] = final_dict
 
         channels_set = collect_channels(scores_dict, metric, region, runs)
 
         for run_id, run in runs.items():
-            reader = WeatherGenReader(run, run_id, private_paths)
+            reader = WeatherGenZarrReader(run, run_id, private_paths)
```
iluise: we should test that this does not crash with mlflow or skip the runs that do not use zarrReaders

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -101,19 +511,182 @@ def get_inference_config(self):
...
+
+class WeatherGenJSONReader(WeatherGenReader):
+
+    def __init__(self, eval_cfg: dict, run_id: str, private_paths: dict | None = None, actual_eval_cfg: dict = {}):
+        super().__init__(eval_cfg, run_id, private_paths)
+        # is this the best way to learn which steps and samples are available? 
+        stream=list(self.eval_cfg.streams.keys())[0]
+        region=actual_eval_cfg.regions[0]
+        metric=actual_eval_cfg.metrics[0] 
+        dummy = self.load_scores( stream, region, metric )
```
iluise: why dummy here?

iluise (commented):
I went through the code and I like the implementation. I will test it after it is updated to develop. 

p.s. do you have ideas on how we could split the weathergen_reader in multiple files? it is getting quite long and it might be useful to slice it into more files

clessig:
> I went through the code and I like the implementation. I will test it after it is updated to develop.
> 
> p.s. do you have ideas on how we could split the weathergen_reader in multiple files? it is getting quite long and it might be useful to slice it into more files

On the file length: we should split it up into one file per class, and have the different files in a sub-directory readers.

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
s6sebusc: the JSON reader needs to somehow know about the existing fsteps, samples and ens, so it can check_availability later. I guess that info is only found in the JSON files themselves, or is there another way?

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
s6sebusc: I changed it back to just use the generic WeatherGenReader here, I think that should work since it only uses the reader to find the `from_run_id` .

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -116,6 +96,201 @@ def get_inference_config(self):
...
+                return stream.get(key, default)
+        return default
+
+class WeatherGenJSONReader(WeatherGenReader):
+
+    def __init__(self, eval_cfg: dict, run_id: str, private_paths: dict | None = None, region: str = "global", metric: str = "rmse"):
+        super().__init__(eval_cfg, run_id, private_paths)
+        # is this the best way to learn which steps and samples are available? 
+        stream=list(self.eval_cfg.streams.keys())[0]
+        dummy = self.load_single_score( stream, region, metric )
```
iluise: I find that just relying on the first metric is sub-optimal, as the metrics could have been computed with different settings at different times. I would pass all metrics to the reader and assign self.samples, self.fsteps and self.ens as the intersection of the values among all metrics. I'd throw a warning if some metrics have some values that get excluded (and I would print the excluded samples/fsteps/ens in case). It takes a bit more time, but loading json files is fast anyway, so let's make it more user friendly ;)

iluise (commented):
Thanks for updating the branch. It works, but I'd make one change before merging it. See comments. 

Also, you should lint the code by using `./scripts/actions.sh lint` fix the errors that appear on screen and commit again.

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
s6sebusc: The new version of the JSONreader now goes through all streams, metrics, regions and the remembers the common fsteps, samples, ens. Logically, one would think that this could be done using `check_availability`, but `check_availability` expects that the reader already knows what fsteps etc. it can read, and also modifies the requested fsteps etc., so I just couldn't get it to work "cleanly".

iluise (approved):
Very nice!

iluise (approved):

PR #1507, opened 2025-12-19
Cropping with healpix disks and random walks, and possible overlaps

### LINKED ISSUE #1112: Implement global and local cropping and combine with StreamData

Implement global and local cropping for student-teacher training (where possible making use of existing healpix level masking), and combine this with StreamData so that appropriately masked fields are delivered to the student and teacher. 

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation

shmh40:
## Description

Based on @wael-mika's work this implements healpix-based cropping, using the healpix library to create disks (rather than crops like images). NB: see #1112 for plots of what the standard crop (with geodesic disk) look like, and what to be aware of with healpix_level.

To do: discuss with @sophie-xhonneux how exactly we want to determine the overlaps for different methods, or whether we can largely control this from the config (global crops always very large, local crops smaller, so we almost always have some overlap). Removed some overlap-based code that used the ratio to select the centre point but was never called anyway.

## Issue Number

Closes #1112 


src/weathergen/datasets/masking.py
```diff
@@ -475,6 +482,45 @@ def _generate_cell_mask(
...
 
+        elif strategy == "cropping_healpix":
+            # Spatial cropping: select contiguous region and keep it (mask rest)
+            # This is the inverse of healpix masking
+            hl_data = self.healpix_level_data
+            hl_mask = cfg.get("hl_mask")
+            assert hl_mask is not None and hl_mask < hl_data, (
+                "For cropping_healpix, cfg['hl_mask'] must be set and < data level."
+            )
+            num_parent_cells = 12 * (4**hl_mask)
```
sophie-xhonneux: is the number of parent cells `12* (4**(hl_mask-1))`

sophie-xhonneux: or I am misunderstanding the word `parent` here

src/weathergen/datasets/masking.py
```diff
@@ -475,6 +482,45 @@ def _generate_cell_mask(
                 mask = np.zeros(num_cells, dtype=bool)
                 mask[child_indices] = True
 
+        elif strategy == "cropping_healpix":
```
sophie-xhonneux, line 468: how come there is no constraint on the masks that can be chosen? (I might have forgotten this is done elsewhere, but the comments to the function still mention an argument `constraint_keep_mask`)

src/weathergen/datasets/masking.py
shmh40: Possibly misunderstanding. This "parent" is the level at which we do the cropping, i.e. we can choose to mask at hl_mask = 0. We call it parent because from this level, where we do the masking, we then need to project back down to the "child" level of e.g. level 5 at which we actually have the data.

src/weathergen/datasets/masking.py
shmh40: We currently handle this in _get_mask where we have the options "subset", "independent" etc. In these cases we produce a new student mask, and then we take the intersection/disjoint etc. with the teacher mask which has already been created. The user still has to think about what makes sense in terms of rates that are set in the config for e.g. global and local crops, if they want lots of overlap or not very much overlap. We should remove constraint_keep_mask, thanks.

src/weathergen/datasets/masking.py
```diff
@@ -47,7 +50,7 @@ class Masker:
     Attributes:
         masking_rate (float): The base rate at which tokens are masked.
         masking_strategy (str): The strategy used for masking (e.g., "random",
-        "block", "healpix", "channel").
+        "block", "healpix", "cropping_healpix", "channel").
```
clessig: `channel` is not functional at the moment. And I think the whole argument list here is very outdated. Either we update or remove it for the moment, to be added again later when things have stabilized.

clessig:
The integration test currently fails with:

```
  File "/users/lessig/santis/WeatherGenerator/src/weathergen/run_train.py", line 176, in train_with_args
    trainer.run(cf, devices)
  File "/users/lessig/santis/WeatherGenerator/src/weathergen/train/trainer.py", line 343, in run
    self.train(mini_epoch)
  File "/users/lessig/santis/WeatherGenerator/src/weathergen/train/trainer.py", line 367, in train
    for bidx, batch in enumerate(dataset_iter):
                       ^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/lessig/santis/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/utils/data/dataloader.py", line 708, in __next__
    data = self._next_data()
           ^^^^^^^^^^^^^^^^^
  File "/users/lessig/santis/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/utils/data/dataloader.py", line 1480, in _next_data
    return self._process_data(data)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/lessig/santis/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/utils/data/dataloader.py", line 1505, in _process_data
    data.reraise()
  File "/users/lessig/santis/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/_utils.py", line 733, in reraise
    raise exception
AssertionError: Caught AssertionError in DataLoader worker process 0.
Original Traceback (most recent call last):
  File "/users/lessig/santis/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/utils/data/_utils/worker.py", line 349, in _worker_loop
    data = fetcher.fetch(index)  # type: ignore[possibly-undefined]
           ^^^^^^^^^^^^^^^^^^^^
  File "/users/lessig/santis/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/utils/data/_utils/fetch.py", line 42, in fetch
    data = next(self.dataset_iter)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/lessig/santis/WeatherGenerator/src/weathergen/datasets/multi_stream_data_sampler.py", line 763, in __iter__
    batch = self._get_batch(idx, forecast_dt)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/lessig/santis/WeatherGenerator/src/weathergen/datasets/multi_stream_data_sampler.py", line 654, in _get_batch
    masks_streams, num_source_samples, num_target_samples = self._get_source_target_masks(mode)
                                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/lessig/santis/WeatherGenerator/src/weathergen/datasets/multi_stream_data_sampler.py", line 610, in _get_source_target_masks
    masks[stream_info["name"]] = self.tokenizer.masker.build_samples_for_stream(
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/lessig/santis/WeatherGenerator/src/weathergen/datasets/masking.py", line 315, in build_samples_for_stream
    target_mask, mask_params = self._get_mask(
                               ^^^^^^^^^^^^^^^
  File "/users/lessig/santis/WeatherGenerator/src/weathergen/datasets/masking.py", line 385, in _get_mask
    mask, params = self._generate_cell_mask(num_cells, strategy, masking_strategy_config)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/lessig/santis/WeatherGenerator/src/weathergen/datasets/masking.py", line 484, in _generate_cell_mask
    assert hl_mask is not None and hl_mask < hl_data, (
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: For cropping_healpix, cfg['hl_mask'] must be set and < data level.
```

I think `small_multi_stream.yaml` has just many things not specified that are expected, including `masking_straegy` and because of this we hit the error?

config/default_config.yml
```diff
@@ -110,7 +110,119 @@ training_config:
...
+  #     num_samples: 1  # if student-teacher, the number of local (student) views to generate
+  #     masking_strategy_config : { rate : 0.4 }
+  #   - masking_strategy: "random"
+  #     num_samples: 1  # if student-teacher, the number of local (student) views to generate
+  #     masking_strategy_config : { rate : 0.4 }
+   
+# ========================================
+# CROPPING EXAMPLES (cropping_healpix strategy)
+# ========================================
+# Spatial cropping: select and KEEP a spatially contiguous region, masking everything else
```
clessig, line 164: We need to find a better place for these examples.

src/weathergen/datasets/masking.py
```diff
@@ -312,19 +320,18 @@ def build_samples_for_stream(
...
                         source_mask, mask_params = self._get_mask(
                             num_cells=num_cells,
                             strategy=source_cfg.get("masking_strategy"),
-                            masking_strategy_config=source_cfg.get("masking_strategy_config", {}),
                             target_mask=target_mask,
+                            masking_strategy_config=source_cfg.get("masking_strategy_config", {}),
                             relationship=source_cfg.get("relationship", "independent"),
                         )
                         source_masks.add_mask(source_mask, mask_params, source_cfg)
+                        # TODO: proper correspondence between source and target
```
clessig: Why is there (potentially) a problem here?

src/weathergen/datasets/masking.py
```diff
@@ -475,6 +476,45 @@ def _generate_cell_mask(
                 mask = np.zeros(num_cells, dtype=bool)
                 mask[child_indices] = True
 
+        elif strategy == "cropping_healpix":
+            # Spatial cropping: select contiguous region and keep it (mask rest)
```
clessig: Can we have this in a separate function

src/weathergen/datasets/masking.py
```diff
@@ -483,3 +523,164 @@ def _generate_cell_mask(
...
+
+    # separate functions for the different methods of producing spatially contiguous regions
+    def _select_disk(self, center_cell: int, num_cells_to_select: int, nside: int) -> set[int]:
+        """
+        Select cells in a disk shape by expanding layer by layer.
+        """
+        selected = {center_cell}
+        frontier = {center_cell}
+
+        while len(selected) < num_cells_to_select and frontier:
```
clessig, line 572: How slow is this? This looks like potentially a big bottleneck to me. Is this running on `healpix_level` or a coarser level?

src/weathergen/datasets/masking.py
```diff
@@ -483,3 +523,164 @@ def _generate_cell_mask(
...
+                np.cos(all_lat) * np.cos(all_lon),
+                np.cos(all_lat) * np.sin(all_lon),
+                np.sin(all_lat),
+            ],
+            axis=1,
+        )
+        # Compute angular distances and select closest cells
+        dot_products = np.clip(np.dot(all_xyz, center_xyz), -1.0, 1.0)
+        angular_distances = np.arccos(dot_products)
+        selected = np.argsort(angular_distances)[:num_cells_to_select]
```
clessig, line 663: Could we also select all cells within a specific cone here? Wouldn't this be a good way for student-teacher relationship: student and teacher centers are x degree apart and both are cones of y_1 and y_2 degrees, respectively.

clessig (commented):
I was just running training for a bit and convergence is very slow: MSE is around 0.6 after 1000 steps (single GPU, hl_mask: 4). We should think about if this is to be expected and discuss. With the first example convergence looks better.

Also, what is the performance of these strategies?

src/weathergen/datasets/masking.py
shmh40: Updated it to be current

config/default_config.yml
shmh40: Agreed. Hedgedoc?

src/weathergen/datasets/masking.py
shmh40: Old comment, removed

src/weathergen/datasets/masking.py
shmh40: Done. Also added function for same healpix-specific calculations that happen both in healpix masking and cropping

src/weathergen/datasets/masking.py
shmh40: It is actually not too bad. It runs on the coarser level of hl_mask that we choose, but even on level 3 it doesn't seem considerably slower.

s/sec=1.173
s/sec=1.129
s/sec=1.136

src/weathergen/datasets/masking.py
shmh40: This could be good. Next PR, I can make an issue and tag @wael-mika

clessig: Yes, this should be a separate PR. And we can ablate before merging.

shmh40:
> I was just running training for a bit and convergence is very slow: MSE is around 0.6 after 1000 steps (single GPU, hl_mask: 4). We should think about if this is to be expected and discuss. With the first example convergence looks better.
> 
> Also, what is the performance of these strategies?

I would expect very slow convergence and high loss with MTM training with high masking rate and contiguous mask like we have here. Trying to predict e.g. 80% of all cells based on a local spatial crop of 20% is very hard/unconstrained, it should just be predicting climatology for a load of the distant cells I expect. 

image

Performance doesn't seem to be constrained by this, it is not dissimilar to standard masking, e.g.

disk:

s/sec=1.173
s/sec=1.129
s/sec=1.136

geodesic_disk: 

s/sec=1.134
s/sec=1.156
s/sec=1.138

clessig (approved):
Thanks @wael-mika , @shmh40 . Looks good now. Let's follow up with the PR about specifying cone/disk angle distributions and cone/disk sizes. (For me thinking about a cone centered at the Earth's center is better than a disk; it makes the relationships easier to characterize and parametrize).

PR #1520, opened 2025-12-25
Fix sample calculation jumping on run continuation

### LINKED ISSUE #587: Wrong `istep` when continue training with different nb of nodes
### What happened?

Continued an experiment and changing the nb of epochs as well as the nb of nodes. 
As a result the number of samples were calculated incorrectly, continuing at double nb. of samples. 

Image

Image

First I trained on a single node for 32 epochs. 
Then I continued training, simultaneously doubling nb of nodes and nb of epochs. 
Probably the first caused the problem. 

`../WeatherGenerator-private/hpc/launch-slurm.py --from-run-id h1c9ksya --options num_epochs=96`

### Version

develop

Linux

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

ECMWF

TillHae:
## Description

This PR addresses an issue where the number of samples and the starting epoch were calculated incorrectly when continuing a training run with a different number of nodes.
- Added samples as a config parameter. This parameter is always the current cumulative data processed across all training segments.
- Captured `world_size_original` in `run_train.py` before the environment initialization overwrites it. This ensures the model correctly interpretes the context of the run it is continuing from.
- EMA decay schedules now use total samples processed, preventing "jumps" or resets when scaling.
- Included Backward Compatibility in `TrainLogger` to safely initialize the new samples parameter from istep for legacy runs.

## Issue Number

Closes #587 


src/weathergen/train/trainer.py
```diff
@@ -351,17 +351,8 @@ def run(self, cf, devices, run_id_contd=None, mini_epoch_contd=None):
...
-            mini_epoch_base = int(self.cf.istep / len(self.data_loader))
-        else:
-            len_per_rank = (
-                len(self.dataset) // (self.world_size_original * cf.batch_size_per_gpu)
-            ) * cf.batch_size_per_gpu
-            mini_epoch_base = int(
-                self.cf.istep
-                / (min(len_per_rank, cf.samples_per_mini_epoch) * self.world_size_original)
-            )
+        # recover mini_epoch from cumulative samples (invariant to world size change)
```
tjhunter: we do not store the mini_epoch in the config state? that sounds dangerous, it should be the source of truth

src/weathergen/utils/train_logger.py
```diff
@@ -69,6 +69,13 @@ def __init__(self, cf, path_run: Path) -> None:
         self.cf = cf
         self.path_run = path_run
 
+        # Fallback for older runs where samples was not persisted
+        if "samples" not in self.cf or self.cf.samples == 0:
+            if self.cf.istep > 0:
+                self.cf.samples = (
```
tjhunter: also put a warning or info message, this is a workaround that may be inaccurate.

config/default_config.yml
```diff
@@ -338,6 +338,7 @@ log_validation: 0
 streams_output: ["ERA5"]
 
 istep: 0
+samples: 0
```
tjhunter: can you add a comment to indicate that these elements are the "dynamic" elements of the configuration that are updated during training.

Later, we should split all this info away from the  config itself. We are just using the config to store state.

tjhunter (commented):
@TillHae  thanks for looking into this annoying issue. Here are some initial comments.

src/weathergen/train/trainer.py
TillHae: I've created a separate issue for that #1543

clessig: > I've created a separate issue for that #1543

We removed `mini-epoch` from the config a while ago, and for good reasons. We cannot keep it consistent with istep. We agreed (there's a design doc) that istep should be the single source of truth and all functionality should entirely rely on this. Mini-epoch should just be a convenience that is used say for terminal output to orient the user.

TillHae:
This PR will be reopened as a new one.

PR #1554, opened 2026-01-06
Cropping overlap control using cone distance 

### LINKED ISSUE #1515: Explore student-teacher view overlap with positions and cones
In healpix_cropping masking, we could control overlap between teacher and student views with the following suggestion on PR #1507: 

could we also select all cells within a specific cone here? Wouldn't this be a good way for student-teacher relationship: student and teacher centers are x degree apart and both are cones of y_1 and y_2 degrees, respectively.

_Originally posted by @clessig in https://github.com/ecmwf/WeatherGenerator/pull/1507#discussion_r2641368239_

The relevant code is in the _select_geodesic_disk function in masking.py. We would also additionally need to pass the teacher information on to the student in build_samples_for_stream.

@wael-mika let us know what you think of this, whether you think it is interesting?

wael-mika:
## Description

This PR implements cone distance for spatial sampling. The core logic uses angular distance (in degrees/radii) to determine the spacing between the centers of the Teacher and Student cones/crops.

Key Features:
Geodesic Support: Currently optimized to work specifically with geodesic_disk.
Multi-Level Sampling: Allows Student and Teacher crops to be sampled from different hl_level masks.
Performance Optimizations: I’ve implemented caching for the Healpix object initialization to improve efficiency.
Geometric Logic: Uses spherical trigonometry to calculate destination points given a starting (lon, lat), distance, and azimuth (bearing).

Configuration:
Usage Example: Refer to default_config for the necessary parameters and an explanation of how center_distance_degrees is used for overlap control.
Documentation: I have included long/detailed docstrings for all new functions to make the review process easier.

## Issue Number

Closes #1515  

image
image
image


config/default_config.yml
```diff
@@ -248,6 +248,44 @@ training_config:
...
+        rate: 0.6                      # Teacher cone: 60% of sphere (~108° radius)
+        method: "geodesic_disk"        # Must be geodesic_disk for cone distance
+  model_input:
+    - masking_strategy: "cropping_healpix"
+      num_samples: 1
+      masking_strategy_config:
+        hl_mask: 3                     # Can be different from teacher
+        rate: 0.4                      # Student cone: 40% of sphere (~72° radius)
+        method: "geodesic_disk"        # Must be geodesic_disk
+        center_distance_degrees: 180    # REQUIRED: Student center is 45° from teacher
```
shmh40: This comment is a little unclear, the center_distance_degrees is required, but the center_distance_degrees can be any chosen angle right?

shmh40: Perhaps you could also include a range as below

src/weathergen/datasets/masking.py
```diff
@@ -81,6 +82,7 @@ def __init__(self, cf: Config):
         # number of healpix cells
         self.healpix_level_data = cf.healpix_level
         self.healpix_num_cells = 12 * (4**cf.healpix_level)
+        self._hp_cache = {}
```
shmh40, line 128: Good, I think this is helpful

src/weathergen/datasets/masking.py
```diff
@@ -351,6 +356,8 @@ def _get_mask(
             instance default if None.
         masking_strategy_config : dict | None
             Optional override of strategy config (e.g., {'hl_mask': 3}).
+        target_metadata : dict | None
```
shmh40: Perhaps state here that this is required when we do overlap with the cones? Otherwise it is unused right?

config/default_config.yml
wael-mika: Indeed, the comment is confusing as I had to set center_distance_degrees to 45 before submitting the PR ( I was trying different values to see if they work)
Any angle you choose should work as demonstrated in the example below.

src/weathergen/datasets/masking.py
wael-mika: True, it helped speeding up the computation (we do not have to repeat hp calculation every single time)

src/weathergen/datasets/masking.py
wael-mika: Agree, this is only used when doing cone overlap. I am not sure if there is a more elegant way to do it without adding target_metadata

src/weathergen/datasets/masking.py
```diff
@@ -374,6 +381,33 @@ def _get_mask(
...
+                "relationship 'cone_distance' requires 'center_distance_degrees' in config"
+            )
+
+            # Get teacher center cell and level
+            teacher_center_cell = target_metadata.get("center_cell", None)
+            assert teacher_center_cell is not None, (
+                "relationship 'cone_distance' requires 'center_cell' in target_metadata"
+            )
+            # Get teacher's hl_mask level to properly convert coordinates
+            teacher_hl_mask = target_metadata.get("hl_mask", masking_strategy_config.get("hl_mask", 0))
```
shmh40: Should this default to 0? Perhaps should check that hl_mask is specified with an assert like the above lines? Do we default to 0 in other parts of the code?

shmh40:
Fine for me. The overlap control is guarded with appropriate if statements so only affects things when we specify it in the config and the figures make it very clear. The logic in _get_mask is fine, and most code is appropriately in separate functions at the bottom of the file. Since this is very much a research feature and should not affect pipelines I am happy to merge once we have updated to develop and done the integration test.

Because of the delay (sorry about this) the branch needs to be merged with develop, and the configs should be removed @wael-mika, if that is ok, to keep develop clean, but it would be great to have an example config documented somewhere, just not in develop as already it is not very clear for users imo. Thank you very much for this work and sorry again for the delay.

shmh40 (commented):
Just needs to be updated to develop!

wael-mika:
Thanks a lot @shmh40 for the review.
I agree with you, it is a research feature for data augmentation, which is hopefully well separated and maintained within the code. The purpose is try it in multiple experiments and see if it provides values for different pre-training techniques. 
I will do my best to document it well and explain it carefully in our GitLab repo, update it, and merge it by the start of next week!

src/weathergen/datasets/masking.py
```diff
@@ -408,21 +421,198 @@ def _get_mask(
             mask = target_mask
             return mask, {}
 
+        # handle cone distance relationship
+        elif relationship == "cone_distance":
```
clessig, line 424: The bodies of the different `elif` branches need to go to separate functions.

src/weathergen/datasets/masking.py
```diff
@@ -408,21 +421,198 @@ def _get_mask(
...
+                center_distance_degrees = float(self.rng.choice(possible_values))
+            else:
+                center_distance_degrees = masking_strategy_config.get("center_distance_degrees")
+                assert center_distance_degrees is not None, (
+                    "relationship 'cone_distance' requires 'center_distance_degrees' or "
+                    "'center_distance_degrees_random: true' in config"
+                )
+
+            # Get teacher center cell from explicit metadata (not instance state)
+            teacher_center_cell = target_metadata.get("center_cell", None)
```
clessig: How is the `center_cell` specified?

src/weathergen/datasets/masking.py
```diff
@@ -408,21 +421,198 @@ def _get_mask(
...
+            )
+            params = {
+                "center_cell": student_center_cell,
+                "hl_mask": masking_strategy_config.get("hl_mask", 0),
+                "rate": masking_strategy_config.get("rate", 0.5),
+            }
+            return mask, params
+
+        # Handle contained_cone relationship (geometry-aware subset)
+        elif relationship == "contained_cone":
```
clessig, line 431: Same as above

src/weathergen/datasets/masking.py
```diff
@@ -408,21 +421,198 @@ def _get_mask(
...
+            else:
+                center_distance_degrees = masking_strategy_config.get("center_distance_degrees")
+                assert center_distance_degrees is not None, (
+                    "relationship 'cone_distance' requires 'center_distance_degrees' or "
+                    "'center_distance_degrees_random: true' in config"
+                )
+
+            # Get teacher center cell from explicit metadata (not instance state)
+            teacher_center_cell = target_metadata.get("center_cell", None)
+            assert teacher_center_cell is not None, (
```
clessig: Do we need the assert statements? If we use `target_metadata["center_cell"]` then we also hit an assert.

src/weathergen/datasets/masking.py
```diff
@@ -408,21 +421,198 @@ def _get_mask(
...
-            mask = mask & target_mask
+            # Check if geometry-aware processing is available
+            if (
+                strategy == "cropping_healpix"
+                and target_metadata is not None
+                and target_metadata.get("center_cell") is not None
+                and target_metadata.get("rate") is not None
+            ):
+                # Use geometry-aware contained_cone for spatial contiguity
+                logger.debug("Using geometry-aware contained_cone for subset with cropping_healpix")
```
clessig: Remove

src/weathergen/datasets/masking.py
```diff
@@ -408,21 +421,198 @@ def _get_mask(
...
-            mask = mask & (~target_mask)
+            # Check if geometry-aware processing is available
+            if (
+                strategy == "cropping_healpix"
+                and target_metadata is not None
+                and target_metadata.get("center_cell") is not None
+                and target_metadata.get("rate") is not None
+            ):
+                # Use geometry-aware separated_cone for spatial contiguity
+                logger.debug(
```
clessig: Remove

src/weathergen/datasets/masking.py
```diff
@@ -702,3 +919,389 @@ def _prepare_healpix_based_masking(self, cfg, keep_rate):
         num_parents_to_keep = int(np.round(keep_rate * num_parent_cells))
 
         return hl_mask, num_parent_cells, num_children_per_parent, num_parents_to_keep
+
+    def _get_hp_obj(self, healpix_level: int) -> hp.HEALPix:
```
clessig, line 1012: What is the runtime of the computations that are cached?

src/weathergen/datasets/masking.py
```diff
@@ -702,3 +919,389 @@ def _prepare_healpix_based_masking(self, cfg, keep_rate):
...
+
+        Note:
+            Cache is stored in self._hp_cache dictionary, initialized in __init__
+        """
+        if healpix_level not in self._hp_cache:
+            nside = 2**healpix_level
+            self._hp_cache[healpix_level] = hp.HEALPix(nside=nside, order="nested")
+        return self._hp_cache[healpix_level]
+
+    def _get_destination_latlon(
```
clessig, line 1032: better function name

src/weathergen/datasets/masking.py
```diff
@@ -702,3 +919,389 @@ def _prepare_healpix_based_masking(self, cfg, keep_rate):
...
+        self, origin_lon_rad: float, origin_lat_rad: float, distance_rad: float, azimuth_rad: float
+    ) -> tuple[float, float]:
+        """
+        Calculate destination point on sphere using great circle navigation.
+
+        Given a starting point (lon, lat), a distance, and an azimuth (bearing),
```
clessig: Given a starting point (lon, lat), a distance, and an azimuth (bearing),
        computes the destination point using spherical trigonometry.

src/weathergen/datasets/masking.py
```diff
@@ -702,3 +919,389 @@ def _prepare_healpix_based_masking(self, cfg, keep_rate):
...
+                - destination_lon_rad: Destination longitude in radians, normalized to [-π, π]
+                - destination_lat_rad: Destination latitude in radians [-π/2, π/2]
+
+        Mathematical Foundation:
+            Uses the spherical law of cosines for latitude and the spherical
+            law of sines for longitude. See:
+            https://www.movable-type.co.uk/scripts/latlong.html
+        """
+        # Calculate destination latitude using spherical law of cosines
+        dest_lat_rad = np.arcsin(
```
clessig, line 1061: These computations would likely be much simpler in R^3

src/weathergen/datasets/masking.py
wael-mika: Option 1: You can specify it in the config (This is according to your request to anchor Europe for example and keep the center fixed through the training if we want to deal with different datasets which might include a local one)
Option 2: If no anchor is configured, _select_spatially_contiguous_cells(..., center_cell=None) picks a random HEALPix cell in [0, 12 * 4**hl_mask) and save it in masking_params["center_cell"]
Option 3: For spatially aware relations ( cone_distance, contained_cone, separated_cone), the center cell is derived from the target, and then the center cell of the student is calculated based on the teacher's center

src/weathergen/datasets/masking.py
wael-mika: Done!

src/weathergen/datasets/masking.py
wael-mika: Done!

src/weathergen/datasets/masking.py
wael-mika: We can raise ValueError instead if we do not get a valid center:
```
teacher_center_cell = target_metadata.get("center_cell")
if teacher_center_cell is None:
    raise ValueError(
        "relationship 'cone_distance' requires 'center_cell' in target_metadata"
    )
```

src/weathergen/datasets/masking.py
wael-mika: _project_point_on_sphere?

src/weathergen/datasets/masking.py
wael-mika: From my understanding, it would require extra steps, convert lon/lat to a unit vector, build local east/north tangent vectors at the origin, turn azimuth into a tangent direction, advance/rotate on the sphere, convert back to lon/lat

If we are doing a lot of sphere math, I agree with you, but for the purpose of getting the destination-from-bearing on the sphere, these formulas are the most direct.

src/weathergen/datasets/masking.py
wael-mika: Updated!

src/weathergen/datasets/masking.py
wael-mika: It was not significant, I will test it again

src/weathergen/datasets/masking.py
wael-mika: Done!

src/weathergen/datasets/masking.py
wael-mika: Done

src/weathergen/datasets/masking.py
wael-mika: The cached work is sub-microsecond and not a meaningful runtime bottleneck compared with the actual masking logic. It is about 0.00006 to 0.00008 ms. The runtime is dominated by the actual masking operations like cone_search_lonlat, coordinate transforms, and mask construction, not by this cache.

wael-mika:
I updated the script to address your comments, tested it interactively with all possible combinations of masking. I also prepared an md doc to eexplain how to use these augmentations, which will be uploaded to the documentation on GitLab after merging.

PR #1590, opened 2026-01-13
Transformer predictor for JEPA

### LINKED ISSUE #1587: Implement transformer based predictors for JEPA

We need to implement a transformer based predictor for JEPA, following the literature (see https://www.overleaf.com/project/68dd158b1e327e168d9ceebd)

We will use the existing attention blocks and MLP layers for this. This resembles the forecasting engine architecture. A quick look revelas that it is not very straight forward to refactor these two to a single class, so first we will implement it as a seperate class.

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation

kctezcan:
## Issue Number

Closes #1587 


kctezcan:
Some design choices to discuss:
1) the latent space has the default dimensions of 12288x2048. The optimal predictor dimension is 384. We use a linear layer without bias at the entrance of the predictor to map 2048->384 and as the last layer another linear map without bias to map 384->2048. These dimensions are all config parameters.
2) We hardcoded 4 as the  MLP factor (factor for the hidden dimension in the MLP layers) as this is the optimal value from the literature.
3) we have introduced new parameters only in the jepa config: `pred_xxx`

config/default_config_jepa.yml
```diff
@@ -97,6 +103,7 @@ ema_halflife_in_thousands: 1e-3
 # ### Example validation and training config for student-teacher with JEPA
 validation_config:
   losses:
+    LossPhysical: {weight: 0.0, loss_fcts: [['mse', 0.8], ['mae', 0.2]]}
```
sophie-xhonneux: can we remove this if instead of setting the weight to 0?

config/default_config_jepa.yml
```diff
@@ -106,6 +113,7 @@ training_config:
   training_mode: "student_teacher"  # "masking", "student_teacher", "forecast"
   target_and_aux_calc: "EMATeacher"
   losses :  
+    LossPhysical: {weight: 0.0, loss_fcts: [['mse', 0.8], ['mae', 0.2]]}
```
sophie-xhonneux: Same here

sophie-xhonneux (approved):
Broadly, I am happy to approve the comments are addressed.

But I would like to talk about whether the global assimilation engine makes sense to be a part of the encoder or if for JEPA we should instead use the AggregationEngine (Global transformer without the masked tokens)

config/default_config_jepa.yml
kctezcan: this file is removed with the new configs

config/default_config_jepa.yml
kctezcan: this file is removed with the new configs

kctezcan:
I renamed the new parameters related to the JEPA predictor from `pred_` to `sslpred_` because I realized there were already parameters called `pred_` for the decoder (i.e. prediction heads)

sophie-xhonneux (approved):

config/config_physical_jepa.yml
```diff
@@ -65,6 +65,12 @@ fe_impute_latent_noise_std: 0.0  # 1e-4
 forecast_att_dense_rate: 1.0
 with_step_conditioning: True # False
 
+sslpred_num_blocks: 12
```
clessig, line 68: Where would this go into the new config structure?

src/weathergen/model/engines.py
```diff
@@ -835,6 +835,78 @@ class LatentState:
...
+        self.patch_token = patch_token
+
+        self.pred_blocks = nn.ModuleList()
+
+        # first map to intermediate_dim to introduce a bottleneck
+        self.pred_blocks.append(nn.Linear(in_dim, intermediate_dim, bias=False))
+
+        for _ in range(self.cf.sslpred_num_blocks):
+            self.pred_blocks.append(
+                MultiSelfAttentionHead(
```
clessig, line 863: We should have a transformer_block module for attention + MLP (can/should be in a separate PR)

src/weathergen/model/model.py
```diff
@@ -445,10 +446,12 @@ def create(self) -> "Model":
                         patch_token=True,
                     )
                 elif loss == "JEPA":
-                    self.latent_heads[loss] = LatentPredictionHead(
+                    self.latent_heads[loss] = TransformerPredictionHead(
```
clessig, line 449: Why is here no if-statement to chose between LatentPredictionHead and TransformerPredictionHead? LatentPredictionHead should be renamed to LatentPredictionHeadMLP (and then I would also prefer LatentPredictionHeadTransformer).

config/config_physical_jepa.yml
clessig: Option A is model but it's ultimately something specific to the JEPA loss term.

sophie-xhonneux:
I clean up the PR a bit, but didn't know how to open a PR against this PR so I put it here https://github.com/ecmwf/WeatherGenerator/pull/1649

clessig:
> I clean up the PR a bit, but didn't know how to open a PR against this PR so I put it here #1649

You need open the PR against the repo Kerem is using (MeteoSwiss fork of WeatherGenerator).

PR #1634, opened 2026-01-16
Enabling Geoinfo Feature for Anemoi Datasets

### LINKED ISSUE #1428: Enable geoinfo configuration for anemoi datasets

For the ERA5 stream, we have several channels that can be interpreted as geoinfo: `z`, `lsm`, `sdor`, and `slor` (altitude, land–sea mask, etc.). Similarly, for the CERRA stream we have `al` (which I assume corresponds to altitude). I am not sure whether there are additional geoinfo variables for CERRA. I haven't checked thoroughly.

Both ERA5 and CERRA streams are of type anemoi. I was wondering: can we add a line in the ERA5 stream configuration such as `geoinfo : [ "z", "lsm", "sdor", "slor"]` so that the model knows these channels should be treated as geoinfo? 

Currently, we can use these channels as forcings (e.g. by using `target_exclude: ...` but we cannot specify them as geoinfo (so that they are used in the decoder). This `geoinfo` feature for anemoi datasets should be treated in the same way as for #PR860 (https://github.com/ecmwf/WeatherGenerator/pull/860), i.e. that the user can specify `geoinfos: ...` in the stream config.

_No response_

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation

wael-mika:
## Description

Geoinfo channels are static/constant-in-time variables such as altitude, land-sea mask, orography, etc. These variables provide geographic context to the model and can be used in the decoder for location-aware predictions.

Previously, geoinfo was only supported for OBS-type streams. This implementation extends geoinfo support to anemoi-type streams (ERA5, CERRA, etc.).

## brief summary of the changes introduced in this pull request.

- Caching Logic: Added _cached_geoinfo to store data after the first read. Updated _get() to pull from this cache instead of hitting the disk repeatedly, and ensured init_empty() resets it.
- Channel Selection: Added select_geoinfo_channels() to filter the config for static fields (is_constant_in_time=True).
- Initialization: Updated __init__ to wire everything together—populating the channel indices, calculating normalization stats (mean/stdev), and priming the cache.

## Issue Number

Closes #1428 


clessig:
Geoinfo channels do not have to be constant in time. They can be dynamic variables and the goal of the issue is that any channel can be a geoinfo channel. So no caching is possible.

wael-mika:
Thanks for your comment @clessig. I will adapt accordingly.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -240,6 +240,19 @@ def __init__(
 
         self.mini_epoch = 0
 
+        # Log geoinfo configuration per stream
```
clessig: Please remove. This will be printed as part of the config

config/streams/era5_1deg_cerra/cerra.yml
```diff
@@ -0,0 +1,33 @@
+# (C) Copyright 2024 WeatherGenerator contributors.
```
clessig, line 1: Let's please remove all the config files. We need to find a place for them that is not the main repo.

clessig (commented):
See comments.

src/weathergen/datasets/multi_stream_data_sampler.py
wael-mika: Sure, I will

config/streams/era5_1deg_cerra/cerra.yml
wael-mika: The configs are here only for the reviewer to test the work, I will remove them before submitting the PR

wael-mika:
> Geoinfo channels do not have to be constant in time. They can be dynamic variables and the goal of the issue is that any channel can be a geoinfo channel. So no caching is possible.

As some of geoinfo channels are constant in time, do you think we should check whether that and cache the channel if it is constant? or treat them all as non-constant in time?

clessig:
> > Geoinfo channels do not have to be constant in time. They can be dynamic variables and the goal of the issue is that any channel can be a geoinfo channel. So no caching is possible.
> 
> As some of geoinfo channels are constant in time, do you think we should check whether that and cache the channel if it is constant? or treat them all as non-constant in time?

Let's treat them all as non-constant in time. The benefit would be so small that it's not worth the additional complexity.

wael-mika:
How do you think we should handle the NaN values in the geoinfo channels @clessig? 
For example in Cerra, the channels al, lsm, and orog contain a lot of NaN values which will effect the normalization which leads to "Skipping prediction" errors
My proposal is to mask nans with zeros in the base reader

clessig:
> How do you think we should handle the NaN values in the geoinfo channels @clessig? For example in Cerra, the channels al, lsm, and orog contain a lot of NaN values which will effect the normalization which leads to "Skipping prediction" errors My proposal is to mask nans with zeros in the base reader

NaN values are handled later in the MultiStreamDataSampler. So nothing has to be done from your side.

clessig:
Will be merged as #1760 onto develop.

PR #1655, opened 2026-01-19
Skipping missing scores in JSONreader

### LINKED ISSUE #1605: Make the JSONReader fail silently if a metric is not there

The JSONReader now fails if one metric is not present, but as for the zarr case, if some variable is not present the code should use skip it and fire a warning that the metric was not found. 

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation

s6sebusc:
## Description
Previously, the evaluation with `type: json` would only work if all requested metrics exist. Now the code is supposed to skip missing metrics, as well as requested map-plots, and recommend the user set `type: zarr` if possible. 

## Issue Number
Closes [1605](https://github.com/ecmwf/WeatherGenerator/issues/1605)


packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -227,7 +227,10 @@ def _process_stream(
 
     # Parallel plotting
     if stream_dict.get("plotting"):
-        plot_data(reader, stream, global_plotting_opts)
+        if type_ == "zarr":
```
iluise: plotting is only enabled with type == zarr.  

no need to do the if/else here. we can just do: 

```
if stream_dict.get("plotting") and type_ == "zarr":
    plot_data(...)
 
```

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -241,15 +244,23 @@ def _process_stream(
...
+            metrics_to_compute = missing_metrics if missing_metrics else metrics
 
-        metric_list_to_json(reader, stream, stream_computed_scores, regions)
-        scores_dict = merge(stream_loaded_scores, stream_computed_scores)
+            stream_computed_scores = calc_scores_per_stream(
+                reader, stream, regions_to_compute, metrics_to_compute, plot_score_maps
+            )
+            metric_list_to_json(reader, stream, stream_computed_scores, regions)
+            scores_dict = merge(stream_loaded_scores, stream_computed_scores)
+        else:
```
iluise: I think there's a better way to make this fail silently. The JSONReader should never return values for `missing_scores` here:
https://github.com/ecmwf/WeatherGenerator/blob/develop/packages/evaluate/src/weathergen/evaluate/run_evaluation.py#L236

so it will skip the computation part.

The final dictionary is just the sum of loaded + computed. For JSON we will just have the loaded part.

https://github.com/ecmwf/WeatherGenerator/blob/develop/packages/evaluate/src/weathergen/evaluate/run_evaluation.py#L252 

We can briefly chat if it's not clear.

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -217,37 +217,40 @@ def _process_stream(
...
         stream,
         regions,
         metrics,
     )
     scores_dict = stream_loaded_scores
 
-    if missing_metrics or plot_score_maps:
-        regions_to_compute = list(set(missing_metrics.keys())) if missing_metrics else regions
-        metrics_to_compute = missing_metrics if missing_metrics else metrics
+    if computable_metrics or plot_score_maps:
```
iluise: can we put this here:

```
if recomputable_metrics != None or (plot_score_maps && type_ == "zarr"):
     regions_to_compute ...

```

and remove the above:

```
if plot_score_maps and type_ != "zarr":
        _logger.info("score maps skipped, use type:zarr for that.")
        plot_score_maps = False
```

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -296,6 +296,12 @@ def get_data(self, *args, **kwargs):
         # it can still happen when a particular score was available for a different channel
         raise ValueError(f"Missing JSON data for run {self.run_id}.")
 
+    def check_computability(self, metrics):
```
iluise: can we rename it like: `get_recomputable_metrics` ?

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -196,8 +197,8 @@ def load_scores(
                 # all other cases: recompute scores
                 missing_metrics.setdefault(region, []).append(metric)
                 continue
-
-        return local_scores, missing_metrics
+        computable_metrics = self.check_computability(missing_metrics)
```
iluise: what about a more explicit name like e.g. `recomputable_missing_metrics` since they come from `missing_metrics`?

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -217,37 +217,40 @@ def _process_stream(
...
-    if stream_dict.get("plotting"):
+    if stream_dict.get("plotting") and type_ == "zarr":
         plot_data(reader, stream, global_plotting_opts)
 
     # Scoring per stream
     if not stream_dict.get("evaluation"):
         return run_id, stream, {}
 
-    stream_loaded_scores, missing_metrics = reader.load_scores(
+    stream_loaded_scores, computable_metrics = reader.load_scores(
```
iluise: small thing, but can we change it into `recomputable_metrics`?

iluise (approved):
Thanks! yes this is much better now! Approving with some minor style naming suggestions :)

iluise (approved):

PR #1660, opened 2026-01-20
SYNOP datareader functionality for German stations

TillHae:
## Description

This PR changes the SYNOP datareader to also work on german SYNOP station data instead of just the one from MetNorway. The following bullet points are an explanation of all my changes.

**The bullet points are partially outdated**

- added height dimension, which affected the base class as well
- German data does not need to handle missing values as they are already included as np.nan (e.g. ll. 164-166)
- German data is structured after time and station_id. Geoinfo has to be flattened to 1-D before usage (ll.105-113)
- German data dimensions needed to be transformed as they are (variables, spatial, time) instead of (variables, time, spatial) (ll. 231-238)
- I've added a file for testing, which will be removed after the review (The test only works on juwels, as the underlying data is only available there)

## Issue Number

This PR closes no issue, but is related to #862 

### DIFF
--- a/config/streams/era5_nppatms_synop/synop.yml
+++ b/config/streams/era5_nppatms_synop/synop.yml
@@ -6,6 +6,8 @@
 SurfaceCombined :
   type : obs
   filenames : ['observations-ea-ofb-0001-1979-2023-combined-surface-v2.zarr']
+  mean_key : ''     # key for stored mean in dataset
+  stdev_key : ''    # key for stored stdev in dataset
   loss_weight : 1.0
   masking_rate : 0.6
   masking_rate_none : 0.05
--- a/config/streams/era5_nppatms_synop/synop.yml
+++ b/config/streams/era5_nppatms_synop/synop.yml
@@ -6,8 +6,9 @@
 SurfaceCombined :
   type : obs
   filenames : ['observations-ea-ofb-0001-1979-2023-combined-surface-v2.zarr']
-  mean_key : ''     # key for stored mean in dataset
-  stdev_key : ''    # key for stored stdev in dataset
+  mean_key :       # key for stored mean in dataset
+  stdev_key :      # key for stored stdev in dataset
+  fillvalue :      # fill value for missing data
   loss_weight : 1.0
   masking_rate : 0.6
   masking_rate_none : 0.05
--- a/config/streams/era5_nppatms_synop/synop.yml
+++ b/config/streams/era5_nppatms_synop/synop.yml
@@ -4,8 +4,8 @@
 # 2 : conventional observations
 
 SurfaceCombined :
-  type : obs
-  filenames : ['observations-ea-ofb-0001-1979-2023-combined-surface-v2.zarr']
+  type : station
+  filenames : []
   mean_key :       # key for stored mean in dataset
   stdev_key :      # key for stored stdev in dataset
   fillvalue :      # fill value for missing data
--- a/src/weathergen/datasets/data_reader_base.py
+++ b/src/weathergen/datasets/data_reader_base.py
@@ -626,6 +626,18 @@ class DataReaderTimestep(DataReaderBase):
             self.time_window_handler,
         )
 
+    def _clip_lat(lats: NDArray) -> NDArray[np.float32]:
+        """
+        Clip latitudes to the range [-90, 90] and ensure periodicity.
+        """
+        return (2 * np.clip(lats, -90.0, 90.0) - lats).astype(np.float32)
+
+    def _clip_lon(lons: NDArray) -> NDArray[np.float32]:
+        """
+        Clip longitudes to the range [-180, 180] and ensure periodicity.
+        """
+        return ((lons + 180.0) % 360.0 - 180.0).astype(np.float32)
+
 
 # to avoid rounding issues
 # The basic time precision is 1 millisecond.
--- a/src/weathergen/datasets/data_reader_base.py
+++ b/src/weathergen/datasets/data_reader_base.py
@@ -185,7 +185,7 @@ class ReaderData:
     datetimes: NDArray[NPDT64]
 
     @staticmethod
-    def empty(num_data_fields: int, num_geo_fields: int, num_coord_fields: int = 2) -> "ReaderData":
+    def empty(num_data_fields: int, num_geo_fields: int) -> "ReaderData":
         """
         Create an empty ReaderData object
 
@@ -195,7 +195,7 @@ class ReaderData:
             Empty ReaderData object
         """
         return ReaderData(
-            coords=np.zeros((0, num_coord_fields), dtype=np.float32),
+            coords=np.zeros((0, 2), dtype=np.float32),
             geoinfos=np.zeros((0, num_geo_fields), dtype=np.float32),
             data=np.zeros((0, num_data_fields), dtype=np.float32),
             datetimes=np.zeros((0,), dtype=np.datetime64),
--- a/src/weathergen/datasets/data_reader_base.py
+++ b/src/weathergen/datasets/data_reader_base.py
@@ -232,8 +232,8 @@ def check_reader_data(rdata: ReaderData, dtr: DTRange) -> None:
     """
 
     assert rdata.coords.ndim == 2, f"coords must be 2D {rdata.coords.shape}"
-    assert rdata.coords.shape[1] == 2 or rdata.coords.shape[1] == 3, (
-        f"coords must have 2 or 3 columns (lat, lon(, height)), got {rdata.coords.shape}"
+    assert rdata.coords.shape[1] == 2, (
+        f"coords must have 2 columns (lat, lon), got {rdata.coords.shape}"
     )
     assert rdata.geoinfos.ndim == 2, f"geoinfos must be 2D, got {rdata.geoinfos.shape}"
     assert rdata.data.ndim == 2, f"data must be 2D {rdata.data.shape}"
--- a/src/weathergen/datasets/data_reader_base.py
+++ b/src/weathergen/datasets/data_reader_base.py
@@ -185,7 +185,7 @@ class ReaderData:
     datetimes: NDArray[NPDT64]
 
     @staticmethod
-    def empty(num_data_fields: int, num_geo_fields: int) -> "ReaderData":
+    def empty(num_data_fields: int, num_geo_fields: int, num_coord_fields: int = 2) -> "ReaderData":
         """
         Create an empty ReaderData object
 
@@ -195,7 +195,7 @@ class ReaderData:
             Empty ReaderData object
         """
         return ReaderData(
-            coords=np.zeros((0, 2), dtype=np.float32),
+            coords=np.zeros((0, num_coord_fields), dtype=np.float32),
             geoinfos=np.zeros((0, num_geo_fields), dtype=np.float32),
             data=np.zeros((0, num_data_fields), dtype=np.float32),
             datetimes=np.zeros((0,), dtype=np.datetime64),
@@ -232,8 +232,8 @@ def check_reader_data(rdata: ReaderData, dtr: DTRange) -> None:
     """
 
     assert rdata.coords.ndim == 2, f"coords must be 2D {rdata.coords.shape}"
-    assert rdata.coords.shape[1] == 2, (
-        f"coords must have 2 columns (lat, lon), got {rdata.coords.shape}"
+    assert rdata.coords.shape[1] == 2 or rdata.coords.shape[1] == 3, (
+        f"coords must have 2 or 3 columns (lat, lon(, height)), got {rdata.coords.shape}"
     )
     assert rdata.geoinfos.ndim == 2, f"geoinfos must be 2D, got {rdata.geoinfos.shape}"
     assert rdata.data.ndim == 2, f"data must be 2D {rdata.data.shape}"
--- a/src/weathergen/datasets/data_reader_base.py
+++ b/src/weathergen/datasets/data_reader_base.py
@@ -626,18 +626,6 @@ class DataReaderTimestep(DataReaderBase):
             self.time_window_handler,
         )
 
-    def _clip_lat(self, lats: NDArray) -> NDArray[np.float32]:
-        """
-        Clip latitudes to the range [-90, 90] and ensure periodicity.
-        """
-        return (2 * np.clip(lats, -90.0, 90.0) - lats).astype(np.float32)
-
-    def _clip_lon(self, lons: NDArray) -> NDArray[np.float32]:
-        """
-        Clip longitudes to the range [-180, 180] and ensure periodicity.
-        """
-        return ((lons + 180.0) % 360.0 - 180.0).astype(np.float32)
-
 
 # to avoid rounding issues
 # The basic time precision is 1 millisecond.
@@ -706,3 +694,17 @@ def get_dataset_indexes_timestep(
     end_didx = start_didx + int((dtr.end - dtr.start - t_epsilon) / period)
 
     return (np.arange(start_didx, end_didx + 1, dtype=np.int64), dtr)
+
+
+def _clip_lat(lats: NDArray) -> NDArray[np.float32]:
+    """
+    Clip latitudes to the range [-90, 90] and ensure periodicity.
+    """
+    return (2 * np.clip(lats, -90.0, 90.0) - lats).astype(np.float32)
+
+
+def _clip_lon(lons: NDArray) -> NDArray[np.float32]:
+    """
+    Clip longitudes to the range [-180, 180] and ensure periodicity.
+    """
+    return ((lons + 180.0) % 360.0 - 180.0).astype(np.float32)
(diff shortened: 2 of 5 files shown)


TillHae:
@wael-mika

TillHae:
@clessig should I also complete the open TODOs that are mentioned through comments as part of this PR?

clessig:
> @clessig should I also complete the open TODOs that are mentioned through comments as part of this PR?

Yes, this would be great! Let me know if you have questions about this.

TillHae:
> > @clessig should I also complete the open TODOs that are mentioned through comments as part of this PR?
> 
> Yes, this would be great! Let me know if you have questions about this.

Yes, I don't know what exactly is meant with the caching of the mean and standard deviation. It looks like they are already stored in the object. Do you also want them to be stored in the dataset file for the next time the dataset is read?

clessig:
> > > @clessig should I also complete the open TODOs that are mentioned through comments as part of this PR?
> > 
> > 
> > Yes, this would be great! Let me know if you have questions about this.
> 
> Yes, I don't know what exactly is meant with the caching of the mean and standard deviation. It looks like they are already stored in the object. Do you also want them to be stored in the dataset file for the next time the dataset is read?

They are computed on the fly if they are not present in the data file. That's expensive and not an option if the dataset if bigger. We should also not modify the existing data file; it should be considered immutable. We could generate an auxiliary file with the data, but then we need to decide where to store it.

TillHae:
> > > > @clessig should I also complete the open TODOs that are mentioned through comments as part of this PR?
> > > 
> > > 
> > > Yes, this would be great! Let me know if you have questions about this.
> > 
> > 
> > Yes, I don't know what exactly is meant with the caching of the mean and standard deviation. It looks like they are already stored in the object. Do you also want them to be stored in the dataset file for the next time the dataset is read?
> 
> They are computed on the fly if they are not present in the data file. That's expensive and not an option if the dataset if bigger. We should also not modify the existing data file; it should be considered immutable. We could generate an auxiliary file with the data, but then we need to decide where to store it.

It looks like they are always computed on the fly even when they are present in the data file. I think the best way to handle it is to have just one file where we store a hashmap like structure with the datasets as the key and their corresponding properties stored as values. This is the simplest solution to manage and enables us to seamlessly append datasets and properties to the file.

clessig:
> > > > > @clessig should I also complete the open TODOs that are mentioned through comments as part of this PR?
> > > > 
> > > > 
> > > > Yes, this would be great! Let me know if you have questions about this.
> > > 
> > > 
> > > Yes, I don't know what exactly is meant with the caching of the mean and standard deviation. It looks like they are already stored in the object. Do you also want them to be stored in the dataset file for the next time the dataset is read?
> > 
> > 
> > They are computed on the fly if they are not present in the data file. That's expensive and not an option if the dataset if bigger. We should also not modify the existing data file; it should be considered immutable. We could generate an auxiliary file with the data, but then we need to decide where to store it.
> 
> It looks like they are always computed on the fly even when they are present in the data file. I think the best way to handle it is to have just one file where we store a hashmap like structure with the datasets as the key and their corresponding properties stored as values. This is the simplest solution to manage and enables us to seamlessly append datasets and properties to the file.

Having one file is appealing at first sight but it comes with a lot of question marks: most datasets already contain the information and we shouldn't duplicate it since we otherwise run the risk that it becomes inconsistent. Having one file for all other datasets would mean we need to copy it to all other HPCs whenever something for one dataset changes. This also seems sub-optimal. 

My suggestion is to have a stream config argument that specifies which field in the data file contains mean and stdev. If it's not specified or does not exist, then we fall back to computing it.

TillHae:
@clessig the test script also works for the MetNo data. I just needed to change some values in the stream_info.

TillHae:
@wael-mika @clessig I'd love to get this finished. If you think there is anything left to do please let me know so we can finally get everything done.

config/streams/era5_nppatms_synop/synop.yml
```diff
@@ -6,6 +6,8 @@
 SurfaceCombined :
   type : obs
   filenames : ['observations-ea-ofb-0001-1979-2023-combined-surface-v2.zarr']
+  mean_key : ''     # key for stored mean in dataset
+  stdev_key : ''    # key for stored stdev in dataset
```
wael-mika: Is this needed for all synops or specific to the German Synop data?

clessig: We need the option to specify this

src/weathergen/datasets/data_reader_base.py
```diff
@@ -185,7 +185,7 @@ class ReaderData:
     datetimes: NDArray[NPDT64]
 
     @staticmethod
-    def empty(num_data_fields: int, num_geo_fields: int) -> "ReaderData":
+    def empty(num_data_fields: int, num_geo_fields: int, num_coord_fields: int = 2) -> "ReaderData":
```
clessig: coords are always lat/lon and size=2. What is your third coord? `height` is a geoinfo

src/weathergen/datasets/data_reader_base.py
```diff
@@ -626,6 +626,18 @@ def _get_dataset_idxs(self, idx: TIndex) -> tuple[NDArray[np.int64], DTRange]:
             self.time_window_handler,
         )
 
+    def _clip_lat(self, lats: NDArray) -> NDArray[np.float32]:
```
clessig: Can we put this at the end of the file

src/weathergen/datasets/data_reader_synop.py
```diff
@@ -88,21 +88,50 @@ def __init__(
             self.ds = ds
             self.len = len(ds)
 
-        self.offset_data_channels = 4
-        self.fillvalue = ds["air_temperature"][0, 0].values.item()
-        self.channels_file = [k for k in self.ds.keys()]
+        # handle MetNo SYNOP netCDF files
+        if "air_temperature" in ds.keys():
```
clessig: This is not generic. We cannot have a case distinction here but everything should be sufficiently configurable from the stream config. E.g., for the fillvalue above, this can be specified in the stream config.

src/weathergen/datasets/data_reader_synop.py
```diff
@@ -88,21 +88,50 @@ def __init__(
...
+            assert False, "Unknown dataset type"
+
+        if self.ds_type == "MetNo":
+            self.fillvalue = ds["air_temperature"][0, 0].values.item()
+
+        self.channels_file = list(ds.keys())
 
-        # caches lats and lons
+        # helper to get 1D station-static data for german stations
+        def _get_1d(var_name):
```
clessig: What does this function do?

src/weathergen/datasets/data_reader_synop.py
```diff
@@ -195,18 +259,29 @@ def _get(self, idx: TIndex, channels_idx: list[int]) -> ReaderData:
         # subsetting is pushed to the ctor via frequency argument; this also ensures that no sub-
         # sampling is required here
         sel_channels = [self.channels_file[i] for i in channels_idx]
-        data = self.ds[sel_channels].isel(time=slice(didx_start, didx_end)).to_array().values
+        data = self.ds[sel_channels].isel(time=slice(didx_start, didx_end)).to_array()
+
+        # Ensure dimensions are (variables, time, spatial)
+        # German files are (variables, spatial, time)
+        # MetNo files are (variables, time, spatial)
+        spatial_dim = "location" if "location" in data.dims else "station_id"
```
clessig: This should bee solved through appropriate indexing, i.e. one has an index list for variables, time, and the spatial coordinates. Then the order shouldn't matter.

src/weathergen/datasets/data_reader_synop.py
```diff
@@ -195,18 +259,29 @@ def _get(self, idx: TIndex, channels_idx: list[int]) -> ReaderData:
...
+            data = data.transpose("variable", "time", spatial_dim)
+
+        data = data.values
         # flatten along time dimension
         data = data.transpose([1, 2, 0]).reshape((data.shape[1] * data.shape[2], data.shape[0]))
-        # set invalid values to NaN
-        mask = data == self.fillvalue
-        data[mask] = np.nan
+        # set invalid values to NaN for MetNo nc files
+        if self.ds_type == "MetNo":
```
clessig:
```
if self.fillvalue is not None
```

src/weathergen/datasets/data_reader_synop.py
```diff
@@ -195,18 +259,29 @@ def _get(self, idx: TIndex, channels_idx: list[int]) -> ReaderData:
...
+            mask = data == self.fillvalue
+            data[mask] = np.nan
 
-        # construct lat/lon coords
+        # construct lat/lon/height coords
         latlon = np.concatenate(
             [
                 np.expand_dims(self.latitudes, 0),
                 np.expand_dims(self.longitudes, 0),
+                np.expand_dims(self.heights, 0),
```
clessig: height is geoinfo

config/streams/era5_nppatms_synop/synop.yml
clessig: But `observations-ea-ofb-0001-1979-2023-combined-surface-v2.zarr` is type obs and not the new type.

src/weathergen/datasets/data_reader_base.py
wael-mika: Agree, this would break much of the code in the current develop! The heigh should be in geoinfo

src/weathergen/datasets/data_reader_synop.py
```diff
@@ -88,21 +88,50 @@ def __init__(
...
+        if "air_temperature" in ds.keys():
+            self.ds_type = "MetNo"
+        elif "TT_10" in ds.keys():
+            self.ds_type = "Germany"
+        else:
+            self.ds_type = "Unknown"
+            assert False, "Unknown dataset type"
+
+        if self.ds_type == "MetNo":
+            self.fillvalue = ds["air_temperature"][0, 0].values.item()
```
wael-mika: So the logic here is to extract this value and use it to fill NaN values, but this can silently corrupt valid data by treating the first observed value as the fill value. Can you please check if this behaviour is inteded with MetNo?

tests/test_german_synop_reader.py
```diff
@@ -0,0 +1,39 @@
+import numpy as np
+from pathlib import Path
+from weathergen.datasets.data_reader_synop import DataReaderSynop
+from weathergen.datasets.data_reader_base import TimeWindowHandler
+
+def test_reader():
+    test_file = Path("/p/project1/hclimrep/hauer1/synop/data/datasetv1/2019.nc")
```
wael-mika, line 7: Does this data file exist on Santis?

src/weathergen/datasets/data_reader_synop.py
clessig: This is always a risk with non-NaN fillvalues, which didn't exist historically.

tests/test_german_synop_reader.py
TillHae: No, it only exists on the JSC systems

src/weathergen/datasets/data_reader_base.py
TillHae: I made the mistake to have height as a coord in the german dataset. I will fix that and then this change becomes obsolete

src/weathergen/datasets/data_reader_synop.py
TillHae: It is used to extract the static values (coords and geoinfo). I will make some changes to the german dataset, which should solve this problem.

TillHae:
I still have to make changes to the german synop station dataset to test it.

TillHae:
I didn't test it on the german synop station dataset yet, but for the one by MetNorway everything works for me. Ones the german one is regenerated I can test it on it as well.

TillHae:
> I didn't test it on the german synop station dataset yet, but for the one by MetNorway everything works for me. Ones the german one is regenerated I can test it on it as well.

I finally got to test it on the German one and it works as well:
```
Reader initialized successfully.
Data shape: (39672, 1)
Coords shape: (39672, 2)
Geoinfos shape: (39672, 1)
Test finished!
```

TillHae:
@clessig I still think the way we load the mean and stdev values can be improved. We should make sure that once they are computed, we save them so subsequent runs can load them directly without recomputing. What do you think is the best way to handle that?

clessig:
> @clessig I still think the way we load the mean and stdev values can be improved. We should make sure that once they are computed, we save them so subsequent runs can load them directly without recomputing. What do you think is the best way to handle that?

Ideally, mean and stdev should be determined in a precomputation step. It can then be stored directly in the dataset (as it's done for anemoi and our observation datasets) or alongside it (e.g. on meteocloud). Would this be possible?

TillHae:
> > @clessig I still think the way we load the mean and stdev values can be improved. We should make sure that once they are computed, we save them so subsequent runs can load them directly without recomputing. What do you think is the best way to handle that?
> 
> Ideally, mean and stdev should be determined in a precomputation step. It can then be stored directly in the dataset (as it's done for anemoi and our observation datasets) or alongside it (e.g. on meteocloud). Would this be possible?

It should be. Do you know anyone who is familiar with that so I can get in touch with them or is there any documentation I can use to get a better feeling on how it works for Anemoi?

clessig:
> > > @clessig I still think the way we load the mean and stdev values can be improved. We should make sure that once they are computed, we save them so subsequent runs can load them directly without recomputing. What do you think is the best way to handle that?
> > 
> > 
> > Ideally, mean and stdev should be determined in a precomputation step. It can then be stored directly in the dataset (as it's done for anemoi and our observation datasets) or alongside it (e.g. on meteocloud). Would this be possible?
> 
> It should be. Do you know anyone who is familiar with that so I can get in touch with them or is there any documentation I can use to get a better feeling on how it works for Anemoi?

Which data file are you working with?

TillHae:
> > > > @clessig I still think the way we load the mean and stdev values can be improved. We should make sure that once they are computed, we save them so subsequent runs can load them directly without recomputing. What do you think is the best way to handle that?
> > > 
> > > 
> > > Ideally, mean and stdev should be determined in a precomputation step. It can then be stored directly in the dataset (as it's done for anemoi and our observation datasets) or alongside it (e.g. on meteocloud). Would this be possible?
> > 
> > 
> > It should be. Do you know anyone who is familiar with that so I can get in touch with them or is there any documentation I can use to get a better feeling on how it works for Anemoi?
> 
> Which data file are you working with?

I work with the German SYNOP data I created (`/p/project1/hclimrep/hauer1/synop/data/datasetv1/`) and the SYNOP data from MetNorway (`/p/project1/weatherai/myhre1/metno_observations_v3.nc`). Both datasets are still on JUWELS.

clessig:
> > > > > @clessig I still think the way we load the mean and stdev values can be improved. We should make sure that once they are computed, we save them so subsequent runs can load them directly without recomputing. What do you think is the best way to handle that?
> > > > 
> > > > 
> > > > Ideally, mean and stdev should be determined in a precomputation step. It can then be stored directly in the dataset (as it's done for anemoi and our observation datasets) or alongside it (e.g. on meteocloud). Would this be possible?
> > > 
> > > 
> > > It should be. Do you know anyone who is familiar with that so I can get in touch with them or is there any documentation I can use to get a better feeling on how it works for Anemoi?
> > 
> > 
> > Which data file are you working with?
> 
> I work with the German SYNOP data I created (`/p/project1/hclimrep/hauer1/synop/data/datasetv1/`) and the SYNOP data from MetNorway (`/p/project1/weatherai/myhre1/metno_observations_v3.nc`). Both datasets are still on JUWELS.

There's no general policy or something like this. The multitude of data readers are there so that we can work with different solutions on a per dataset level (the more homogeneous it is, though, the fewer data readers we need).

TillHae:
> > > > > > @clessig I still think the way we load the mean and stdev values can be improved. We should make sure that once they are computed, we save them so subsequent runs can load them directly without recomputing. What do you think is the best way to handle that?
> > > > > 
> > > > > 
> > > > > Ideally, mean and stdev should be determined in a precomputation step. It can then be stored directly in the dataset (as it's done for anemoi and our observation datasets) or alongside it (e.g. on meteocloud). Would this be possible?
> > > > 
> > > > 
> > > > It should be. Do you know anyone who is familiar with that so I can get in touch with them or is there any documentation I can use to get a better feeling on how it works for Anemoi?
> > > 
> > > 
> > > Which data file are you working with?
> > 
> > 
> > I work with the German SYNOP data I created (`/p/project1/hclimrep/hauer1/synop/data/datasetv1/`) and the SYNOP data from MetNorway (`/p/project1/weatherai/myhre1/metno_observations_v3.nc`). Both datasets are still on JUWELS.
> 
> There's no general policy or something like this. The multitude of data readers are there so that we can work with different solutions on a per dataset level (the more homogeneous it is, though, the fewer data readers we need).

In that case I think the current solution isn't even that bad. We should just standardize the keys so users don't configure them manually leading to many different keys in our datasets. I would suggest to just use "mean" and "stdev". Do you want to still keep the option to configure keys for reading so that the users can change them when the datasets they want to use have specific ones?

PR #1678, opened 2026-01-23
Write forecast output incrementally from develop

### LINKED ISSUE #1596: Forecast output needs to be written incrementally
### What happened?

_No response_

_No response_

ankitpatnala:
## Description

Currently, the logging of output for all the forecast_steps happens at once. Because of which, all predictions are stored and then written.
This memory consumption is linearly proportional to forecat_steps. With increase in number of forecast_steps, it becomes difficult.

## Issue Number

Closes #1596  

[Hedgedoc linked to this issue](https://gitlab.jsc.fz-juelich.de/hedgedoc/uA79XQ2wQUaCgjEO9mAsAA#)


ankitpatnala:
cc @grassesi

config/default_config.yml
```diff
@@ -184,6 +184,7 @@ training_config:
       time_step: 06:00:00
       num_steps: 2
       policy: "fixed"
+      forecast_chunk_size: 10
```
grassesi: since the full name is `forecasting.forecast_chunk_size`, `chunk_size` should be sufficient => `forecasting.chunk_size`

PR #1689, opened 2026-01-24
Clessig/jk/develop/1654 fix fsteps

clessig:
## Description

Cleanups and improvements to target branch.

## Issue Number

-

### DIFF
--- a/packages/common/src/weathergen/common/config.py
+++ b/packages/common/src/weathergen/common/config.py
@@ -738,23 +738,23 @@ def validate_forecast_policy_and_steps(forecast_cfg: OmegaConf, mode: str):
         "than 1 or a non-empty list and all of its elements must be integers greater than 1."
     )
 
-    # check forecast offset
-    if forecast_cfg.get("offset") is not None:
-        assert isinstance(forecast_cfg.offset, int), TypeError(valid_forecast_offset)
-        if forecast_cfg.offset == 0:
-            if isinstance(forecast_cfg.num_steps, int):
-                assert forecast_cfg.num_steps in [0, 1], valid_forecast_steps_offset0
-            else:
-                raise TypeError(valid_forecast_steps_offset0)
-        elif forecast_cfg.offset == 1:
-            assert forecast_cfg.policy, (provide_forecast_policy, valid_forecast_policies)
-            if isinstance(forecast_cfg.num_steps, int):
-                assert forecast_cfg.num_steps > 0, valid_forecast_steps_offset1
-            elif isinstance(forecast_cfg.num_steps, ListConfig) and len(forecast_cfg.num_steps) > 0:
-                assert all(step > 0 for step in forecast_cfg.num_steps), (
-                    valid_forecast_steps_offset1
-                )
-            else:
-                raise TypeError(valid_forecast_steps_offset1)
+    # get output_offset or set default to 0 as in multi_stream_data_sampler.py
+    output_offset = forecast_cfg.get("offset", 0)
+    assert isinstance(output_offset, int), TypeError(valid_forecast_offset)
+    if output_offset == 0:
+        if isinstance(forecast_cfg.num_steps, int):
+            assert forecast_cfg.num_steps in [0, 1], valid_forecast_steps_offset0
         else:
-            raise TypeError(valid_forecast_offset)
+            raise TypeError(valid_forecast_steps_offset0)
+    elif output_offset == 1:
+        assert forecast_cfg.policy, (provide_forecast_policy, valid_forecast_policies)
+        if isinstance(forecast_cfg.num_steps, int):
+            assert forecast_cfg.num_steps > 0, valid_forecast_steps_offset1
+        elif isinstance(forecast_cfg.num_steps, ListConfig) and len(forecast_cfg.num_steps) > 0:
+            assert all(step > 0 for step in forecast_cfg.num_steps), (
+                valid_forecast_steps_offset1
+            )
+        else:
+            raise TypeError(valid_forecast_steps_offset1)
+    else:
+        raise TypeError(valid_forecast_offset)
--- a/packages/common/src/weathergen/common/config.py
+++ b/packages/common/src/weathergen/common/config.py
@@ -715,6 +715,10 @@ def validate_forecast_policy_and_steps(forecast_cfg: OmegaConf, mode: str):
                         if `forecast_step` is negative while `forecast.policy` is provided, or
                         if any of the forecast steps in a list are negative.
     """
+
+    if len(forecast_cfg) == 0:
+        return
+
     provide_forecast_policy = (
         f"'{mode}.forecast.policy' must be specified when '{mode}.forecast.num_steps' is not zero "
         f"and '{mode}.forecast.offset' is 1. "
@@ -733,19 +737,24 @@ def validate_forecast_policy_and_steps(forecast_cfg: OmegaConf, mode: str):
         f"For '{mode}.forecast.offset: 0', '{mode}.forecast.num_steps' must be an integer greater "
         "than 1 or a non-empty list and all of its elements must be integers greater than 1."
     )
-    assert isinstance(forecast_cfg.offset, int), TypeError(valid_forecast_offset)
-    if forecast_cfg.offset == 0:
-        if isinstance(forecast_cfg.num_steps, int):
-            assert forecast_cfg.num_steps in [0, 1], valid_forecast_steps_offset0
+
+    # check forecast offset
+    if forecast_cfg.get("offset") is not None:
+        assert isinstance(forecast_cfg.offset, int), TypeError(valid_forecast_offset)
+        if forecast_cfg.offset == 0:
+            if isinstance(forecast_cfg.num_steps, int):
+                assert forecast_cfg.num_steps in [0, 1], valid_forecast_steps_offset0
+            else:
+                raise TypeError(valid_forecast_steps_offset0)
+        elif forecast_cfg.offset == 1:
+            assert forecast_cfg.policy, (provide_forecast_policy, valid_forecast_policies)
+            if isinstance(forecast_cfg.num_steps, int):
+                assert forecast_cfg.num_steps > 0, valid_forecast_steps_offset1
+            elif isinstance(forecast_cfg.num_steps, ListConfig) and len(forecast_cfg.num_steps) > 0:
+                assert all(step > 0 for step in forecast_cfg.num_steps), (
+                    valid_forecast_steps_offset1
+                )
+            else:
+                raise TypeError(valid_forecast_steps_offset1)
         else:
-            raise TypeError(valid_forecast_steps_offset0)
-    if forecast_cfg.offset == 1:
-        assert forecast_cfg.policy, (provide_forecast_policy, valid_forecast_policies)
-        if isinstance(forecast_cfg.num_steps, int):
-            assert forecast_cfg.num_steps > 0, valid_forecast_steps_offset1
-        elif isinstance(forecast_cfg.num_steps, ListConfig) and len(forecast_cfg.num_steps) > 0:
-            assert all(step > 0 for step in forecast_cfg.num_steps), valid_forecast_steps_offset1
-        else:
-            raise TypeError(valid_forecast_steps_offset1)
-    else:
-        raise TypeError(valid_forecast_offset)
+            raise TypeError(valid_forecast_offset)
--- a/src/weathergen/datasets/batch.py
+++ b/src/weathergen/datasets/batch.py
@@ -100,16 +100,6 @@ class Sample:
         assert self.streams_data.get(stream_name, -1) != -1, "stream name does not exist"
         return self.streams_data[stream_name]
 
-    def get_output_len(self) -> int:
-        for _, sdata in self.streams_data.items():
-            output_len = sdata.get_output_len()
-        return output_len
-
-    def get_forecast_idxs(self) -> int:
-        for _, sdata in self.streams_data.items():
-            forecast_idxs = sdata.get_forecast_idxs()
-        return forecast_idxs
-
 
 class BatchSamples:
     """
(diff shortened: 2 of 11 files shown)


packages/common/src/weathergen/common/config.py
```diff
@@ -733,19 +737,24 @@ def validate_forecast_policy_and_steps(forecast_cfg: OmegaConf, mode: str):
...
-            raise TypeError(valid_forecast_steps_offset0)
-    if forecast_cfg.offset == 1:
-        assert forecast_cfg.policy, (provide_forecast_policy, valid_forecast_policies)
-        if isinstance(forecast_cfg.num_steps, int):
-            assert forecast_cfg.num_steps > 0, valid_forecast_steps_offset1
-        elif isinstance(forecast_cfg.num_steps, ListConfig) and len(forecast_cfg.num_steps) > 0:
-            assert all(step > 0 for step in forecast_cfg.num_steps), valid_forecast_steps_offset1
+
+    # check forecast offset
+    if forecast_cfg.get("offset") is not None:
```
Jubeku: We set a default forecast_offset=0 in the multi_stream_data_sampler - so we should still do the the checks below if ` forecast_cfg.get("offset") == None`.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -98,24 +98,34 @@ def __init__(
...
             # no forecast policy specified so set neutral default for no forecasting
-            self.forecast_delta_dt = np.timedelta64(0, "ms")
-            self.list_num_forecast_steps = [0]
-            self.forecast_offset = 0
+            self.list_num_forecast_steps = np.array([0], dtype=np.int32)
+            self.output_offset = 0
             self.forecast_policy = None
+            self.time_step = np.timedelta64(0, "ms")
+
+        fsm = self.list_num_forecast_steps[0]
```
Jubeku, line 126: Why [0]? `list_num_forecast_steps` is not necessarily sorted.

Jubeku: If I understand correctly, the forecast_steps (if several options are given) are drawn further down for each batch step. So maybe the dataset indices should be reduced by the maximum possible forecast length?

src/weathergen/datasets/multi_stream_data_sampler.py
Jubeku: `timestep_idx` instead of `idx`?

Jubeku: `tidx` instead of `idx`?

src/weathergen/datasets/batch.py
```diff
@@ -118,11 +108,15 @@ class BatchSamples:
 
     samples: list[Sample]
     tokens_lens: torch.Tensor | None
+    output_steps: int
+    output_idxs: list[int]
     device: str | None
 
-    def __init__(self, streams: dict, num_samples: int) -> None:
+    def __init__(self, streams: dict, num_samples: int, output_steps, output_idxs) -> None:
```
Jubeku, line 115: That doesn't fit to how the function is called in the sampler.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -633,7 +628,14 @@ def _get_batch(self, idx: int, num_forecast_steps: int):
...
 
-        batch = ModelBatch(self.streams, num_source_samples, num_target_samples)
+        num_output_steps = self._get_output_length(num_forecast_steps)
+        batch = ModelBatch(
+            self.streams,
+            num_source_samples,
+            num_target_samples,
+            self.output_offset,
+            num_output_steps,
+        )
```
Jubeku, line 638: The arguments are mixed up with regards to the class init function. Also, we don't need offset - it should be the output length (i.e. num_output_steps) and the output_idxs which need to be constructed as I did previously in the stream_data: 
`self.forecast_idxs = [ fs for fs in range(forecast_offset, forecast_offset + num_forecast_steps)]`

Jubeku: (But I like the fact that we are passing this now directly to the batch)

Jubeku (changes requested):

src/weathergen/datasets/batch.py
clessig: > That doesn't fit to how the function is called in the sampler.

This function is only accessed through ModelBatch.

packages/common/src/weathergen/common/config.py
clessig: Which line was this? Still cannot see it in the deleted code above.

src/weathergen/datasets/multi_stream_data_sampler.py
clessig: I didn't rethink the logic here but just moved things to more appropriate places in the function. I think it should also be a separate PR to fix/check this.

src/weathergen/datasets/multi_stream_data_sampler.py
clessig: Yes, good catch

src/weathergen/datasets/multi_stream_data_sampler.py
clessig: No, it's correct: it's ModelBatch and not BatchSamples.

src/weathergen/datasets/multi_stream_data_sampler.py
clessig: Yes

packages/common/src/weathergen/common/config.py
Jubeku: Here: https://github.com/ecmwf/WeatherGenerator/blob/clessig/jk/develop/1654_fix_fsteps/src/weathergen/datasets/multi_stream_data_sampler.py#L107
So you could specify no offset in the config but still specify num_steps: 1

src/weathergen/model/model.py
```diff
@@ -545,11 +549,11 @@ def print_num_parameters(self) -> None:
         ]
         print("-----------------")
 
-    def get_latent_state(self, z, tokens):
+    def tokens_to_latent_state(self, z, tokens) -> LatentState:
         return LatentState(
-            register_tokens=z[:, : self.register_token_idx],
-            class_token=z[:, self.register_token_idx : self.class_token_idx],
-            patch_tokens=z[:, self.class_token_idx :],
+            register_tokens=z[:, self.register_token_idxs] if z is not None else z,
```
Jubeku: Maybe `[...] if z is not None else None` is more readable?

src/weathergen/model/model.py
```diff
@@ -567,65 +571,64 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
...
-            for name, head in self.latent_heads.items():
-                output.add_latent_prediction(0, name, head(latent_state))
-            # prediction for final step
-            output = self.predict(model_params, 0, tokens, batch, output)
+        # roll-out in latent space, iterate over all steps but generate output only as requested
+        output_idxs = batch.get_output_idxs()
+        for step in range(batch.get_output_len()):
+            # apply forecasting engine
+            if self.forecast_engine:
+                tokens = self.forecast_engine(tokens, step)
```
Jubeku, line 590: You always want to call forecast_engine already for step 0 and overwrite the tokens at step 1 when your first output_idx is 1 (aka your first forecast step)?

src/weathergen/model/model.py
```diff
@@ -567,65 +571,64 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
...
+        batch: ModelBatch,
+        output: ModelOutput,
+    ) -> ModelOutput:
+        """
+        Compute latent predictions
         """
 
-        tokens = self.forecast_engine(tokens, fstep)
+        # safe latent prediction
+        tokens_pre_norm = self.latent_pre_norm(tokens) if step == 0 else None
```
Jubeku: Are `tokens_pre_norm` the tokens _post_ the pre-model-LayerNorm?

src/weathergen/model/model.py
```diff
@@ -567,65 +571,64 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
         output = ModelOutput(batch.get_output_len())
 
         tokens, posteriors = self.encoder(model_params, batch)
+        output.add_latent_prediction(0, "posteriors", posteriors)
```
Jubeku, line 579: This will be overwritten when output_idxs contain 0. But maybe that's supposed to be?

src/weathergen/train/target_and_aux_module_base.py
```diff
@@ -98,13 +98,14 @@ def update_state_post_opt_step(self, istep, batch, model, **kwargs):
...
+        output_idxs = batch.get_output_idxs()
+        assert len(output_idxs) > 0
 
-        targets = TargetAuxOutput(batch.get_output_len(), forecast_idxs)
+        targets = TargetAuxOutput(batch.get_output_len(), output_idxs)
 
         # collect all targets, concatenating across batch dimension since this is also how it
         # happens for predictions in the model
-        timestep_idxs = [0] if len(forecast_idxs) == 0 else forecast_idxs
+        timestep_idxs = [0] if len(output_idxs) == 0 else output_idxs
```
Jubeku: That's not needed anymore if we always enforce `len(output_idxs) > 0` (see assertion in line 102)

packages/common/src/weathergen/common/config.py
Jubeku: I now see https://github.com/ecmwf/WeatherGenerator/blob/clessig/jk/develop/1654_fix_fsteps/src/weathergen/train/trainer.py#L126
So I think it should be: `if forecast_cfg is not None:`

src/weathergen/train/target_and_aux_module_base.py
```diff
@@ -98,13 +98,14 @@ def update_state_post_opt_step(self, istep, batch, model, **kwargs):
...
         for stream_name in stream_names:
             # collect targets for all forecast steps
             for t_idx in timestep_idxs:
```
Jubeku: We can directly loop over `output_idxs`.

Jubeku (commented):
`output_idxs` was not only renamed from `forecast_idxs` but also the definition changed so that it always contains all output indices, not only in the case for forecasting. 
`len(output_idxs) == batch.get_output_len()` unless `forecast.offset==1`, then `len(output_idxs) == batch.get_output_len() -1`.

Jubeku:
Running inference I got:
```
num_forecast_steps at mini_epoch=0 : 6
  0%|                                                                                                                                           | 0/16 [00:20<?, ?it/s]
  File "/users/jkuehner/CODE/WeatherGenerator/src/weathergen/run_train.py", line 78, in inference_from_args
    trainer.inference(cf, devices, args.from_run_id, args.mini_epoch)
  File "/users/jkuehner/CODE/WeatherGenerator/src/weathergen/train/trainer.py", line 213, in inference
    self.validate(0, self.test_cfg, self.batch_size_test_per_gpu)
  File "/users/jkuehner/CODE/WeatherGenerator/src/weathergen/train/trainer.py", line 535, in validate
    preds = self.model(
            ^^^^^^^^^^^
  File "/users/jkuehner/CODE/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1739, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/jkuehner/CODE/WeatherGenerator/.venv/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1750, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/jkuehner/CODE/WeatherGenerator/src/weathergen/model/model.py", line 590, in forward
    output = self.predict_decoders(model_params, step, tokens, batch, output)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/users/jkuehner/CODE/WeatherGenerator/src/weathergen/model/model.py", line 649, in predict_decoders
    tokens = tokens[:, (self.aux_token_idxs[-1] + 1) :]
                        ~~~~~~~~~~~~~~~~~~~^^^^
IndexError: list index out of range
[6] > /users/jkuehner/CODE/WeatherGenerator/src/weathergen/model/model.py(649)predict_decoders()
-> tokens = tokens[:, (self.aux_token_idxs[-1] + 1) :]
```

Command:
`uv run inference --from_run_id=eu7fp1nm --samples=16 --options training_config.forecast.num_steps=6 zarr_store=zip`

Jubeku:
...it's because `self.aux_token_idxs = []` is empty, so `self.aux_token_idxs[-1] ` throws an error

src/weathergen/model/model.py
clessig: No, `posteriors` is only written here (`latents` is set in `predict_latent()`).

src/weathergen/model/model.py
clessig: If `output_idxs[0] = 1` and `self.forecast_engine is not None` then this is the correct behavior.  If `output_idxs[0] = 0` and  `self.forecast_engine is None` we also obtain the correct behavior. We could enforce that `output_idxs[0] = 1 and self.forecast_engine is None` is an invalid config option.

src/weathergen/model/model.py
clessig: `tokens_pre_norm` is what was previously called `z`. But the name is wrong, it should be `tokens_post_norm`, yes.

Jubeku:
But then we should iterate over output_idxs instead of range(output_len), no?
________________________________

src/weathergen/datasets/multi_stream_data_sampler.py
Jubeku: Will open a PR on this.

Jubeku:
> ...it's because `self.aux_token_idxs = []` is empty, so `self.aux_token_idxs[-1] ` throws an error

Fixed this by introducing `self.first_patch_token_idx` which replaces `self.aux_token_idxs`.

Jubeku (approved):

PR #1690, opened 2026-01-25
Merged forecast step fixes and SSL fixes

clessig:
## Description

This branch combines the changes in #1610 and #1656 and consolidated them for merging onto develop.

## Issue Number

Closes #1608
Closes #1673
Closes #1654

## Checklist before asking for review

-   [ ] I have performed a self-review of my code
-   [ ] My changes comply with basic sanity checks:
      - I have fixed formatting issues with `./scripts/actions.sh lint`
      - I have run unit tests with `./scripts/actions.sh unit-test`
      - I have documented my code and I have updated the docstrings.
      - I have added unit tests, if relevant
-   [ ] I have tried my changes with data and code:
      - I have run the integration tests with `./scripts/actions.sh integration-test`
      - (bigger changes) I have run a full training and I have written in the comment the run_id(s): `launch-slurm.py --time 60`
      - (bigger changes and experiments) I have shared a hegdedoc in the github issue with all the configurations and runs for this experiments
-   [ ] I have informed and aligned with people impacted by my change:
    - for config changes: the MatterMost channels and/or a design doc
    - for changes of dependencies: the MatterMost software development channel

PR #1727, opened 2026-01-27
[1687][model] Flag for setting eval mode (on develop)

### LINKED ISSUE #1687: Introduce flag to set all frozen modules to `eval` mode

The model currently allows to specify modules to freeze via the `freeze_modules` parameter in the config. However, this does not mean that those parts of the model will act deterministically (e.g. due to dropout). In particular, for an ongoing experiment in the development of the diffusion model we are trying to overfit to a single encoded sample in latent space. This requires that at each iteration the frozen modules reliably output the same latent state. Hence we need a another parameter in the config that enforces not only freezing, but also sets the same modules into `eval()` mode. 

This might be useful more generally, but it should be discussed whether this is a desirable setting for other fine-tuning tasks.

Relates to #1379.

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation

moritzhauschulz:
## Description

This PR introduces the flag described in https://github.com/ecmwf/WeatherGenerator/issues/1687.

## Issue Number

Closes #1687 

### DIFF
--- a/src/weathergen/model/model_interface.py
+++ b/src/weathergen/model/model_interface.py
@@ -70,6 +70,9 @@ def init_model_and_shard(
         if re.fullmatch(cf.freeze_modules, name) is not None:
             logger.info(f"Froze weights {name}")
             freeze_weights(module)
+            module.eval()
+        else:
+            logger.info(f"Did not freeze weights {name}")
 
     # TODO: this should be handled in the encoder to be close where q_cells is defined
     if "q_cells" in cf.freeze_modules:
--- a/src/weathergen/model/model_interface.py
+++ b/src/weathergen/model/model_interface.py
@@ -70,7 +70,6 @@ def init_model_and_shard(
         if re.fullmatch(cf.freeze_modules, name) is not None:
             logger.info(f"Froze weights {name}")
             freeze_weights(module)
-            module.eval()
         else:
             logger.info(f"Did not freeze weights {name}")
 
--- a/src/weathergen/model/model_interface.py
+++ b/src/weathergen/model/model_interface.py
@@ -11,7 +11,6 @@
 
 import itertools
 import logging
-import re
 from pathlib import Path
 
 import omegaconf
@@ -33,10 +32,7 @@ from weathergen.model.attention import (
 from weathergen.model.ema import EMAModel
 from weathergen.model.layers import MLP
 from weathergen.model.model import Model, ModelParams
-from weathergen.model.utils import (
-    freeze_weights,
-    apply_fct_to_blocks
-)
+from weathergen.model.utils import apply_fct_to_blocks, freeze_weights
 from weathergen.train.target_and_aux_module_base import PhysicalTargetAndAux
 from weathergen.train.target_and_aux_ssl_teacher import EMATeacher
 from weathergen.utils.distributed import is_root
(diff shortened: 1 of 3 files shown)


src/weathergen/model/model_interface.py
```diff
@@ -70,6 +70,8 @@ def init_model_and_shard(
         if re.fullmatch(cf.freeze_modules, name) is not None:
             logger.info(f"Froze weights {name}")
             freeze_weights(module)
+        else:
```
clessig: We need to remove this. Otherwise we always have the entire model printed in the output.

src/weathergen/train/trainer.py
```diff
@@ -405,6 +406,15 @@ def train(self, mini_epoch):
         cf = self.cf
         self.model.train()
 
+        # TODO: ablate this vs. just using no_grad
+        for name, module in self.model.named_modules():
+            name = module.name if hasattr(module, "name") else name
+            # avoid the whole model element which has name ''
+            if name == "":
```
clessig: Can we use 

```
re.fullmatch(cf.freeze_modules, name) is not None and name != ""
```

src/weathergen/train/trainer.py
```diff
@@ -405,6 +406,15 @@ def train(self, mini_epoch):
         cf = self.cf
         self.model.train()
 
+        # TODO: ablate this vs. just using no_grad
```
clessig: Remove comment

src/weathergen/train/trainer.py
```diff
@@ -405,6 +406,15 @@ def train(self, mini_epoch):
...
+        for name, module in self.model.named_modules():
```
clessig: This should go to a separate function. Would it be possible to have the same function for the freezing code in model_interface and here, i.e. something like

```
def apply_fct_to_model_blocks( names, fct) : 

    for name, module in model.named_modules():
        name = module.name if hasattr(module, "name") else name
        # avoid the whole model element which has name ''
        if (re.fullmatch(cf.freeze_modules, name) is not None) and (name is not ""):
            logger.info(f"Froze weights {name}")
            fct(module)
```

and fct is then either `freeze_module()` or `module_to_eval()` (we probably find other use cases).

clessig (commented):
Thanks, some smaller things to improve the code quality. Can we also add this patch:

```
diff --git a/src/weathergen/train/trainer.py b/src/weathergen/train/trainer.py
index 5ae1beab..e7b807b9 100644
--- a/src/weathergen/train/trainer.py
+++ b/src/weathergen/train/trainer.py
@@ -288,9 +288,8 @@ class Trainer(TrainerBase):
 
         # if with_fsdp then parameter count is unreliable
         if is_root():
-            if cf.with_fsdp:
-                logger.warning("Trainable parameters are inaccurate with FSDP enabled.")
-            # self.model.print_num_parameters()
+            if not cf.with_ddp:
+                self.model.print_num_parameters()
```

It's something we didn't properly clean up in #1690. But also needed to test the code, in my opinion.

src/weathergen/model/model_interface.py
moritzhauschulz: I will remove this, but actually I think it is quite good to know which modules are not frozen. For example, the config prepared initially for training the diffusion engine was missing some modules that should have been frozen. I only discovered this after printing all modules. That's why I think an exhaustive list is quite useful to avoid unintended behaviour, but perhaps it should be aggregated a bit more concisely (e.g. in a list). If you agree I can open a separate issue.

src/weathergen/train/trainer.py
moritzhauschulz: Okay, that's a good idea.

clessig (approved):
Looks good. Thanks for the `apply_fct_to_blocks()` function.

Issue #1085, opened 2025-10-14
duplicate date times in default inference


iluise:
### What happened?

Im having an issue understanding why running over all samples for the same forecast step I get valid_times that overlap. 
 
 
```
with ZarrIO(fname_zarr) as zio:
   samples = zio.samples
   forecast_step = 2
   times= [] 
   for sample in samples: 
       times.append(np.unique(zio.get_data(sample, stream, forecast_step).prediction.as_xarray().valid_time).squeeze())
len(times) --> 244
np.unique(times) --> 206
```

samples should be the initialization time, so for the same forecast step I would have expected 244 different times.

I launched the default inference and I get the following dates: 
 ```
start date 22-10-01T00 
end date 2022-11-21T18 
 
[...]

PR #1749, opened 2026-01-29
[1438] Adding padding to end_date to avoid duplicate samples

### LINKED ISSUE #1438: Duplicate samples during inference due to different length assumtions in MultiStreamDataReader
### setup
run inference on a model trained with $n_{fstep}$ forecast steps and $n_{samples}*dt\geq t_{end} - t_{start}$.
Where  $n_{fstep}=$`--forecast_steps`, $n_{samples}=$`--samples`, $dt=$`--step_hours`, $t_{start}=$`--start`, $t_{end}=$`--end`
### expected result:
No duplicate samples are generated

### actual result:
`n+1` duplicate samples are generated

As an example inference can be run on [w6khbe9g](https://dbc-080c3210-c159.cloud.databricks.com/ml/experiments/384213844828345/runs/047c6529805942668d28ed437cdc42c9?o=3342786197435636) using:

`uv run inference --samples 244` => will contain 4 duplicate samples

For information on the history of this issue and how to avoid triggering the issue refer to #1085

This code can be used to check if the duplicate samples persist:
```python
from pathlib import Path
import numpy as np
from weathergen.common.io import ZarrIO

RUN_ID = "<MY_INFERENCE_RUN_ID>"
results = Path(f"results/{RUN_ID}/validation_chkpt00000_rank0000.zarr")

with ZarrIO(results) as zio:
  samples = zio.samples
  forecast_step = 0
  times = []
  for sample in samples:
    # one unique datetime per sample
    data = zio.get_data(sample, "ERA5", forecast_step).prediction.as_xarray().valid_time
    print("|", end="")
    times.append(np.unique(data).squeeze())

print()
times = np.array(times)
n_duplication = times.size - np.unique(times).size
print(times.size, np.unique(times).size, n_duplication)
indexes = np.argsort(times)
duplicate_indexes = indexes[:n_duplication*2]
print(f"duplicate indices: {duplicate_indexes}") # randomly distributed
print(f"duplicate times: {times[duplicate_indexes]}")
print(f"min: {times.min()}, max: {times.max()}")
```

enssow:
## Description

TimeWindowHandler doesn't produce enough available forecast initilisation times to choose for samples when run inference on a model trained with $n_{fstep}$ forecast steps and $n_{samples}*dt\geq t_{end} - t_{start}$.
Where  $n_{fstep}=$`--forecast_steps`, $n_{samples}=$`--samples`, $dt=$`--step_hours`, $t_{start}=$`--start`, $t_{end}=$`--end`
(See #1438 and #1085) for more info

This PR provides this padding by working out how many available individual initialisation times there are and adjusting the end of the time window to accomodate that and taking into account the extra time needed to accomodate the number of forecast steps to rollout to

## Issue Number

Closes #1438 #2059


enssow:
Tested on SANTIS for:
- `uv run inference --from-run-id f4duf5ji --samples 254 --streams-output ERA5 --options training_config.forecast.num_steps=5` 
- `uv run inference --from-run-id f4duf5ji --samples 254 --streams-output ERA5`
both now do not return duplication warning and inference_id `cixysv6l` was run to completion with success

ankitpatnala:
Thanks @enssow for handling this issue
I tested the code using 
`srun uv run inference --from-run-id f4duf5ji --samples 10 -start="2022-10-01" -end="2022-10-02" --streams-output ERA5 --options training_config.forecast.num_steps=5`
The code functioned way it has been described. 

But I still do not know what will be a better strategy; should we pad with available dates or decrease the num_samples to defined range date. What if there is no data after the defined end_date. It will throw an error or return empty tensors.

ankitpatnala:
Can you run some inference with this options
`srun uv run inference --from-run-id f4duf5ji --samples 10 -start=2022-10-01 -end=2022-10-02 --streams-output ERA5 --options training_config.forecast.num_steps=10 training_config.forecast.time_step=03:00:00`

I saw some unwanted behaviour there
```
Logging set up. Logs are in ./output/uv3yi4ac
DDP initialization: rank=0, world_size=1
Using adjusted end date 2022-10-03T00:00:00.000000000 instead of 2022-10-02T00:00:00.000000000
TimeWindowHandler: start=2022-10-01T00:00:00.000000000, end=2022-10-03T00:00:00.000000000, len=06:00:00, step=06:00:00`
```

clessig:
@enssow : what is the root cause of the length mismatch?

enssow:
@clessig this is now ready for review, reverted some of the changes per the meeting - should i publicise it on the PR channel or would you prefer to give it a look? thank you in advance!

enssow:
Expected behaviour: Given a specified date range (either given or using the default test_config) if the user requests more samples than are available (calculated using the length of time required to fit samples seperated by `step_timedelta` and the number of forecast steps from the last sample) the multistreamdatasampler will prioritise the specified date range and reduce the number of samples requested, raising a warning for the user. If the number of available samples is non-positive the user either needs to reduce the number of forecast steps or increase the date range and an AssertionError is raising that breaks the code.

enssow:
Tested on SANTIS with varying start, end dates, num_samples and num_steps:
`uv run inference --from-run-id wmway1iv --options training_config.forecast.num_steps=5 test_config.output.num_samples=5 test_config.start_date=2022-10-01 test_config.end_date=2022-10-03`

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -77,68 +77,129 @@ def collect_datasources(stream_datasets: list, idx: int, type: str, rng) -> IORe
...
+        self.time_step = self.forecast_cfg.get("time_step", np.timedelta64(0, "ms"))
+        self.forecast_policy = self.forecast_cfg.get("policy", None)
 
-        self.repeat_data = cf.data_loading.get("repeat_data_in_mini_epoch", False)
+        if isinstance(self.forecast_cfg.num_steps, int):
+            steps = [self.forecast_cfg.num_steps]
+        else:
+            steps = self.forecast_cfg.num_steps
+
+        self.list_num_forecast_steps = np.array(steps, dtype=np.int32)
```
grassesi: This function can be simplified and inlined:

```python
def __init__():
    ...
    forecast_cfg = FORECAST_DEFAULTS | mode_cfg.get("forecast", {})
    self.output_offset = forecast_cfg["offset"]
    self.time_step = forecast_cfg["time_step"]
    self.forecast_policy = forecast_cfg["policy"]
    steps = np.array(forecast_cfg.num_steps, dtype=np.int32).reshape(-1)
    self.list_num_forecast_steps = np.array(steps, dtype=np.int32)
    ...
```
where `FORECAST_DEFAULTS` is defined as module level constant:
```python
FORECAST_DEFAULTS = {
    "offset": 0,
    "time_step": np.timedelta64(0, "ms"),
    "policy": None,
    "num_steps": np.array([0], dtype=np.int32),
}
```

This is shorter, more direct and will avoid creating a new member (`self.forecast_cfg`) and the side-effect heavy method `_init_forecast_cfg`

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -77,68 +77,129 @@ def collect_datasources(stream_datasets: list, idx: int, type: str, rng) -> IORe
...
+        self.healpix_level = cf.healpix_level
+        self.num_healpix_cells = 12 * 4**self.healpix_level
+        self.masker = Masker(cf.healpix_level, stage, self.streams, self.mode_cfg)
+        self.tokenizer = TokenizerMasking(cf.healpix_level, self.masker)
 
-        self.mode_cfg = mode_cfg
-        self.samples_per_mini_epoch = mode_cfg.samples_per_mini_epoch
-        self.shuffle = mode_cfg.shuffle
+        self.forecast_cfg = mode_cfg.get("forecast", {})
+        self._init_forecast_cfg()
```
grassesi: see my comment on `_init_forecast_cfg()`

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -200,57 +261,69 @@ def __init__(
 
                 self.streams_datasets[stream_info["name"]] += [ds]
```
grassesi: Thanks for separating this part out. However it would be cool if the initialized streams_datasets are passed as a return value. There is also a second more subtle side effect of this method: it adds some information to the global stream configs (eg. channel names). Right now this is hard to avoid, but I would like to see it mentioned in the docstring of this method.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -77,68 +77,129 @@ def collect_datasources(stream_datasets: list, idx: int, type: str, rng) -> IORe
...
             logger.info(self.time_window_handler)
+        self.index_range = tw.get_index_range()
 
-        index_range = self.time_window_handler.get_index_range()
-        perms_len = int(index_range.end - index_range.start)
+        # check samples per mini epoch
+        self.samples_per_mini_epoch = mode_cfg.samples_per_mini_epoch
+        self.check_samples()
+        self.calc_baseperms()
+        self._init_stream_datasets(cf)
```
grassesi: `self.streams_datasets = self._init_stream_datasets(cf)` (see my comment on `_init_stream_datasets`)

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -77,68 +77,129 @@ def collect_datasources(stream_datasets: list, idx: int, type: str, rng) -> IORe
...
+        ) // self.step_timedelta
+
+        assert available_samples > 0, (
+            "There is an insufficient date range to \
+accomodate any number of samples or forecast steps"
+        )
+
+        # choose correct num samples
+        if not self.repeat_data and self.samples_per_mini_epoch:
+            if self.samples_per_mini_epoch >= available_samples:
```
grassesi, line 166: These are not comparable quantities: samples_per_mini_epoch includes the additional factor `self.batch_size` whereas available samples is the number of possible init_times == batches. But there can be multiple samples, sampled from one init time.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -77,68 +77,129 @@ def collect_datasources(stream_datasets: list, idx: int, type: str, rng) -> IORe
...
+                logger.warning(
+                    f"There are only {available_samples} available_samples, \
+samples_per_mini_epoch reduced to {available_samples} to avoid repeating data. \
+Set repeat_data_in_mini_epoch to True if this is undesired."
+                )
+                self.samples_per_mini_epoch = available_samples - 1
```
grassesi: see l. 171

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -77,68 +77,129 @@ def collect_datasources(stream_datasets: list, idx: int, type: str, rng) -> IORe
...
+        if not self.repeat_data:
+            assert n_duplicates <= 0
+
+    def calc_baseperms(self):
+        """This calculates the base permutation array and
+        depends on fsm so must be repeated for __init__ and reset"""
+        perms_len = int(self.index_range.end - self.index_range.start)
+        perms_len -= (self.fsm + self.output_offset) * (self.time_step // self.step_timedelta)
+        self.base_perms = np.arange(perms_len)
+
```
grassesi, line 192: This methods only purpose is to provide `self.base_perms`, which itself is only ever used once in `reset()`. returning here would reduce the namespace of this class.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -200,57 +261,69 @@ def __init__(
...
+            self.calc_baseperms()
+
+        perms = self.base_perms.copy()
+
+        # rng changed, repeat if needed
+        if self.repeat_data and len(perms) < self.samples_per_mini_epoch:
+            perms = np.tile(perms, self.samples_per_mini_epoch // len(perms))
+            filler = self.rng.choice(
+                perms,
+                size=self.samples_per_mini_epoch - len(perms),
```
grassesi: see my comment on `calc_baseperms`: perms holds batch indices for the entire mini epoch. The length is therefore off by a factor of `self.batch_size` from `self.samples_per_mini_epoch`.

grassesi (changes requested):
I appreciate very much you breaking up the constructor. However I would like to take the opportunity to also reduce the excessive use of object state to pass around information that can be handled by returns instead. This will make the reset logic much more explicit and easier to follow.
Additionally I think there is a small oversight: When calculating the length of the batch index permutation array and when checking the number of available init_times (and thus number of available batches) you compare against self.samples_per_mini_epoch, which includes the additional factor batch_size.

I have implemented the suggestions here: https://github.com/grassesi/WeatherGenerator/tree/sgrasse/sorcha/1438-fix feel free to merge them if they look good to you. Please tell me if I overlooked anything.

grassesi (approved):

enssow:
Note: adjusted to let samples_per_mini_epoch=1 if available_samples=1, otherwise reutrn error like:
```
uv run inference --from-run-id e3u8p2bn --options test_config.samples_per_mini_epoch=10 test_config.output.num_samples=10 test_config.forecast.num_steps=5 test_config.forecast.offset=1 test_config.start_date=2023-01-10 test_config.end_date=2023-01-12

Samples will be repeated within the time range
0it [00:00, ?it/s]/users/sowens/WeatherGenerator/src/weathergen/train/trainer.py:785: RuntimeWarning: Mean of empty slice
  {np.nanmean(avg_loss)}"""
validation (xjazf2pi) : 000 : 
                        nan


0it [00:00, ?it/s]
FAIL!
```

enssow:
Testing complete, with the fooolowing command `uv run inference --from-run-id e3u8p2bn --options test_config.samples_per_mini_epoch=10 test_config.output.num_samples=10 test_config.forecast.num_steps=5 test_config.forecast.offset=1`
using the extra options of:
| test_config.start_date | test_config.end_date | result                                                                                                                                                       |
| ---------------------- | -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| "2023-01-10"           | "2023-01-11"         | AssertionError: There is an insufficient date range to accomodate any number of samples or forecast steps                                                    |
| "2023-01-10"           | "2023-01-13"         | Sufficient available samples in the time range specified; checked no duplicate samples                                                                       |
| "2023-01-10"           | "2023-01-12"         | There are only 1 available_samples, samples_per_mini_epoch reduced to 1 to avoid repeating data. Set repeat_data_in_mini_epoch to True if this is undesired. |

PR #1752, opened 2026-01-29
Reenable max_num_targets and add shuffle_source and shuffle_target stream config options

### LINKED ISSUE #1751: max_num_targets in stream config has currently no effect
### What happened?

max_num_targets in stream config has currently no effect. It needs to be re-enabled.

_No response_

_No response_

clessig:
## Description

Re-introduce max_num_target stream config option, but at a more appropriate place. Also addding shuffle_source and shuffle_target stream config options.

## Issue Number

Closes #1751 

### DIFF
--- a/src/weathergen/datasets/multi_stream_data_sampler.py
+++ b/src/weathergen/datasets/multi_stream_data_sampler.py
@@ -50,10 +50,10 @@ def collect_datasources(stream_datasets: list, idx: int, type: str, rng) -> IORe
 
     rdatas = []
 
-    # number of points to sub-sample
-    num_subset = -1
-
     for ds in stream_datasets:
+        # number of points to sub-sample
+        num_subset = -1
+
         if type == "source":
             get_reader_data = ds.get_source
             normalize_channels = ds.normalize_source_channels
--- a/src/weathergen/datasets/multi_stream_data_sampler.py
+++ b/src/weathergen/datasets/multi_stream_data_sampler.py
@@ -41,25 +41,33 @@ type StreamName = str
 logger = logging.getLogger(__name__)
 
 
-def collect_datasources(stream_datasets: list, idx: int, type: str) -> IOReaderData:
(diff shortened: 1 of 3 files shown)


src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -41,25 +41,33 @@
...
             get_reader_data = ds.get_target
             normalize_channels = ds.normalize_target_channels
+            num_subset = ds.stream_info.get("max_num_targets", -1)
+            shuffle = ds.stream_info.get("shuffle_target", False)
         else:
             assert False, "invalid value for argument `type`"
 
         # get source (of potentially multi-step length)
-        rdata = get_reader_data(idx).remove_nan_coords()
+        rdata = get_reader_data(idx).shuffle(rng, shuffle, num_subset).remove_nan_coords()
```
shmh40, line 70: I haven't run this yet and need to double check but: let's say we have 2 stream datasets, and the first one is type target and we have set max_num_targets in the stream config for that dataset. Then we are in the first iteration of the for loop, and we go into elif type target and we set num_subset correctly, then this goes to the rdata = get_reader_data and we get the right shuffle with the right subset. If the next stream_datasets in the loop is e.g. "source" dataset, we loop again and go into type == source, where we do not overwrite or set num_subset. Then we go to rdata = get_data_reader and then this runs shuffle func with the num_subset from the previous dataset, which is not intended?

clessig: This problem cannot happen since `target` and `source` are fixed when one iterates over the streams and it's never changed for `source` and set using `get()` in each iteration. But I will still move where `num_subset` is introduced--it's cleaner to reset for each stream.

shmh40: Makes sense. I thought this might be the case, but yes agree looks a bit cleaner to reset it.

shmh40 (approved):

PR #1773, opened 2026-02-02
Do not shard unused parameters

### LINKED ISSUE #1750: Part of network cannot be sharded during loaded if not used

In model_interface.py we are loading the model in load_model().

If I am finteuning a ERA5 model and using ERA5 only as forcing, then when loading the model, the ERA5 specific decoding paraemters do not have sharding parameters and hence cannot be sharded while loading the model. This creates an error.

By adding a check in the parameter loop, we can see if the parameter can be sharded, if not we can load it as the full tensor:
```python
is_model_sharded = cf.with_ddp and cf.with_fsdp
    if is_model_sharded:
        meta_sharded_sd = model.state_dict()
        maybe_sharded_sd = {}
        for param_name, full_tensor in params.items():
            sharded_meta_param = meta_sharded_sd.get(param_name)
            if sharded_meta_param is None:
                print("sharded meta param is None for ", param_name)
                maybe_sharded_sd[param_name] = torch.nn.Parameter(full_tensor)
            else:
                sharded_tensor = distribute_tensor(
                    full_tensor,
                    sharded_meta_param.device_mesh,
                    sharded_meta_param.placements,
                )
                # maybe_sharded_sd[param_name.replace("module.", "")] = nn.Parameter(sharded_tensor)
                maybe_sharded_sd[param_name] = torch.nn.Parameter(sharded_tensor)
``` 

This solves the problem and the model can be loaded as usual again.

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation

kctezcan:
## Description

See #1750 

## Issue Number

Closes #1750 

### DIFF
--- a/src/weathergen/model/model_interface.py
+++ b/src/weathergen/model/model_interface.py
@@ -189,13 +189,17 @@ def load_model(cf, model, device, run_id: str, mini_epoch=-1):
         maybe_sharded_sd = {}
         for param_name, full_tensor in params.items():
             sharded_meta_param = meta_sharded_sd.get(param_name)
-            sharded_tensor = distribute_tensor(
-                full_tensor,
-                sharded_meta_param.device_mesh,
-                sharded_meta_param.placements,
-            )
-            # maybe_sharded_sd[param_name.replace("module.", "")] = nn.Parameter(sharded_tensor)
-            maybe_sharded_sd[param_name] = torch.nn.Parameter(sharded_tensor)
+            if sharded_meta_param is None:
+                logger.info(f"Sharding meta parameters is None for: {param_name}")
+                maybe_sharded_sd[param_name] = torch.nn.Parameter(full_tensor)
+            else:
+                sharded_tensor = distribute_tensor(
+                    full_tensor,
+                    sharded_meta_param.device_mesh,
+                    sharded_meta_param.placements,
+                )
+                # maybe_sharded_sd[param_name.replace("module.", "")] = nn.Parameter(sharded_tensor)
+                maybe_sharded_sd[param_name] = torch.nn.Parameter(sharded_tensor)
         # choose `assign=True` for sharded model since we cannot call `copy_` on meta tensor
         mkeys, ukeys = model.load_state_dict(maybe_sharded_sd, strict=False, assign=True)
 


shmh40 (approved):
Tested and doesn't cause problems for me.

src/weathergen/model/model_interface.py
```diff
@@ -189,13 +189,17 @@ def load_model(cf, model, device, run_id: str, mini_epoch=-1):
...
             sharded_meta_param = meta_sharded_sd.get(param_name)
-            sharded_tensor = distribute_tensor(
-                full_tensor,
-                sharded_meta_param.device_mesh,
-                sharded_meta_param.placements,
-            )
-            # maybe_sharded_sd[param_name.replace("module.", "")] = nn.Parameter(sharded_tensor)
-            maybe_sharded_sd[param_name] = torch.nn.Parameter(sharded_tensor)
+            if sharded_meta_param is None:
+                logger.info(f"Sharding meta parameters is None for: {param_name}")
```
clessig, line 193: Is it correct that `sharded_meta_param is None` means that this is a parameter in the checkpoint that is not present in the current model?

src/weathergen/model/model_interface.py
```diff
@@ -189,13 +189,17 @@ def load_model(cf, model, device, run_id: str, mini_epoch=-1):
...
+                maybe_sharded_sd[param_name] = torch.nn.Parameter(full_tensor)
+            else:
+                sharded_tensor = distribute_tensor(
+                    full_tensor,
+                    sharded_meta_param.device_mesh,
+                    sharded_meta_param.placements,
+                )
+                # maybe_sharded_sd[param_name.replace("module.", "")] = nn.Parameter(sharded_tensor)
+                maybe_sharded_sd[param_name] = torch.nn.Parameter(sharded_tensor)
```
clessig, line 202: Can we please remove the line below.

clessig (changes requested):

clessig:
@kctezcan : can you address the comments so that we can merge this

src/weathergen/model/model_interface.py
kctezcan: practically yes: it means that the parameter is not in the created model since it does not have sharding meta data but it is present somewhere (in this case in the checkpoint). 

it could happen that a parameter is added after the model is created and is not in the list of model parameters, so it does not have sharding meta data, this condition could be true as well. but this would be an unpractical edge case, I believe.

src/weathergen/model/model_interface.py
kctezcan: the comment line "# choose `assign=True` for sharded model since we cannot call `copy_` on meta tensor"

sure, removing it

src/weathergen/model/model_interface.py
clessig: The first case is relevant for us since we have it when we change the streams for fine-tuning. The code works correctly in this case? Could you test please

src/weathergen/model/model_interface.py
clessig: I think I meant this line since it's stale code:

```
# maybe_sharded_sd[param_name.replace("module.", "")] = nn.Parameter(sharded_tensor)
```

sophie-xhonneux:
@kctezcan What is the status on this?

PR #1783, opened 2026-02-03
passing parameters to metrics

### LINKED ISSUE #1475: pass user parameters to individual scores in the evaluation pipeline

Some scores like ets, fbi, pss, etc. depend on external parameters, for example a threshold. Ideally, users should be able to put such parameters in the evaluation config file and the code should propagate them to the appropriate scores. One use case would be users trying to evaluate extreme precipitation, say above 20mm, using ets. 

A simple implementation would pass each parameter to every score that accepts a parameter of that name, otherwise one would have to somehow enable parameters for individual scores in the config file. 

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation

s6sebusc:
## Description

Metrics in the evaluation config can now be either strings or dicts with parameters:
```
evaluation:
  metrics:
    - fbi:
        thresh: 280
    - rmse
... 
```
When a metric is loaded or computed, corresponding parameters are passed to the corresponding score function. 

New JSON files for scores now have a top level key "scores", below which is a list of results for the same score with different parameter settings, with each list entry following the same format as before. Old files with no "scores" key are treated appropriately. 

Upon loading a score, the reader goes through all elements of the JSON file in search of the right parameter settings. When saving scores, `metric_list_to_json` now checks whether the output file is already present. If it is, it goes through the file in search of the right parameters settings. If it finds them (possible when for example fsteps was changed), it replaces that entry with the new values. If the right parameters are not found, the new results are appended at the end of the JSON file.

## Issue Number

Closes [#1475](https://github.com/ecmwf/WeatherGenerator/issues/1475)


s6sebusc:
Map plots with parameters are working, mlflow probably as well but I couldn't test it.

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -151,7 +151,7 @@ def get_channels(self, stream: str) -> list[str]:
         return all_channels
 
     def load_scores(
-        self, stream: str, regions: list[str], metrics: list[str]
+        self, stream: str, regions: list[str], metrics: list[str], metric_parameters: dict = {}
```
iluise: can we merge metrics and metric_parameters here? 

metrics can be either a dictionary, or a list[str]

`metrics_dict: list[str] | dict[str]`

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -209,12 +211,16 @@ def load_single_score(self, stream: str, region: str, metric: str) -> xr.DataArr
...
         )
         _logger.debug(f"Looking for: {score_path}")
+        score = None
         if score_path.exists():
             with open(score_path) as f:
                 data_dict = json.load(f)
-                score = xr.DataArray.from_dict(data_dict)
-        else:
-            score = None
+                if not "scores" in data_dict:
```
iluise: does this break backward compatibility with the old jsons?

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -253,6 +259,7 @@ def __init__(
         private_paths: dict | None = None,
         regions: list[str] | None = None,
         metrics: list[str] | None = None,
+        metric_parameters: dict = {}
```
iluise: again, here I'd put everything into metrics

packages/evaluate/src/weathergen/evaluate/utils/utils.py
```diff
@@ -60,6 +60,7 @@ def calc_scores_per_stream(
     regions: list[str],
     metrics_dict: dict,
```
iluise, line 68: Here we should use `metrics_dict` instead, which is already a dictionary such as:
```
metrics_dict
{'global': ['fbi']}
```

packages/evaluate/src/weathergen/evaluate/utils/utils.py
```diff
@@ -452,6 +455,7 @@ def metric_list_to_json(
     stream: str,
     metrics_dict: list[xr.DataArray],
     regions: list[str],
+    metric_parameters: dict,
```
iluise: same as above

packages/evaluate/src/weathergen/evaluate/utils/utils.py
```diff
@@ -490,10 +495,25 @@ def metric_list_to_json(
...
-
-                _logger.info(f"Saving results to {save_path}")
+                if save_path.exists():
+                    _logger.info(f"{save_path} already present")
+                    with save_path.open("r") as f: 
+                        data_dict = json.load(f)
+                        if not "scores" in data_dict:
+                            data_dict = { "scores": [data_dict] }
+                        for i,score_version in enumerate(data_dict["scores"]):
+                            if score_version["attrs"] == parameters:
```
iluise: uhm, can you make a concrete example of when this overwriting is used?
It seems dangerous, but maybe I am just not understanding the logic :)

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -193,6 +194,7 @@ def _process_stream(
     regions: list[str],
     metrics: list[str],
     plot_score_maps: bool,
+    metric_parameters: dict[str, object],
```
iluise: same as above, integrate with metrics

iluise (commented):
Thanks a lot for working on this! 

For the config I think we can simplify metrics and metrics_parameters in one entry with  something like:
```
metrics:
  - fbi:
      thresh: 0.5
  - rmse
  - l1
  - blah
```

so you have something like:

```
metrics
[{'fbi': {'thresh': 0.5}}, 'rmse', 'l1', 'blah']
```

and you don't need metrics_parameters at all. Then one can extract the metric and the parameters in the loop(s), but in general all is contained in metrics and there's no need to pass additional parameters to the functions. 

What do you think?

packages/evaluate/src/weathergen/evaluate/utils/utils.py
s6sebusc: If you request a metric that has already been computed but not for the same set of fsteps or samples, I think the current behavior is to recompute and replace the JSON file? Accordingly, I would now overwrite that entry in the existing JSON file, which may contain some other configurations of the same metric as well.

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
s6sebusc: this is supposed to restore compatibility with old jsons. Old files don't have the top level key "scores", so I just add that here and proceed as for the new files.

s6sebusc:
> Thanks a lot for working on this!
> 
> For the config I think we can simplify metrics and metrics_parameters in one entry with something like:
> 
> ```
> metrics:
>   - fbi:
>       thresh: 0.5
>   - rmse
>   - l1
>   - blah
> ```
> 
> so you have something like:
> 
> ```
> metrics
> [{'fbi': {'thresh': 0.5}}, 'rmse', 'l1', 'blah']
> ```
> 
> and you don't need metrics_parameters at all. Then one can extract the metric and the parameters in the loop(s), but in general all is contained in metrics and there's no need to pass additional parameters to the functions.
> 
> What do you think?

Agreed, that looks much nicer. I made it parse the metrics from the config such that it will internally always have something like `{'fbi': {'thresh': 0.5}, 'rmse':{}, ... }`, then we just always loop over that dict.

packages/evaluate/src/weathergen/evaluate/utils/utils.py
```diff
@@ -552,10 +552,25 @@ def metric_list_to_json(
                     reader.metrics_dir
                     / f"{run_id}_{stream}_{region}_{metric}_chkpt{reader.mini_epoch:05d}.json"
                 )
-
-                _logger.info(f"Saving results to {save_path}")
+                if save_path.exists():
```
iluise, line 557: these lines can be maybe refactored a bit for readability?

```
metric_dict = metric_data.to_dict()

if save_path.exists():
    _logger.info(f"{save_path} already present")

    with save_path.open("r") as f:
        data_dict = json.load(f)

    # Normalize structure
    scores = data_dict.get("scores")
    if scores is None:
        scores = [data_dict]
        data_dict = {"scores": scores}

    # Try to replace existing metric with same attrs
    for i, existing_score in enumerate(scores):
        if existing_score["attrs"] == metric_data.attrs:
            _logger.warning("Metric with same parameters found, replacing")
            scores[i] = metric_dict
            break
    else:
        scores.append(metric_dict)
        _logger.info(f"Appending results to {save_path}")

else:
    _logger.info(f"Saving results to new file {save_path}")
    data_dict = {"scores": [metric_dict]}
```

iluise (commented):
Looks great. I added one comment about style, then it is good to go for me. Tested it and it works as expected.

packages/evaluate/src/weathergen/evaluate/utils/utils.py
s6sebusc: yea that looks better, I made two small changes to your suggestion:
- metric_dict sounds almost exactly like metrics_dict which was already there so I said `metric_data_dict = metric_data.to_dict()` instead.
- I think `scores = [data_dict];   data_dict = {"scores": scores}` looks suspicious, like it might create infinite recursion (it doesn't really but I think it looks weird) so I slightly rephrased that.

iluise (approved):

PR #1799, opened 2026-02-04
changed model path to shared dir path

### LINKED ISSUE #1790: Integration test fails to run inference
### What happened?

Expected Result: Running the integration test the inference part should run through

Actual Result: The inference part errors with model_path not found in the configuration file.

```
  File "/.../weathergenerator/WeatherGenerator/src/weathergen/run_train.py", line 79, in inference_from_args
    trainer.inference(cf, devices, args.from_run_id, args.mini_epoch)
  File "/.../weathergenerator/WeatherGenerator/src/weathergen/train/trainer.py", line 200, in inference
    self.model, self.model_params = init_model_and_shard(
                                    ^^^^^^^^^^^^^^^^^^^^^
  File "/.../weathergenerator/WeatherGenerator/src/weathergen/model/model_interface.py", line 159, in init_model_and_shard
    model = load_model(cf, model, device, run_id_contd, mini_epoch_contd)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.../weathergenerator/WeatherGenerator/src/weathergen/model/model_interface.py", line 176, in load_model
    path_run = Path(cf.model_path) / run_id
                    ^^^^^^^^^^^^^
  File "/.../weathergenerator/WeatherGenerator/.venv/lib/python3.12/site-packages/omegaconf/dictconfig.py", line 355, in __getattr__
    self._format_and_raise(
  File "/.../weathergenerator/WeatherGenerator/.venv/lib/python3.12/site-packages/omegaconf/base.py", line 231, in _format_and_raise
    format_and_raise(
  File "/.../weathergenerator/WeatherGenerator/.venv/lib/python3.12/site-packages/omegaconf/_utils.py", line 899, in format_and_raise
    _raise(ex, cause)
  File "/.../weathergenerator/WeatherGenerator/.venv/lib/python3.12/site-packages/omegaconf/_utils.py", line 797, in _raise
    raise ex.with_traceback(sys.exc_info()[2])  # set env var OC_CAUSE=1 for full trace
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.../weathergenerator/WeatherGenerator/.venv/lib/python3.12/site-packages/omegaconf/dictconfig.py", line 351, in __getattr__
    return self._get_impl(
           ^^^^^^^^^^^^^^^
  File "/.../weathergenerator/WeatherGenerator/.venv/lib/python3.12/site-packages/omegaconf/dictconfig.py", line 442, in _get_impl
    node = self._get_child(
           ^^^^^^^^^^^^^^^^
  File "/.../weathergenerator/WeatherGenerator/.venv/lib/python3.12/site-packages/omegaconf/basecontainer.py", line 73, in _get_child
    child = self._get_node(
            ^^^^^^^^^^^^^^^
  File "/.../weathergenerator/WeatherGenerator/.venv/lib/python3.12/site-packages/omegaconf/dictconfig.py", line 480, in _get_node
    raise ConfigKeyError(f"Missing key {key!s}")
omegaconf.errors.ConfigAttributeError: Missing key model_path
    full_key: model_path
    object_type=dict
[11] > /.../weathergenerator/WeatherGenerator/.venv/lib/python3.12/site-packages/omegaconf/dictconfig.py(480)_get_node()
-> raise ConfigKeyError(f"Missing key {key!s}")
```

HPC: JUWELS (prop. others except alps-clariden / santis)

_No response_

sbAsma:
## Issue Number

Closes #1790 


packages/common/src/weathergen/common/config.py
```diff
@@ -232,7 +232,7 @@ def load_run_config(run_id: str, mini_epoch: int | None, model_path: str | None)
         if model_path is None:
             path = get_path_model(run_id=run_id)
         else:
-            path = Path(model_path) / run_id
+            path = _get_shared_wg_path() / "models" / run_id
```
TillHae: There is already a [method](https://github.com/ecmwf/WeatherGenerator/blob/develop/packages/common/src/weathergen/common/config.py#L642) for this.
I think we shouldn't change this. It is used when an explicit `model_path` was given. Otherwise the if-else block is redundant and we can just use the part in the if-block.

src/weathergen/model/model_interface.py
```diff
@@ -179,7 +178,7 @@ def load_model(cf, model, device, run_id: str, mini_epoch=-1):
         mini_epoch : The mini_epoch to load. Default (-1) is the latest mini_epoch
     """
 
-    path_run = Path(cf.model_path) / run_id
+    path_run = _get_shared_wg_path() / "models" / run_id
```
TillHae: Same applies here. We could use [this](https://github.com/ecmwf/WeatherGenerator/blob/develop/packages/common/src/weathergen/common/config.py#L642) method directly. Maybe we can add the same if-else block as in the previous case. This enables the user to not add a `model_path` to the config and just use the regular path. But I'm not entirely sure if this is possible here. Maybe someone with more experience on the model and training process knows better. @sophie-xhonneux ?

TillHae (changes requested):

packages/common/src/weathergen/common/config.py
sbAsma: Indeed the `else` is redundant, I will revert and add a logging instead

sbAsma:
Thank you @TillHae for taking the time to review this PR, I added the requested changes ^^

src/weathergen/model/model_interface.py
```diff
@@ -179,7 +178,7 @@ def load_model(cf, model, device, run_id: str, mini_epoch=-1):
         mini_epoch : The mini_epoch to load. Default (-1) is the latest mini_epoch
     """
 
-    path_run = Path(cf.model_path) / run_id
```
TillHae: Let's use the same if-else block from `config.py`. If the user enters a `model_path` we should use it instead of creating it from the `run_id`.

TillHae (changes requested):

src/weathergen/model/model_interface.py
clessig: If there's shared code that needs to be identical in two places, can we please write a function that ensures it's indeed identical.

sbAsma: > Let's use the same if-else block from `config.py`. If the user enters a `model_path` we should use it instead of creating it from the `run_id`.

Hi @TillHae, I changed it to https://github.com/ecmwf/WeatherGenerator/pull/1799/changes#diff-2e85cd26b59453d8cbf3b54ed76e4cf092f911b8063dde53addf8ea81f0ca28dR183

Please let me know if this is what you meant or not.

TillHae: @sbAsma this is exactly what I meant. I think this should also fix the error from #1790. But I still have three small suggestions.
- define a method for this in `config.py`, which we can then use in both places as @clessig requested. In that case we also don't need the `get_model_path` import anymore.
- remove the additional variable `user_defined_path` and just use `path_run` as in `config.py`
- Remove the space before `.exists()` in l. 185

TillHae (changes requested):

src/weathergen/model/model_interface.py
sbAsma: I moved the routine to `config.py` and re-used the method `get_model_path()`. Lemme know if this works.

packages/common/src/weathergen/common/config.py
TillHae, line 644: Can you take a look on this. Maybe this is already everything we need.

src/weathergen/model/model_interface.py
```diff
@@ -179,7 +178,10 @@ def load_model(cf, model, device, run_id: str, mini_epoch=-1):
...
+    # If `run_id` is a full filesystem path to a model run directory, use it
+    # directly. This matches the behaviour of `config.load_run_config`, which
+    # accepts full-path run_id values.
```
TillHae: It looks like this is no longer true. `get_path_model` doesn't handle the case of `run_id` being a full filesystem path.

packages/common/src/weathergen/common/config.py
```diff
@@ -646,6 +648,22 @@ def get_path_model(config: Config | None = None, run_id: str | None = None) -> P
     else:
         msg = f"Missing run_id and cannot infer it from config: {config}"
         raise ValueError(msg)
+
+    # If a user provided `model_path` in the config, prefer it when the
+    # directory for the requested run exists. Otherwise fall back to the
+    # shared models directory.
+    if config is not None:
```
TillHae: I think adding this to `get_path_model` is the wrong idea. We should create an entirely new method, which uses `get_path_model`. Then we can apply it to `config.py` ll. 232-237 and `model_interface.py` l. 184, because that is the actual same code that we have.

TillHae (changes requested):

src/weathergen/model/model_interface.py
sbAsma: as far as I know, and correct me if I am wrong, `run_id` isn't supposed to be a full system path. @grassesi What do you think?

grassesi: If I remember correctly, for some time in the beginning there was a wish that one can load a model directly by specifying a path. Since our infrastructure has matured quite a bit since then, I feel like it is not required anymore, but I dont know if this option is still in use. @clessig do you know if this is still actively used? can we go ahead and remove this?

sophie-xhonneux: Still good to have, especially with git worktrees or when setting up people during a hackathon or locally. Please don't remove.

clessig: I don't think it should be possible that run_id is a model path. The better solution is that we allow for model_path, result_path (and potential slurm_path) to be overwritten in the CLI.

packages/common/src/weathergen/common/config.py
sbAsma: I don't think a new method is needed, `get_path_model()` already does the job

TillHae (approved):
Looks good. Let's just remove the misleading comment and then we can merge the changes.

sbAsma:
> Looks good. Let's just remove the misleading comment and then we can merge the changes.

Oh yeah, you're right!

tjhunter (approved):

sophie-xhonneux (changes requested):
Can we please leave a summary at the end of the PR because there are a number of open discussions in the comments?

src/weathergen/model/model_interface.py
sophie-xhonneux: We should still be able to load a model directly from a path

sbAsma: @clessig is there a way we can make everyone happy?

clessig: > Still good to have, especially with git worktrees or when setting up people during a hackathon or locally. Please don't remove.

@sophie-xhonneux : what is your use case exactly? I don't understand yet why you don't use the default locations with git worktrees.

In the long run, the option to specify `model_path`, `result_path` on the command line would perhaps be useful.

MatKbauer: @clessig's suggestions looks good to me, i.e., having options to specify `model_path` and `result_path` via the command line. Both paths should be the basename only, though, where `run_id` will be added accordingly, as we pass it anyway with `--run-id {RUN_ID}`

tjhunter:
There are 2 issues right now with this PR:
- there is a conflict
- the logic in config.py:get_path_model does not work as expected: it should also check that the config is provided when looking for a run_id. Right now, it just ignores the config being passed.

I just tried `uv run train --options path_model=my_test_path` and it still uses `./results` . It should use `my_test_path` .

We can have a quick chat if it is not clear. I believe it is a small but important fix, happy to help unblocking the fix here.

sbAsma:
Hi @tjhunter and @sophie-xhonneux 

I added the possibility to overwrite model path. However, when I run `uv run train --options path_model=my_test_path` it doesn't get overwritten. Any suggestions?

EDIT: I found that 1) the custom model path is indeed being used. 2) The config that gets printed shows the default model path. I am still confused, and would appreciate help.

EDIT 2: The overwriting now works.

PR #1856, opened 2026-02-17
[1855][eval] Add Europe as a region for evaluation

### LINKED ISSUE #1855: Add europe region for plotting

During plotting, I'd like to conveniently choose Europe as a region to showcase, for example, the effect of the alpine orography on surface temperature predictions.

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation

MatKbauer:
## Description

An entry for region Europe with according latlon coordinates was added to the `regions.py` file such that it can be selected conveniently when evaluating a model.

## Issue Number

Closes #1855


SavvasMel:
@MatKbauer if you can add a plot of how this looks like that would be prefect just for reference reasons.

SavvasMel (approved):

PR #1860, opened 2026-02-17
[1849][model] Add optional latent Zarr writer

### LINKED ISSUE #1849: Add support for exporting latent space from WeatherGenerator

**Description**
We need functionality to export the model’s latent space during inference.

This is required for downstream applications that rely on correlations in the latent space (in particular the Statkraft and Buluttan use cases). At MetNor, we already have an operational pipeline where we can plug in the WeatherGenerator and run it several times per day. The plan is to expose the latent space via a JSON API for experimentation and validation.

**Proposed Approach**
- Export latent space in Zarr format
  - Consistent with decoder outputs
  - Scalable for long rollouts
  - Supports multiple model initializations
- Writing of latent tensor should likely occur in:
  - ForecastingEngine.forward()
- File creation and metadata handling should be separated (location TBD).

@clessig @iluise 

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation

evenmn:
## Description
This PR introduces a writer for the latent vector. To avoid additional config options, I decided to add "latent" as a special case of a stream output, e.g:

`uv run inference ... --options test_config.output.streams=[ERA5,latent]`

The latent output contains the latent vector itself, metadata to map healpix cells to coordinates, as well as extra features which may be useful for explainable AI applications. The combined ERA5 and latent output file takes the form (here for healpix level 5 and 2048 latent channels):

```
├── 0                                                                                                                                                                                                              
│   ├── ERA5                                                                                                                                                                                                       
│   │   ├── 0                                                                                                                                                                                                      
│   │   │   └── source                                                                                                                                                                                             
│   │   │       ├── coords (40320, 2) float32
│   │   │       ├── data (40320, 2) float32
│   │   │       ├── geoinfo (40320, 0) float32
│   │   │       └── times (40320,) datetime64
│   │   ├── 1
│   │   │   ├── prediction
│   │   │   │   ├── coords (40320, 2) float32
│   │   │   │   ├── data (40320, 1, 1) float32
│   │   │   │   ├── geoinfo (40320, 0) float32
│   │   │   │   └── times (40320,) datetime64
│   │   │   └── target
│   │   │       ├── coords (40320, 2) float32
│   │   │       ├── data (40320, 1) float32
│   │   │       ├── geoinfo (40320, 0) float32
│   │   │       └── times (40320,) datetime64
...
│   └── latent
│          ├── 1                                                                                                                                                                                                      
│          │   ├── coords (12288, 2) float32                                                                                                                                                                          
│          │   ├── geoinfo (12288, 0) float32                                                                                                                                                                         
│          │   ├── tokens (12288, 2048) float32                                                                                                                                                                 
│          │   ├── class_token (0, 2048) float32                                                                                                                                                         
│          │   ├── register_tokens (0, 2048) float32                                                                                                                                                     
│          │   └── times (12288,) datetime64 
...
```

Going further, we can visualize the latent vector:

`unzip validation_chkpt00000_rank0000.zip -d test.zarr`

```python
import zarr
import cartopy.crs as ccrs
import matplotlib.pyplot as plt

z = zarr.open("test.zarr", mode="r", zarr_version=3)

coords = z["0/latent/0/coords"][:]
latent = z["0/latent/0/latent_state"][:]

lat = coords[:,0]
lon = coords[:,1]
values = latent[:,0]

fig = plt.figure(figsize=(12,6))
ax = plt.axes(projection=ccrs.PlateCarree())

sc = ax.scatter(lon, lat, c=values, s=10, transform=ccrs.PlateCarree())

ax.coastlines()
plt.colorbar(sc, label="latent feature 0")

plt.savefig("latent_feature_0.png", dpi=300)
```
Healpix level 4:
<img width="3600" height="1800" alt="latent_feature_0" src="https://github.com/user-attachments/assets/2eba2f75-d42b-479a-853a-947f8a746fbc" />

Healpix level 5:
![latent_time](https://github.com/user-attachments/assets/94db7eac-18ad-438e-bad0-f6b2a4237dd3)


## Issue Number
Closes #1849


src/weathergen/utils/validation_io.py
```diff
@@ -116,7 +117,18 @@ def write_output(
...
         output_stream_names = stream_names
 
-    output_streams = {name: stream_names.index(name) for name in output_stream_names}
+    # Allow a pseudo-stream name 'latent' to enable latent writing while
+    # skipping it from the physical streams mapping.
+    latent_stream_name = None
+    output_streams = {}
+    for name in output_stream_names:
+        if name == "latent":
+            latent_stream_name = name
```
clessig: I don't understand the logic here. Wouldn't it be enough to have 

```
if "latent" in output_stream_names
```

in l 158? Do we expect to have have multiple latent states?

src/weathergen/utils/validation_io.py
```diff
@@ -141,6 +153,27 @@ def write_output(
...
+        for t_idx in timestep_idxs:
+            latents_all.append([])
+            latent_pred = model_output.get_latent_prediction(t_idx)
+            # latent_pred is a dict mapping latent_name -> tensor or LatentState
+            n_samples = len(sample_idxs)
+            for i_sample in range(n_samples):
+                per_sample = {}
+                for lname, lval in latent_pred.items():
+                    if isinstance(lval, LatentState):
+                        for field in ("z_pre_norm", "patch_tokens", "register_tokens", "class_token"):
```
clessig: The latent state that should be relevant for the output are the patch_tokens. These are used for the decoder. To be fully future proof we could have an argument which part of LatentState is written although it might be over-engineering.

src/weathergen/utils/validation_io.py
```diff
@@ -141,6 +153,27 @@ def write_output(
     source_windows = (twh.window(idx) for idx in sample_idxs)
     source_intervals = [TimeRange(window.start, window.end) for window in source_windows]
 
+    # collect latent outputs per forecast step and per sample (optional)
+    latents_all = []
+    if latent_stream_name is not None:
```
clessig: This should go to a separate function.

clessig:
@evenmn : can you please make sure the linter and unit tests pass.

clessig:
@grassesi : can you have a look at the changes in io.py. How far are you with refactoring the output writing?

packages/common/src/weathergen/common/io.py
```diff
@@ -608,6 +615,14 @@ def items(self) -> typing.Generator[OutputItem, None, None]:
...
             yield self.extract(ItemKey(int(s), int(fo_s), fi_s))
 
+        # additionally yield latent output items if a latent stream name was provided
+        if self.latent_stream_name is not None and self.latents:
+            for s, fo_s in itertools.product(self.samples, self.forecast_steps):
+                key = ItemKey(int(s), int(fo_s), self.latent_stream_name)
+                latent_item = self._make_latent_item(key)
+                if latent_item is not None:
+                    yield latent_item
+
```
grassesi: Please try to wrap this logic into the above for iteration: having two yielding for loops works but is confusing at best. I see two alternatives:
1. Mix the writing of latent items into the normal loop. Something like
```python
for s, fo_s, fi_s in itertools.product(
        self.samples, self.forecast_steps, self.streams.keys()
):
    key = ItemKey(int(s), int(fo_s), fi_s)
    if fi_s == LATENT_STREAM:
        latent_item = self._make_latent_item(key)
        if latent_item is not None:
            yield latent_item
    else:
        yield self.extract(ItemKey(int(s), int(fo_s), fi_s))
```
2. have the writing of latent items in a separate method:
```python
def latent_items(self):
    if self.latents:
        for s, fo_s in itertools.product(self.samples, self.forecast_steps):
            key = ItemKey(int(s), int(fo_s), LATENT_STREAM)
            latent_item = self._make_latent_item(key)
            if latent_item is not None:
                yield latent_item

...
with zarrio_writer(config.get_path_results(cf, mini_epoch)) as zio:
    for subset in data.items():
        zio.write_zarr(subset)
    for latent in data.latent_items():
        zio.write_zarr(latent)
```

Option 2. is maybe a bit more clearer and more equivalent to the current solution and also provides more flexibility for the future. But Option 1 would be also fine with me.

src/weathergen/utils/validation_io.py
```diff
@@ -116,7 +117,18 @@ def write_output(
...
 
-    output_streams = {name: stream_names.index(name) for name in output_stream_names}
+    # Allow a pseudo-stream name 'latent' to enable latent writing while
+    # skipping it from the physical streams mapping.
+    latent_stream_name = None
+    output_streams = {}
+    for name in output_stream_names:
+        if name == "latent":
+            latent_stream_name = name
+            continue
```
grassesi: Please use a named constant (eg. `LATENT_STREAM`) that is defined in io.py. Add this constant to the "known names" in `stream_names` and no additional logic is needed.

src/weathergen/utils/validation_io.py
```diff
@@ -153,8 +191,10 @@ def write_output(
         target_channels,
         source_channels,
         geoinfo_channels,
-        sample_start,
-        forecast_offset,
+        latents=latents_all,
+        latent_stream_name=latent_stream_name,
+        sample_start=sample_start,
+        forecast_offset=forecast_offset,
```
grassesi, line 187: If a constant is used for the latent stream name (see my comments above), this is only one new argument.

packages/common/src/weathergen/common/io.py
```diff
@@ -583,8 +583,15 @@ class OutputBatchData:
...
     geoinfo_channels: list[list[str]]
 
-    sample_start: int
-    forecast_offset: int
+    # latent outputs: outer list over forecast steps, inner list over samples.
+    # each entry is a dict mapping latent_name -> ndarray
+    latents: list[list[dict]]
+
+    # optional name to use for latent pseudo-stream when yielding latent items
+    latent_stream_name: str | None = None
```
grassesi: Please use a named constant for this.

grassesi (changes requested):
Overall I like the idea of treating latent output just as another stream. One thing that needs to be addressed is that there is some masking logic that prevents the latent stream from trying to be processed during evaluation (esp. the `to_xarray` call will not work). Otherwise just some stylistic remarks.
I saw in the Issue that the latents should be eventually exposed via an JSON API? Would it make sense to already implement this and not try to piggyback on ZarrIO?

evenmn:
> Overall I like the idea of treating latent output just as another stream. One thing that needs to be addressed is that there is some masking logic that prevents the latent stream from trying to be processed during evaluation (esp. the `to_xarray` call will not work). Otherwise just some stylistic remarks. I saw in the Issue that the latents should be eventually exposed via an JSON API? Would it make sense to already implement this and not try to piggyback on ZarrIO?

Thanks for your feedback. Exposing the latent space via an JSON API is useful when running the model operationally. However, I still think we should export the latent state as a Zarr file, since this is useful for other applications, for instance explanable AI.

clessig:
> > Overall I like the idea of treating latent output just as another stream. One thing that needs to be addressed is that there is some masking logic that prevents the latent stream from trying to be processed during evaluation (esp. the `to_xarray` call will not work). Otherwise just some stylistic remarks. I saw in the Issue that the latents should be eventually exposed via an JSON API? Would it make sense to already implement this and not try to piggyback on ZarrIO?
> 
> Thanks for your feedback. Exposing the latent space via an JSON API is useful when running the model operationally. However, I still think we should export the latent state as a Zarr file, since this is useful for other applications, for instance explanable AI.

Yes, json API is a separate issue and this PR should address writing the latent space to disc as zarr.

clessig (commented):
Overall looks good. But validation_io will be refactored and it should be discussed how to best do the latent output going forward.

packages/common/src/weathergen/common/io.py
```diff
@@ -588,7 +591,7 @@ class OutputBatchData:
     latents: list[list[dict]]
 
     # optional name to use for latent pseudo-stream when yielding latent items
-    latent_stream_name: str | None = None
+    latent_stream_name: str | None = LATENT_STREAM
```
grassesi: Please remove: since you are always using the default here anyway it makes no difference if `self.latent_stream_name` or `LATENT_STREAM` is used. But using using `latent_stream_name` clutters up the namespace/interface of `OutputBatchData`

packages/common/src/weathergen/common/io.py
grassesi: self.latent_stream_name is currently an alias to `LATENT_STREAM` which is never None. So please just use `if self.latents` (This should not disregard my previous comment on these lines.)

src/weathergen/utils/validation_io.py
```diff
@@ -112,6 +112,9 @@ def write_output(
 
     # output stream names to be written, use specified ones or all if nothing specified
     stream_names = [stream.name for stream in cf.streams]
+    # include known pseudo-stream names (e.g. latent) so they are treated as known
+    if io.LATENT_STREAM not in stream_names:
+        stream_names.append(io.LATENT_STREAM)
```
grassesi: Please remove and just use: 
```python
if io.LATENT_STREAM in output_stream_names:
    output_streams[io.LATENT_STREAM] = None
```
in ll. 136

src/weathergen/utils/validation_io.py
```diff
@@ -122,7 +125,7 @@ def write_output(
     latent_stream_name = None
     output_streams = {}
     for name in output_stream_names:
-        if name == "latent":
+        if name == io.LATENT_STREAM:
             latent_stream_name = name
```
grassesi: Please remove this if clause, instead implement the suggestion I commented above

packages/common/src/weathergen/common/io.py
grassesi: Please use io.LATENT_STREAM here.

grassesi (changes requested):

grassesi:
> Overall looks good. But validation_io will be refactored and it should be discussed how to best do the latent output going forward.

The choosen approach should translate relatively well into the refactored version.

evenmn:
Thanks to both of you for the feedback, it truly improved this PR. I believe I have incorporated the suggested changes in my latest commit, but I still need to test the implementation before it gets merged in.

clessig:
> Thanks to both of you for the feedback, it truly improved this PR. I believe I have incorporated the suggested changes in my latest commit, but I still need to test the implementation before it gets merged in.

Can you please test it as far as you can, and then I have a final look.

clessig:
@tjhunter : can we open a separate PR to fix the test.

clessig:
@evenmn : for developing/debugging this, it's probably best to use training with

```
# validation config; full validation config is merge of training and validation config
validation_config: 

  samples_per_mini_epoch: 256
  shuffle: False

  start_date: 2023-10-01T00:00
  end_date: 2023-12-31T00:00
  
  # whether to track the exponential moving average of weights for validation
  validate_with_ema: 
    enabled : True
    ema_ramp_up_ratio: 0.09
    ema_halflife_in_thousands: 1e-3

  # parameters for validation samples that are written to disk
  output : {
    # number of samples that are written
    num_samples: 8,
    # write samples in normalized model space
    normalized_samples: False,
    # output streams to write; default all
    streams: null,
    }

  # run validation before training starts (mainly for model development)
  validate_before_training: True
```

 rather than the integration test (which also do evaluation, inference etc). This way the io is triggered right at the beginning.

clessig:
@evenmn : Trainer should not get ModelParams passed. This creates a dependency that shouldn't be there to ensure proper encapsulation. Computing the healpix coords is cheap so if it is needed we can just redo it in validation_io (but using the same function). The cleanest solution would arguably be to attach just the coords to latent output. If it doesn't incur a performance penalty then that's the way to go.

evenmn:
> @evenmn : Trainer should not get ModelParams passed. This creates a dependency that shouldn't be there to ensure proper encapsulation. Computing the healpix coords is cheap so if it is needed we can just redo it in validation_io (but using the same function). The cleanest solution would arguably be to attach just the coords to latent output. If it doesn't incur a performance penalty then that's the way to go.

@clessig: I found it natural to make the healpix coords part of the ModelParams, but I agree that it's not ideal passing it to the trainer. I will move it to validation_io and see if it degrades the performance (which I doubt it will)

evenmn:
@clessig: I just moved the healpix_coords calculations to validation_io.py

MatKbauer:
Do the written latents have a notion of space and time, i.e., are latent healpix tokens attributed with their lat/lon and time coordinates?
EDIT: Found coords and time [here](https://github.com/ecmwf/WeatherGenerator/pull/1860/changes#diff-580d900cb2e6a0d5c44eada1fba49b4a51df3abdf9c7771ae3ffacc6e909b1e4R817). Looks good to me. @evenmn, can you update to develop once more so that we can merge it?

evenmn:
> Do the written latents have a notion of space and time, i.e., are latent healpix tokens attributed with their lat/lon and time coordinates?

@MatKbauer It comes with lat/lon, you can see the output dimensions in the first comment above. The time dimension is wrong (it has coordinate dimensions), but so is the time dimension in the ERA5 outputs

packages/common/src/weathergen/common/io.py
```diff
@@ -351,9 +355,11 @@ def __init__(
         if self.key.with_source:
             self._append_dataset(self.source, "source")
 
-        if self.key.with_target(forecast_offset):
+        if forecast_offset is not None and self.key.with_target(forecast_offset):
             self._append_dataset(self.target, "target")
             self._append_dataset(self.prediction, "prediction")
+        if latent is not None:
+            self.datasets += latent
```
MatKbauer: Let's use the same notation as for `source`, `target`, and `prediction` above, i.e., using `self._append_dataset(latent, "latent")`

src/weathergen/utils/validation_io.py
```diff
@@ -164,13 +186,279 @@ def write_output(
...
+            group_path = f"{global_sample_idx}/{io.LATENT_STREAM}/{t_idx}"
+            
+            # Create or get group (avoid duplicate entries in ZipStore)
+            group = zio.data_root.get(group_path)
+            if group is None:
+                group = zio.data_root.create_group(group_path)
+            else:
+                _logger.debug(
+                    f"Latent group already exists at {group_path}, skipping creation."
+                )
```
MatKbauer: I ran into an issue when writing latents of a 6x1h model, where `.json` entries got overwrittn. Copilot suggested to resolve as follows, which worked for me (find two more suggestions to this function below)

Remove this block
```suggestion
```

src/weathergen/utils/validation_io.py
```diff
@@ -164,13 +186,279 @@ def write_output(
...
+            npoints = _infer_latent_points_for_metadata(latents_for_sample)
+            (
+                coords_array,
+                geoinfo_array,
+                times_array,
+                coords_len,
+                num_register_tokens,
+                num_class_tokens,
+            ) = _build_latent_metadata(cf, batch, sample_idx_in_batch, npoints)
+
```
MatKbauer, line 253: Add the following block here

```suggestion
            # Collect all attributes upfront so they can be passed to
            # create_group in a single call.  Setting attrs individually
            # after creation causes duplicate zarr.json entries in ZipStore.
            group_attrs: dict[str, int | str] = {}
            if coords_array is not None and times_array is not None:
                group_attrs = {
                    "num_extra_tokens": int(num_register_tokens + num_class_tokens),
                    "num_register_tokens": int(num_register_tokens),
                    "num_class_tokens": int(num_class_tokens),
                    "spatial_points": int(coords_array.shape[0]),
                    "coords_order": "lat_lon",
                }
                if npoints is not None:
                    group_attrs["total_points"] = int(npoints)

            # Create or get group (avoid duplicate entries in ZipStore)
            group = zio.data_root.get(group_path)
            if group is None:
                group = zio.data_root.create_group(group_path, attributes=group_attrs)
            else:
                _logger.debug(
                    f"Latent group already exists at {group_path}, skipping creation."
                )
```

src/weathergen/utils/validation_io.py
```diff
@@ -164,13 +186,279 @@ def write_output(
...
+                _write_array(group, "times", times_array)
+                _logger.debug(f"Wrote times shape {times_array.shape} for sample {global_sample_idx}")
+
+                group.attrs["num_extra_tokens"] = int(num_register_tokens + num_class_tokens)
+                group.attrs["num_register_tokens"] = int(num_register_tokens)
+                group.attrs["num_class_tokens"] = int(num_class_tokens)
+                group.attrs["spatial_points"] = int(coords_array.shape[0])
+                if npoints is not None:
+                    group.attrs["total_points"] = int(npoints)
+                group.attrs["coords_order"] = "lat_lon"
```
MatKbauer: Remove this block (has been added above)

```suggestion
```

MatKbauer (changes requested):
I successfully wrote some latent states to disk and appreciate they are contained seamlessly in the `validation_chkpt00000_rank0000.zip`.

Writing the latents with the 6x1h model resulted in some conflicts and I have tested and proposed a solution provided by Claude. There are some other small comments still, please have a look @evenmn.

With the new `.zip` instead of `.zarr` output, I'd suggest to update the PR description where the written latents can be plotted as follows:
```
store = ZipStore("validation_chkpt00000_rank0000.zip", mode="r")
z = zarr.open_group(store=store, zarr_format=3, mode="r")
```
That is, add the `open_group` explicitly to be compatible with the `.zip` format, which now is the default.

Also, please add a note to the PR description telling how latents can be written, i.e., when passing an according stream info to the inference command, such as
```
uv run inference ... --options test_config.output.streams=[ERA5,latent]
```
Edit: just realised that it is mentioned already in the description. Maybe make it more explicit with the `uv run inference` prefix.

@clessig, can you still have a look and merge if you agree?

packages/common/src/weathergen/common/io.py
```diff
@@ -764,6 +783,53 @@ def _extract_sources(
...
+                data = arr
+                channels = [f"{lname}_{i}" for i in range(data.shape[1])]
+
+            ds = OutputDataset(
+                lname, key, source_interval, data, times, coords, geoinfo, channels, []
+            )
+            datasets.append(ds)
+
+        # TODO: missing forecast offset
+        return OutputItem(key=key, forecast_offset=None, latent=datasets)
```
MatKbauer, line 831: Is there a reason for which we don't have `forecast_offset` included here?

src/weathergen/utils/validation_io.py
```diff
@@ -129,7 +141,14 @@ def write_output(
...
 
-    output_streams = {name: stream_names.index(name) for name in output_stream_names}
+    # Allow a pseudo-stream name 'latent' to enable latent writing while
+    # skipping it from the physical streams mapping.
+    # None means latent
+    output_streams: dict[str, int | None] = {
+        name: stream_names.index(name) for name in output_stream_names if name != io.LATENT_STREAM
+    }
+    if io.LATENT_STREAM in output_stream_names:
+        output_streams[io.LATENT_STREAM] = None
```
MatKbauer: Can we make this more specific? I would not understand the meaning of `None` in other places. Is "latent" not an option? If "latent" would undesiredly be picked up by the physical streams mapper (or writer?), then I'd advocate for adding an exception for "latent."

src/weathergen/utils/validation_io.py
```diff
@@ -154,6 +173,14 @@ def write_output(
     source_windows = (twh.window(idx) for idx in sample_idxs)
     source_intervals = [TimeRange(window.start, window.end) for window in source_windows]
 
+    latents_all = (
+        get_latent_output(batch, model_output) if io.LATENT_STREAM in output_streams else None
+    )
+
+    # Create output_streams dict without latent for passing to OutputBatchData
+    # (latent is handled separately)
+    output_streams_physical = {k: v for k, v in output_streams.items() if v is not None}
```
MatKbauer: Maybe adding the "latent" exception case right here is sufficient already.

src/weathergen/utils/validation_io.py
```diff
@@ -162,13 +189,291 @@ def write_output(
...
+
+    # Delete existing store on first batch to avoid "Group already exists" errors
+    # when re-running inference
+    store_path = config.get_path_results(cf, mini_epoch)
+    if batch_idx == 0 and store_path.exists():
+        _logger.info(f"Removing existing output store from previous run: {store_path}")
+        if store_path.is_dir():
+            shutil.rmtree(store_path)
+        else:
+            store_path.unlink()
```
MatKbauer: I'm a bit undecided of whether we want this to happen silently and favour an explicit flag like `overwrite` or such to the inference script instead. For this PR, I'd rather leave it out.

MatKbauer: Admittedly, I found the `Group already exists` quite annoying several times. But I think it's better to ask the user to explicitly overwrite results if they exist already.

src/weathergen/utils/validation_io.py
```diff
@@ -162,13 +189,291 @@ def write_output(
...
+    for latent-only items.
+
+    Also writes coordinate and time metadata using config healpix coordinates.
+    """
+    # Calculate sample start index for this batch
+    sample_start = batch_idx * batch_size
+
+    # Iterate over latent data
+    for t_idx, latents_for_step in enumerate(data.latents):
+        for sample_idx_in_batch, latents_for_sample in enumerate(latents_for_step):
```
MatKbauer: Can we rename
- `latents_for_step` -> `latents_in_step`
- `latents_for_sample` -> `latents_in_sample`
or am I reading this wrong here?

src/weathergen/utils/validation_io.py
```diff
@@ -162,13 +189,291 @@ def write_output(
...
+        n_samples = len(sample_idxs)
+        for i_sample in range(n_samples):
+            per_sample: dict = {}
+            for lname, lval in latent_pred.items():
+                if isinstance(lval, LatentState):
+                    fields = {
+                        "": lval.z_pre_norm,
+                        "register_tokens": lval.register_tokens,
+                        "class_token": lval.class_token,
+                    }
```
MatKbauer, line 455: Our physical ERA5 naming convention differs from that one chosen here for the latents, i.e.,
```
│       ├── latent
│           ├── coords (12288, 2) float32
│           ├── geoinfo (12288, 0) float32
│           ├── latent_state (12288, 512) float32
│           ├── latent_state_class_token (0, 512) float32
│           ├── latent_state_register_tokens (0, 512) float32
│           └── times (12288,) datetime64[ns]
| 
│   ├── ERA5
│   │   ├── 0
│   │   │   └── source
│   │   │       ├── coords (40320, 2) float32
│   │   │       ├── data (40320, 80) float32
│   │   │       ├── geoinfo (40320, 9) float32
│   │   │       └── times (40320,) datetime64[s]
```
Let's make them consistent. I would suggest to rename
- `latent_state` -> `tokens`
- `latent_state_class_token` -> `class_token`
- `latent_state_register_tokens` -> `register_tokens`

Given the latent state is a group itself already, I think we can ommit "latent_state_*" in the specific names and also align it with the `token` name we use in the code.

MatKbauer (commented):
Thanks for updating this, @evenmn! I have tested it again successfully and found the latents in the `validation.zarr`. Before we can merge, please address the comments I've left.

packages/common/src/weathergen/common/io.py
evenmn: The idea here was that the latent-only items do not need the offset for target/prediction inclusion. If forecast_offset is 1 and forecast_step is 0, the offset_key method will return -1, which can cause strange behavior. I may be wrong

src/weathergen/utils/validation_io.py
evenmn: I removed this silent overwrite behavior

packages/common/src/weathergen/common/io.py
MatKbauer: I understand. The core question is: Do we want to write the initial condition `t_0` as well when writing forecast latents, or do we start writing at $t_0+\Delta t$. The current setup would not write `t_0`, which is in line with running a forecast after all.

evenmn: Yes, I think you're right. For a 40 step forecast the latent states are indexed 0 to 39, which might be confusing since index 0 is actually $t_0+\Delta t$. I think it would be useful to include the initial encoded state ($t_0$) as well, and have indices 0 to 40. What do you think?

MatKbauer: Good idea, to include the IC. This might mess up definitions at other places, though, as in forecast mode with `offset: 1`, we only write predictions and targets starting at $t_0 + \Delta t$. @clessig suggested to only start writing the latent at $t_0 + \Delta t$ for consistency. If we want to write latents for the initial condition, we have to run inference differently, e.g., not in forecast mode.

That is, when running in forecast mode, we want to have everything in step 0 to be empty. No predictions, targets, and latents.

If I'm not mistaken, we could realise this by changing [this line](https://github.com/metno/WeatherGenerator/blob/feature/latent-zarr-writer/src/weathergen/utils/validation_io.py#L219) in the latent writer from

now
```python
for t_idx, latents_for_step in enumerate(data.latents):
```

to
```python
for t_idx, latents_for_step in enumerate(data.latents, start=data.forecast_offset):
```

evenmn: I shifted the latent state index by 1 as suggested above, but I did it by changing the group path on [line 228](https://github.com/metno/WeatherGenerator/blob/0cc4622f7c222e35f7b4a9f12a3560094b1458b2/src/weathergen/utils/validation_io.py#L228) from 

`group_path = f"{global_sample_idx}/{io.LATENT_STREAM}/{t_idx}"`
to
`group_path = f"{global_sample_idx}/{io.LATENT_STREAM}/{t_idx + 1}"`

I ignore the IC for now

evenmn:
@clessig @shmh40 I changed the indexing such that the first latent state after the forecast engine has index 1. Index 0 is reserved for the encoded initial conditions, but I didn't include them for now. I have verified that it gives the expected outputs. Please have a look when you have time :)

clessig (approved):

PR #1908, opened 2026-02-23
Sgrasse/develop/1526 update io

### LINKED ISSUE #1526: validation_io can be simplified

validation_io is currently unnecessarily complex since it converts the new data structures for ModelOutput etc to the old list-of-lists based one to be compatible with weathgen.common.io. The latter should be adapated to directly use the new, dict-based data structures.

CC @grassesi 

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation

grassesi:
## Issue Number

- replaces `utils.output.validation_io` module and `common.io.OutputBatchData` class with module `utils.output`
- properly propagate geoinfos from IOReaderData to TargetAuxOutput (+ minor structural improvement: keep data bundled in IOReaderData object during masking/tokenization)
- add geoinfo channel names to stream_config
- structural improvements to `common.io.OutputItem`
- improvements to path retrieval in config
- Additional cleanup/refactoring that does not change behavior can be found in #2280 

Closes #1526 , 


integration_tests/test_output.py
```diff
@@ -0,0 +1,30 @@
...
+# forecast offset 0/1
+# predictions without targets/sources
+# allow incremental writes
+# allow non continouos fsteps
+# always have source at fstep=0
+
+
+
+
+@pytest.fixture
```
florianscheidl: Complete tests or omit?

packages/common/src/weathergen/common/config.py
```diff
@@ -634,9 +634,14 @@ def load_streams(streams_directory: Path) -> list[Config]:
...
-def get_path_run(config: Config) -> Path:
+def get_path_run(config: Config | None = None, run_id: str | None = None) -> Path:
     """Get the current runs results_path for storing run results and logs."""
-    return _get_shared_wg_path() / "results" / get_run_id_from_config(config)
+    if config or run_id:
+        run_id = run_id if run_id else get_run_id_from_config(config)
+    else:
+        msg = f"Missing run_id and cannot infer it from config: {config}"
+        raise ValueError(msg)
+    return _get_shared_wg_path() / "results" / run_id
```
florianscheidl: For brevity, I'd suggest:

```suggestion
    assert config or run_id, ValueError(f"Missing run_id and cannot infer it from config: {config}")
    run_id = get_run_id_from_config(config) if config
    return Path(os.path.join(_get_shared_wg_path(),"results",run_id)
```

src/weathergen/datasets/batch.py
```diff
@@ -27,12 +30,30 @@ class SampleMetaData:
 
 
 class Sample:
-    # keys: stream name, values: SampleMetaData
-    meta_info: dict[str | SampleMetaData]
+    def __init__(self, streams: dict) -> None:
```
florianscheidl: Be more specific with dict types?

src/weathergen/datasets/batch.py
```diff
@@ -27,12 +30,30 @@ class SampleMetaData:
...
+        self.streams_data: dict[str, StreamData | None] = {}
+
+        for stream_info in streams:
+            self.streams_data[stream_info["name"]] = None
+
+    @property
+    def sample_idx(self) -> int:
+        if not self.is_empty():
+            # choose 'first' streams stream_data
+            stream_data = next(iter(self.streams_data.values()))
```
florianscheidl: Do you intentionally name one streams_data and the other stream_data? Otherwise, they should be consistent throughout.

src/weathergen/datasets/batch.py
```diff
@@ -27,12 +30,30 @@ class SampleMetaData:
...
-    # keys: stream_name, values: StreamData
-    streams_data: dict[str, StreamData | None]
+            assert all(
+                stream_data.sample_idx == idx for stream_data in self.streams_data.values()
+            ), "sample_idx should be identical for all streams within one sample."
+
+            return idx
+        else:
+            msg = "Cannot infer sample_idx without any streams added."
+            raise RuntimeError(msg)
```
florianscheidl: We can do it in one iteration instead of two:

```suggestion
    def sample_idx(self) -> int:
        non_empty = [s for s in self.streams_data.values() if s is not None and not s.empty()]
        if not non_empty:
            raise RuntimeError("Cannot infer sample_idx without any streams added.")
        idx = non_empty[0].sample_idx
        assert all(s.sample_idx == idx for s in non_empty), (
            "sample_idx should be identical for all streams within one sample."
        )
        return idx
```

florianscheidl: Alternatively, change output type of is_empty() to list but this might lead to confusion

src/weathergen/datasets/batch.py
```diff
@@ -149,6 +165,7 @@ def get_subset(self, subset: list | None = None):
             bs.tokens_lens = torch.index_select(bs.tokens_lens, 1, torch_idxs)
             return bs
 
+    # unused
```
florianscheidl: Then let's remove it! :D 🪓
And all the other ones you marked below.

src/weathergen/datasets/batch.py
```diff
@@ -160,24 +177,28 @@ def get_num_steps(self) -> int:
...
         return self.output_steps
 
+    # unused
     def get_device(self) -> str | torch.device:
         """
         Get device of tensors in the batch
         """
         return self.device
 
+    # unused
```
florianscheidl: This seems to be used in https://github.com/grassesi/WeatherGenerator/blob/b7c57c16829e2030e58321f8a041ed0963b3acd7/src/weathergen/datasets/batch.py#L289 and below.

src/weathergen/datasets/batch.py
```diff
@@ -195,40 +216,56 @@ def pin_memory(self):
...
-    # can be more complex for strategies like student-teacher training
-    source2target_matching_idxs: np.typing.NDArray[np.int32]
-    target2source_matching_idxs: np.typing.NDArray[np.int32]
-
-    # indices of valid outputs
-    output_idxs: list[int]
-
-    # device of the tensors in the batch
-    device: str | torch.device
+    The data a instance contains is associated with one particular inital sampling window.
```
florianscheidl:
```suggestion
    The data an instance contains is associated with one particular inital sampling window.
```

src/weathergen/datasets/batch.py
```diff
@@ -195,40 +216,56 @@ def pin_memory(self):
...
+            output_steps: number of forecast steps for this batch.
+        """
+        self.init_time: DTRange = init_time
 
         # define forecast indices
-        self.output_offset = output_offset
-        self.output_steps = output_steps
-        self.output_idxs = list(range(output_offset, output_steps))
+        self.output_offset: BatchSamples = output_offset
+        self.output_steps: BatchSamples = output_steps
```
florianscheidl: Confused about the types, shouldn't it be int?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -521,7 +525,9 @@ def _build_stream_data(
 
         return stream_data
 
-    def _get_data_windows(self, base_idx, num_forecast_steps, num_steps_input_max, stream_ds):
+    def _get_data_windows(
+        self, base_idx, num_forecast_steps, num_steps_input_max, stream_ds
+    ) -> tuple[list[IOReaderData], list[IOReaderData]]:
```
florianscheidl: Nice!

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -619,12 +625,15 @@ def _get_batch(self, idx: int, num_forecast_steps: int):
 
         mode = self.mode_cfg.get("training_mode")
         source_cfgs = self.mode_cfg.get("model_input")
-        target_cfgs = self.mode_cfg.get("target_input", {})
+        target_cfgs: typing.Mapping = self.mode_cfg.get("target_input", {})
```
florianscheidl: Why for targets and not for source?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -649,6 +659,7 @@ def _get_batch(self, idx: int, num_forecast_steps: int):
         for stream_info, (stream_name, stream_ds) in zip(
             self.streams, self.streams_datasets.items(), strict=True
         ):
+            stream_info: Config
```
florianscheidl: Do we want to log/raise if it does not have this type?

src/weathergen/datasets/stream_data.py
```diff
@@ -196,7 +196,7 @@ def add_source(
         idx = torch.isnan(self.source_tokens_cells[step])
         self.source_tokens_cells[step][idx] = self.mask_value
 
-    def add_target(
+    def add_target(  # TODO what is the diffference to add_target_values???
```
florianscheidl: It looks like it is not used, so let's remove it?

src/weathergen/datasets/tokenizer_masking.py
```diff
@@ -53,26 +53,34 @@ def reset_rng(self, rng) -> None:
...
+        tokens: list[tuple[list[list[torch.Tensor | None]], list[list[int]]]] = []
         for rdata in data:
             # skip empty data
             if rdata.is_empty():
                 continue
             # tokenize data
-            idxs_cells, idxs_cells_lens = tok(
-                readerdata_to_torch(rdata), token_size, hl, pad_tokens
-            )
+            if stream_info.get("tokenize_spacetime", False):
```
florianscheidl: I think your version is more readible but logically, it makes more sense to me to check the criteria once rather than in each iteration of data. You might find a middle ground by changing the naming?

src/weathergen/model/model.py
```diff
@@ -76,6 +78,28 @@ def get_physical_prediction(
                 pred = pred[sample_idx]
         return pred
 
+    def get_physical_prediction_normalized(
+        self, key: ItemKey, normalizer: typing.Callable
+    ) -> np.typing.NDArray:
+        try:
+            # TODO why is there a tuple => what index should be used
```
florianscheidl, line 92: Lacking context for now 🙈

src/weathergen/train/target_and_aux_module_base.py
```diff
@@ -10,40 +10,51 @@
...
     def add_physical_target(
         self, timestep_idx: int, stream_name: StreamName, pred: torch.Tensor
     ) -> None:
         self.physical[timestep_idx][stream_name] = pred
 
+    # currently unused?
     def add_latent_target(self, timestep_idx: int, latent_name: str, pred: torch.Tensor) -> None:
         self.latent[timestep_idx][latent_name] = pred
 
+    # is broken => pred[sample_idx] must be pred["target"][sample_idx]
```
florianscheidl: I don't see where it's used?

src/weathergen/train/target_and_aux_module_base.py
```diff
@@ -58,6 +69,38 @@ def get_physical_target(
...
+    def get_physical_target_normalized(
+        self, key: ItemKey, normalizer: typing.Callable
+    ) -> PhysicalTarget:
+        try:
+            stream_targets = self.physical[key.forecast_step][key.stream]
+        except (KeyError, IndexError) as e:
+            msg = f"Cannot find physical target data for key: {key}"
+            raise ValueError(msg) from e
+
+        # is it a performance issue if I dont convert/move the entire tensor at once?
```
florianscheidl: Maybe too big for this PR (and I lack context for now) but I think converting all at once would be a performance win..

src/weathergen/train/trainer.py
```diff
@@ -604,17 +613,22 @@ def validate(self, mini_epoch, mode_cfg, batch_size):
...
-                            targets_and_auxs,
+                        fstep_start = 0
+                        targets = [
+                            targets
+                            for loss_name, targets in targets_and_auxs.items()
+                            if loss_name == PHYSICAL_LOSS_KEY
+                        ]
+                        try:
+                            # targets for all physical outputs should be the same
+                            targets = targets[0]
```
florianscheidl: Missing an all? - one's a list, other's a TargetAuxOutput.

src/weathergen/utils/validation_io.py
```diff
@@ -8,20 +8,34 @@
...
 from weathergen.common.io import TimeRange, zarrio_writer
+from weathergen.datasets.batch import ModelBatch
 from weathergen.datasets.data_reader_base import TimeWindowHandler
+from weathergen.model.model import ModelOutput
+from weathergen.train.target_and_aux_module_base import TargetAuxOutput
 
 _logger = logging.getLogger(__name__)
 
 
 def write_output(
```
florianscheidl: If we refactor to use output.py, we're no longer using write_output. Should we remove it?

src/weathergen/train/trainer.py
```diff
@@ -45,13 +48,14 @@
     get_target_idxs_from_cfg,
 )
 from weathergen.utils.distributed import is_root
+from weathergen.utils.output import Writer
 from weathergen.utils.train_logger import TrainLogger, prepare_losses_for_logging
 from weathergen.utils.utils import get_dtype
-from weathergen.utils.validation_io import write_output
```
florianscheidl, line 52: Let's remove the source before we merge.

src/weathergen/utils/output.py
```diff
@@ -0,0 +1,229 @@
...
+
+class Writer:
+    def __init__(
+        self,
+        config: Config,
+        val_cfg: Config,
+        streams: list[Config],
+    ):
+        streams = {stream.name: stream for stream in streams}
+        # TODO: nice: dont store all config
```
florianscheidl: Don't completely understand the reason for this - why not do the following below:

`
self._streams = {
    stream.name: stream for stream in streams if stream.name in _output_streams
}
`

src/weathergen/utils/output.py
```diff
@@ -0,0 +1,229 @@
...
+        return _ExtractedData("target", target.data, coords.times, coords.coords, coords.geoinfos)
+
+    def extract_prediction(self, key: ItemKey) -> _ExtractedData:
+        try:
+            data = self._predictions.get_physical_prediction_normalized(key, self._normalizer)
+        except Exception as e:
+            # TODO: if preds are empty so create copy of target and add ensemble dimension
+            # preds = [targets[0].clone().unsqueeze(0)]
+            raise ValueError("not handled yet") from e
+            data = self._target(key)
```
florianscheidl: Clean up?

src/weathergen/utils/output.py
```diff
@@ -0,0 +1,229 @@
...
+
+    @property
+    def source_interval(self) -> TimeRange:
+        window = self._batch.init_time
+        return TimeRange(window.start, window.end)
+
+    @property
+    def samples(self) -> list[int]:
+        # TODO check: data._batch.source_samples.samples
+        sampels = list(self._batch.target_samples.sample_idxs)
```
florianscheidl: Typo

src/weathergen/utils/output.py
```diff
@@ -0,0 +1,229 @@
...
+        # The order of iteration is important here:
+        # streams is the outermost and samples the innermost loop variable
+        # This is important since normalization is best done at a per stream/fstep basis
+        for stream, forecast_step, sample_idx in itertools.product(
+            streams, forecast_steps, samples
+        ):
+            yield ItemKey(sample_idx, forecast_step, stream)
+            n_samples += 1
+
+        self._sample_start += n_samples
```
florianscheidl: Not sure if this works correctly being a generator and interleaved with extract. See comment in itemize().

src/weathergen/utils/output.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def itemize(
+        self, data: _BatchOutputData, fstep_range: range
+    ) -> typing.Generator[OutputItem, None, None]:
+        """Iterate over possible output items"""
+        fstep_start = min(fstep_range)
+
+        # TODO: check: filter for empty items?
+        for key in self.keys(fstep_range, data.samples):
+            yield self.extract(data, key, fstep_start)
```
florianscheidl:
```suggestion
    def itemize(
        self, data: _BatchOutputData, fstep_range: range
) -> typing.Generator[OutputItem, None, None]:
    fstep_start = min(fstep_range)
    sample_start = self._sample_start  # capture before keys() runs

    for key in self.keys(fstep_range, data.samples):
        yield self.extract(data, key, fstep_start, sample_start)

```

florianscheidl: See comment in .keys() and .extract()

src/weathergen/utils/output.py
```diff
@@ -0,0 +1,229 @@
...
+        ):
+            yield ItemKey(sample_idx, forecast_step, stream)
+            n_samples += 1
+
+        self._sample_start += n_samples
+
+    def extract(self, data: _BatchOutputData, key: ItemKey, fstep_start: int) -> OutputItem:
+        raw_key = ItemKey(
+            key.sample - self._sample_start, key.forecast_step - fstep_start, key.stream
+        )
```
florianscheidl:
```suggestion
    def extract(self, data: _BatchOutputData, key: ItemKey, fstep_start: int, sample_start: int) -> OutputItem:
        raw_key = ItemKey(
            key.sample - sample_start, key.forecast_step - fstep_start, key.stream
        )
```

src/weathergen/utils/output.py
```diff
@@ -0,0 +1,229 @@
...
+            return None
+        else:
+            return OutputDataset(
+                name=self.name,
+                item_key=key,
+                data=self.data,
+                times=self.times,
+                coords=self.coords,
+                geoinfo=self.geoinfos,
+                source_interval=invariants.source_interval,
```
florianscheidl, line 297: Probably missing something conceptually but we're not using the source_channel here.

florianscheidl (changes requested):

florianscheidl (changes requested):
Many nice changes, to assess the functional improvements, I'd need to more time to understand the changes in depth.

Some higher-level comments:
1. For typing, it might be nice to add some static typechecking like pyright or ty (Astral)
2. Would be helpful to have more docstrings/explainers for the new classes and functions introduced in Writer
3. What do we use for logging? I've seen a couple of raise and specific errors but no logging.

I'll test it on our HPC once we've implemented the fixes.

integration_tests/test_output.py
grassesi: I will omit it :+1:

src/weathergen/datasets/batch.py
grassesi: yes definitively 😅

src/weathergen/datasets/batch.py
grassesi: yes thanks

src/weathergen/datasets/batch.py
grassesi: actually this is not even a dict :grimacing: but a list. I corrected it

src/weathergen/datasets/batch.py
grassesi: yes since I only choose data from one stream.

src/weathergen/utils/output.py
grassesi: No this is indeed a bug: depending on what dataset this is we should use source or target channels:
- "source": source channels
- "target", "prediction": target channels

src/weathergen/utils/validation_io.py
grassesi: Yes this to easier test, but it should be removed before merging

src/weathergen/datasets/multi_stream_data_sampler.py
grassesi: I added it :)

src/weathergen/datasets/multi_stream_data_sampler.py
grassesi: We can assume it is always correct, the static type analysis in my IDE just could not find it. I can remove it if you want to.

src/weathergen/train/trainer.py
grassesi: Per LossCalculator a TargetAuxOutput instance exists. Usually there is only one physical LossCalculator but nothing prevents the user from configuring multiple. That is why we need to make a choice here. I renamed the variables to make it more clear: `targets = physical_targets_all[0]`

src/weathergen/utils/output.py
grassesi: There is no interleaving here: execution in `Writer.keys` is halted at the yield until the next key is requested in `Writer.items` then execution resumes where it was halted. You can verify for yourself using this snippet:
```python
class A:
    def __init__(self):
        self.start = 2
    def items(self):
        print("start execution of items loop")
        for key in self.keys():
          print("yielding item")
          yield self.start + key
    def keys(self):
        n_keys=0
        for key in range(5):
            print("yielding key")
            yield key
            print("incrementing key counter")
            n_keys+=1
        print("incrementing start")
        self.start+=n_keys
```
```python
a = A()
items = a.items()
# => no exection yet since generator is evaluated lazily
list(items)
```
```sh
start execution of items loop
yielding key
yielding item
incrementing key counter
yielding key
yielding item
incrementing key counter
yielding key
yielding item
incrementing key counter
yielding key
yielding item
incrementing key counter
yielding key
yielding item
incrementing key counter
incrementing start
```
result: `print(items)` => `[2, 3, 4, 5, 6]`
result the second time: `print(list(a.item()))` => `[7, 8, 9, 10, 11]`

grassesi: Let me know if it is fine for you to leave it as is, otherwise I am also fine to implement your suggestion.

src/weathergen/utils/output.py
grassesi: see above

src/weathergen/utils/output.py
grassesi: see above

src/weathergen/train/target_and_aux_module_base.py
grassesi: I agree, fortunately the number of blocks retrieved from this tensor is usually quite low (`training_config.target_input.<target_name>.num_samples`). So it is rather a matter of: will this hurt badly? If yes some caching logic needs to be put in place, since every call to `get_physical_prediction_normalized` and `get_physical_target_normalized` is independently issued (We extract one output item at a time)

src/weathergen/train/target_and_aux_module_base.py
grassesi: My guess is that it was designed to be used by the loss calculation and the output, but in the loss calculation the attributes are used directly, and for the output I implemented the `get_physical_target_normalized` method which suited my needs better.

src/weathergen/utils/output.py
grassesi: Yes I will just remove the try except. In `get_physical_prediction_normalized` is an assert, checking preds is non-empty.

packages/common/src/weathergen/common/config.py
grassesi: I changed it also in `get_path_model` to keep the logic consistent :+1:

src/weathergen/utils/output.py
grassesi: `streams` should be at some point converted into a dict instead of a list. And the keys of streams are used as a default value for `_output_streams`.

src/weathergen/utils/output.py
florianscheidl: Thanks for clarifying! We can leave as is then :)

packages/common/src/weathergen/common/io.py
```diff
@@ -439,7 +440,9 @@ def _get_datasets(self, key: ItemKey):
     def _get_group(self, item: ItemKey, create: bool) -> zarr.Array | zarr.Group:
         assert self.data_root is not None, "ZarrIO must be opened before accessing data."
         if create:
-            assert self.data_root.get(item.path) is None, "Group already exists, stop overwriting"
+            assert self.data_root.get(item.path) is None, (
+                f"Group at {ItemKey} already exists, stop overwriting"
```
clessig: Group at {ItemKey} already exists. Aborting.

packages/common/src/weathergen/common/io.py
```diff
@@ -353,6 +353,7 @@ def __init__(
             self._append_dataset(self.source, "source")
 
         if self.key.with_target(forecast_offset):
+            # TODO requiring target data currently prevents predictions with unknowable target
```
clessig: Does the code work when target is None? Or should we have a tensor of size `(0,num_channels)`?

src/weathergen/datasets/batch.py
clessig: No, let's please keep it. It defines a more or less complete interface for ModelBatch. We don't want to always need to change this class when we use it elsewheree.

src/weathergen/datasets/batch.py
```diff
@@ -195,40 +216,56 @@ def pin_memory(self):
...
+        source2target_matching_idxs: index of corresponding target indices.
+        target2source_matching_idxs: index of corresponding source indices.
+        output_idxs: Forecast step indices (including offset) for this batch. The number of
+          forecast steps is constant per batch, this amortized through data parallel training.
+        device: device of the tensors in this batch.
+    """
 
     def __init__(
         self,
+        init_time: DTRange,
```
clessig: `source_time_range`

src/weathergen/datasets/batch.py
clessig: This should replace the whole paragraph:

`Class representing one batch processed by the model.`

This is not the place to explain how some of the general processing works.

src/weathergen/datasets/batch.py
```diff
@@ -195,40 +216,56 @@ def pin_memory(self):
...
+    Note that `output_offset` is the difference in forecast steps
+    between corresponding source and target samples.
+    Source and target samples are in 1-to-1 correspondance for classical training modes
+    (e.g. MTM, forecasting), but can be more complex for strategies like student-teacher training.
+    This relationship is expressed in `source2target_matching_idxs`
+    and `target2source_matching_idxs`.
+
+
+    Attributes:
+        init_time: Initial sampling window used to costruct samples.
```
clessig: source_time_range: temporal window of source data

src/weathergen/datasets/batch.py
```diff
@@ -195,40 +216,56 @@ def pin_memory(self):
...
+        source_samples: sources for model
```
clessig: inputs for models

src/weathergen/datasets/batch.py
```diff
@@ -195,40 +216,56 @@ def pin_memory(self):
...
+        target_samples: targets for TargetAuxCalculator
```
clessig: inputs for TargetAuxCalculator that determine targets

src/weathergen/datasets/batch.py
```diff
@@ -195,40 +216,56 @@ def pin_memory(self):
...
+        source2target_matching_idxs: index of corresponding target indices.
+        target2source_matching_idxs: index of corresponding source indices.
+        output_idxs: Forecast step indices (including offset) for this batch. The number of
+          forecast steps is constant per batch, this amortized through data parallel training.
```
clessig: amortized is the wrong word. But everything after `for this batch` should be removed

`Forecast step indices`

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -197,6 +198,7 @@ def __init__(
                     if ds.target_channel_weights is not None
                     else [1.0 for _ in ds.target_channels]
                 )
+                stream_info["geoinfo_channels"] = ds.geoinfo_channels
```
clessig: Thanks!

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -477,6 +479,8 @@ def _build_stream_data(
             modes :
             stream_data :
             base_idx: Time index for this sample
+                => no its tidx (index of n-th target mask, no info of the time index for this sample
```
clessig: Remove. Has been fixed on develop. (Also good to ping people if you find a potential bug)

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -733,6 +752,7 @@ def __iter__(self) -> ModelBatch:
         logger.info(f"iter_start={iter_start}, iter_end={iter_end}, len={self.len}")
 
         # create new shuffeling
+        # should return permutations instead of relying on self.perms
```
clessig: Remove or fix.

src/weathergen/datasets/stream_data.py
clessig: This function should be removed.

src/weathergen/datasets/tokenizer_utils.py
```diff
@@ -192,21 +195,25 @@ def tokenize_spacetime(
...
         mask = t == rdata.datetimes
         rdata_cur = IOReaderData(
             rdata.coords[mask], rdata.geoinfos[mask], rdata.data[mask], rdata.datetimes[mask]
         )
         idxs_cur, idxs_cur_lens = tokenize_space(rdata_cur, token_size, hl, pad_tokens)
 
-        # collect data for all time steps
-        idxs_cells = [t + tc for t, tc in zip(idxs_cells, idxs_cur, strict=True)]
-        idxs_cells_lens = [t + tc_l for t, tc_l in zip(idxs_cells_lens, idxs_cur_lens, strict=True)]
+        # append tokens/n_tokens for current time step for each healpix cell to existing tokens
```
clessig: I don't see how this is equivalent to the above: `idxs_cells` is not updated in your code

src/weathergen/datasets/tokenizer_utils.py
```diff
@@ -382,6 +389,7 @@ def return_empty(rdata, idxs_cells_lens):
     else:
         coords_local = torch.tensor([])
 
+    # geoinfos information is empedded into coords_local here
```
clessig: empedded -> embedded ?

src/weathergen/model/model.py
```diff
@@ -76,6 +78,28 @@ def get_physical_prediction(
...
+            pred = self.physical[key.forecast_step][key.stream][0][key.sample]
+        except (KeyError, IndexError) as e:
```
clessig, line 95: No exceptions in the model!

src/weathergen/model/model.py
clessig: The index should be forecast step.

src/weathergen/model/model.py
```diff
@@ -76,6 +78,28 @@ def get_physical_prediction(
...
+            msg = f"Cannot find prediction data for key: {key}"
+            raise ValueError(msg) from e
+
+        # is it a performance issue if I dont convert/move the entire tensor at once?
+        pred = pred.to(torch.float32).detach().cpu().numpy()
+        pred = normalizer(key.stream, pred)
+
+        assert isinstance(pred, np.ndarray), "Invalid buffer type."  # noqa: TID251
```
clessig, line 103: How could this be triggered? In l92 you explicitly cast to np.array

src/weathergen/model/model.py
```diff
@@ -76,6 +78,28 @@ def get_physical_prediction(
...
+        # TODO What to do when preds are empty,when does it occur,
+        # how does it show (empty tensor or missing key?)
+        assert len(pred) > 0
```
clessig, line 106: This could happen if a stream is masked completely, e.g., or if a dataset does not exist for a specific time step. So we need to handle this.

src/weathergen/train/target_and_aux_module_base.py
```diff
@@ -10,40 +10,51 @@
...
+from weathergen.datasets.batch import SampleMetaData
 from weathergen.model.engines import LatentState
 
 type StreamName = str
 
 
 @dataclasses.dataclass
+class PhysicalTarget:
+    data: np.typing.NDArray[np.float32]
+    coords: np.typing.NDArray[np.float32]
```
clessig, line 28: Why no geoinfos?

src/weathergen/train/target_and_aux_module_base.py
```diff
@@ -10,40 +10,51 @@
...
+        ] = [{} for _ in range(len_target)]
+        self.latent: list[dict[str, torch.Tensor | LatentState]] = [{} for _ in range(len_target)]
+        self.aux_outputs: dict[str, torch.Tensor] = {}
 
     def add_physical_target(
         self, timestep_idx: int, stream_name: StreamName, pred: torch.Tensor
     ) -> None:
         self.physical[timestep_idx][stream_name] = pred
 
+    # currently unused?
```
clessig: Remove. Should be used e.g. for latent diffusion model.

src/weathergen/train/target_and_aux_module_base.py
clessig: Yes, but please fix.

src/weathergen/train/target_and_aux_module_base.py
```diff
@@ -58,6 +69,38 @@ def get_physical_target(
                 pred = pred[sample_idx]
         return pred
 
+    # TODO guarantee every buffer retrieved only once
+    def get_physical_target_normalized(
+        self, key: ItemKey, normalizer: typing.Callable
+    ) -> PhysicalTarget:
+        try:
```
clessig, line 75: The model needs to guarantee that all the entries exist.

src/weathergen/train/target_and_aux_module_base.py
```diff
@@ -58,6 +69,38 @@ def get_physical_target(
...
+        # is it a performance issue if I dont convert/move the entire tensor at once?
+        coords = (
+            stream_targets["target_coords"][key.sample].to(torch.float32).detach().cpu().numpy()
+        )
+        times = stream_targets["target_times"][key.sample]
+
+        data = stream_targets["target"][key.sample].to(torch.float32).detach().cpu().numpy()
+        data = normalizer(key.stream, data)
+
+        assert isinstance(data, np.ndarray), "Invalid data buffer type."  # noqa: TID251
```
clessig: How can this not be a numpy array when you call `.numpy()` right before?

src/weathergen/train/target_and_aux_module_base.py
```diff
@@ -58,6 +69,38 @@ def get_physical_target(
...
+        assert isinstance(coords, np.ndarray), "Invalid coords buffer type."  # noqa: TID251
+        assert isinstance(times, np.ndarray), "Invalid datetimes buffer type."  # noqa: TID251
+
+        if len(coords) == 0:  # TODO can this be removed?
```
clessig, line 104: It probably patches a problem elsewhere but let's leave it if it was there before.

src/weathergen/train/trainer.py
```diff
@@ -45,13 +48,14 @@
...
 
 logger = logging.getLogger(__name__)
 
 # cfg_keys_to_filter = ["losses", "model_input", "target_input"]
+PHYSICAL_LOSS_KEY = "physical"
```
clessig, line 57: I don't see the benefit of this indirection

src/weathergen/train/trainer.py
```diff
@@ -153,6 +158,10 @@ def init(self, cf: Config, devices):
             config.get_path_model(cf).mkdir(exist_ok=True, parents=True)
 
         self.train_logger = TrainLogger(cf, config.get_path_run(self.cf))
+        if use_test_config:
```
clessig: Remove before merging.

src/weathergen/train/trainer.py
```diff
@@ -178,7 +187,7 @@ def get_target_aux_calculators(self, mode_cfg):
 
     def inference(self, cf, devices, run_id_contd, mini_epoch_contd):
         # general initalization
-        self.init(cf, devices)
+        self.init(cf, devices, True)
```
clessig: Remove

src/weathergen/train/trainer.py
```diff
@@ -604,17 +613,22 @@ def validate(self, mini_epoch, mode_cfg, batch_size):
...
-                            preds,
-                            targets_and_auxs,
+                        fstep_start = 0
+                        targets = [
+                            targets
+                            for loss_name, targets in targets_and_auxs.items()
+                            if loss_name == PHYSICAL_LOSS_KEY
+                        ]
+                        try:
+                            # targets for all physical outputs should be the same
```
clessig, line 632: I don't understand this statement

src/weathergen/train/trainer.py
```diff
@@ -604,17 +613,22 @@ def validate(self, mini_epoch, mode_cfg, batch_size):
...
-                            mode_cfg,
-                            batch_size,
-                            mini_epoch,
-                            bidx,
-                            denormalize_data_fct,
-                            batch,
-                            preds,
-                            targets_and_auxs,
+                        fstep_start = 0
+                        targets = [
```
clessig: I don't think this unpacking belongs here. I rather have it in the writing, or you hide it in another function. The code in `validate()` is complex enough as is.

src/weathergen/train/trainer.py
```diff
@@ -604,17 +613,22 @@ def validate(self, mini_epoch, mode_cfg, batch_size):
...
+                            targets
+                            for loss_name, targets in targets_and_auxs.items()
+                            if loss_name == PHYSICAL_LOSS_KEY
```
clessig, line 629: What about latent outputs?

src/weathergen/utils/output.py
```diff
@@ -0,0 +1,229 @@
...
+        val_cfg: Config,
+        streams: list[Config],
+    ):
+        streams = {stream.name: stream for stream in streams}
+        # TODO: nice: dont store all config
+
+        try:  # TODO: do this in config
+            _output_streams = val_cfg.output.streams
+            if _output_streams is None:
+                raise ConfigAttributeError("")
```
clessig: This should generate a warning and then terminate regularly

src/weathergen/utils/output.py
```diff
@@ -0,0 +1,229 @@
...
+        except ConfigAttributeError:
+            _output_streams = list(streams.keys())
+
+        self._streams = {
+            name: config for name, config in streams.items() if name in _output_streams
+        }
+        try:  # TODO: do this in config
+            self._forecast_offset = val_cfg.forecast.offset
```
clessig: Why not use `self._forecast_offset = val_cfg.forecast.get( "offset", 0)`--that's much more readable than the try/except

src/weathergen/utils/output.py
```diff
@@ -0,0 +1,229 @@
...
+        coords = self._target_coordinates(target)
+
+        return _ExtractedData("target", target.data, coords.times, coords.coords, coords.geoinfos)
+
+    def extract_prediction(self, key: ItemKey) -> _ExtractedData:
+        try:
+            data = self._predictions.get_physical_prediction_normalized(key, self._normalizer)
+        except Exception as e:
+            # TODO: if preds are empty so create copy of target and add ensemble dimension
+            # preds = [targets[0].clone().unsqueeze(0)]
```
clessig: We need this. What's the problem?

src/weathergen/datasets/batch.py
```diff
@@ -134,6 +146,10 @@ def to_device(self, device):
 
         return self
 
+    @property
+    def sample_idxs(self) -> list[int]:
```
clessig: Why is this needed?

clessig (commented):
See comments. I will also rebase.

src/weathergen/datasets/batch.py
grassesi: At least I was missing this context when processing the data, where should I put the information instead?

src/weathergen/model/model.py
grassesi: Since this is a important interface function, I want to have postcondition checks be extra sure I return the correct data.

src/weathergen/train/target_and_aux_module_base.py
grassesi: geoinfos is part of the coords buffer: First two columns are lat,lon everything else is geoinfos. I currently handle this in the output module, but I can also implement it here.

src/weathergen/train/target_and_aux_module_base.py
grassesi: Since this is an interface function I want to provide useful errors.

src/weathergen/utils/output.py
grassesi: The try except also handles the case where a training/val/test config does not contain a "forecast" section. Can I always assume this is always there I am happy to change.

src/weathergen/utils/output.py
grassesi: Currently not every training/validation/test config implements a "output" section, this guarantees a fallback/default. If the output section is missing should the default instead be no output/ no-op writer? Failing instead would mean that configs where no output is needed would require to specify a "output" section.

clessig: If there's no output section or no streams are specified or the specified streams are not present should all generate a warning and terminate regularly.

src/weathergen/utils/output.py
clessig: `val_cfg.get( "forecast", {}).get( "offset, 0)` is better

src/weathergen/datasets/batch.py
clessig: MultiStreamDataSampler is responsible for assembling batches

src/weathergen/train/target_and_aux_module_base.py
clessig: It would perhaps be clearer to not overload coords that elsewher are always lat/lon

src/weathergen/train/target_and_aux_module_base.py
clessig: The dict access will fail and if anything fails in the targt_aux_calculator an expert needs to have a look

src/weathergen/train/trainer.py
grassesi: If I understood correctly each subsection in `training_config.losses` defines one TargetAuxCalculator instance, for each of these instances a TargeAuxOutput will be generated, containing the targets for multiple loss functions.
For now I have only ever seen one physical TargetAuxCalculator and/or one ssl TargetAuxCalculator. Are there usecases that I am not aware of where it would make sense to configure something along the lines of:
```yaml
training_config:
  "physical": {
        type: LossPhysical,
        weight: 0.5,
        loss_fcts: {
          "mse": {
            weight: 0.8,
            target_source_correspondence: { 0 : { 0 : "complement"} },
          },
        },
        target_and_aux_calc: "Physical",
    },
  "physical2": {
        type: LossPhysical,
        weight: 0.5,
        loss_fcts: {
            ...
        },
        target_and_aux_calc: "Physical",
    },
```
If not the comment above is superfluous, but otherwise there are competing definitions of what gridpoints to output that needs to somehow be resolved.

src/weathergen/train/trainer.py
grassesi: If possible I would like to keep this out of scope for now and wait for #1860 to be finalized.

src/weathergen/model/model.py
grassesi: Thanks for explaining :), I assume in this case pred is not none but just an empty tensor?

src/weathergen/model/model.py
grassesi: What is the rationale for this? This is a important data retrieval interface that is exclusively used by downstream code not the model itself, providing useful exceptions here highlights this fact. This also increases the consistency in behavior between `ModelOutput.get_physical_prediction_normalized` and TargetAuxOutput.get_physical_target_normalized.

src/weathergen/datasets/tokenizer_utils.py
grassesi: `idxs_cells` and `idxs_cells_lens` are of type `list[list[Any]]` and `list[list[Any]]` respectively. Iterating over over them yields references to the list objects. Extending the list object in the loop will automatically update the list object in the original list since they share the same reference. You can check this behavior with this snippet:
```python
ll1: list[list[str]] = [["a"], ["b"], ["c"]]
ll2: list[list[int]] = [[0], [1], [2]]
for l1, l2 in zip(ll1, ll2, strict=True):
  l1.extend(l2)

print(ll1)
```
=> `[['a', 0], ['b', 1], ['c', 2]]`

packages/common/src/weathergen/common/io.py
grassesi: No currently neither option works. But it should work with target=None Writing an entire extra dataset to zarr that is totally empty is a bad band aid solution. I think we should rename `ItemKey.with_target` to `ItemKey.with_prediction` (since it is the prediction we are always interested in if the target is present it just means some ground truth is available and if not that is also fine) and require just that prediction is not None. Will implement.

src/weathergen/datasets/batch.py
grassesi: Just quality of life: Use `batch.target_samples.sample_idxs` instead of `sample_idxs = [sample .sample_idx for sample in batch.target_samples.samples]`

src/weathergen/datasets/multi_stream_data_sampler.py
grassesi: I think this is a nice would be a nice refactoring for the MultiStreamDataSampler, but its not needed right now, I will just file an issue: #2059

src/weathergen/train/trainer.py
grassesi: This just highlights that this is a constant with special meaning, the config module would be an alternative place to define this constant, since the constant is related to a value defined in the config.

src/weathergen/datasets/batch.py
clessig: Why do we need `sample .sample_idx`--this is not on current develop

PR #1916, opened 2026-02-24
[1914][1596] Implemented step_callback for incremental output of forecast outputs

### LINKED ISSUE #1596: Forecast output needs to be written incrementally

### LINKED ISSUE #1914: Incremental forward function for long-term roll-out

To perform arbitarily long roll-out, the predictions need to be written incrementally. This requires:
1. a forward function that can be continued, i.e. takes latents as input (this might also be useful for latent-based decoder runs)
2. running with empty targets
3 incremental writing of lead time dimension 

It needs to be decided if this can easily be broken up into 2 or 3 PRs, or if it's best done in one

CC @grassesi : the third point is also relevant for you

_No response_

### Area

- [x] model
- [x] science
- [ ] documentation

evenmn:
## Description

Enables incremental output writing during inference by writing forecast predictions after each step (or every N steps) instead of accumulating all steps in memory. Maintains constant memory usage regardless of forecast length, addressing limitations for high-resolution, multi-step forecasts.

### Changes

- **model.py**: Added optional `step_callback` parameter to `Model.forward()`
- **validation_io.py**: Refactored with modular architecture (`_prepare_batch_data()`, `_process_timestep()`, `StreamingOutputWriter`)
- **trainer.py**: Updated `validate()` for streaming callback support

## Issue Number

Closes #1914 
Closes #1596

### DIFF
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -220,14 +220,8 @@ validation_config:
     normalized_samples: False,
     # output streams to write; default all
     streams: null,
-    # streaming output configuration for inference
-    streaming: {
-      # enable streaming mode: write output after each N forecast steps instead of all at once
-      enabled: False,
-      # number of forecast steps to process before writing to disk
-      # if null, writes after each step (num_steps=1)
-      num_steps: null,
-    }
+    # number of forecast steps to accumulate before writing to disk, default no streaming
+    fstep_chunk_size: null,
     }
 
   # run validation before training starts (mainly for model development)
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -220,6 +220,14 @@ validation_config:
     normalized_samples: False,
     # output streams to write; default all
     streams: null,
+    # streaming output configuration for inference
+    streaming: {
+      # enable streaming mode: write output after each N forecast steps instead of all at once
+      enabled: False,
+      # number of forecast steps to process before writing to disk
+      # if null, writes after each step (num_steps=1)
+      num_steps: null,
+    }
     }
 
   # run validation before training starts (mainly for model development)
--- a/src/weathergen/model/model.py
+++ b/src/weathergen/model/model.py
@@ -614,20 +614,13 @@ class Model(torch.nn.Module):
             z_pre_norm=tokens,
         )
 
-    def forward(
-        self, model_params: ModelParams, batch: ModelBatch, step_callback=None
-    ) -> ModelOutput:
+    def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
         """Forward pass of the model
 
         Tokens are processed through the model components, which were defined in the create method.
         Args:
             model_params : Query and embedding parameters
             batch : Batch of data
-            step_callback : Optional callback function to be called after each forecast step.
-                           Called as step_callback(step, output) where step is the forecast step
-                           index and output is the partial ModelOutput for that step.
-                           Used for streaming output writing - allows writing after each step
-                           instead of accumulating all steps in memory.
         Returns:
             A list containing all prediction results
         """
@@ -652,10 +645,6 @@ class Model(torch.nn.Module):
             output = self.predict_decoders(model_params, step, tokens, batch, output)
             # latent predictions (raw and with SSL heads)
             output = self.predict_latent(model_params, step, tokens, batch, output)
-
-            # invoke callback for streaming output if provided
-            if step_callback is not None:
-                step_callback(step, output)
 
         return output
 
--- a/src/weathergen/model/model.py
+++ b/src/weathergen/model/model.py
@@ -614,7 +614,9 @@ class Model(torch.nn.Module):
             z_pre_norm=tokens,
         )
 
-    def forward(self, model_params: ModelParams, batch: ModelBatch, step_callback=None) -> ModelOutput:
+    def forward(
+        self, model_params: ModelParams, batch: ModelBatch, step_callback=None
+    ) -> ModelOutput:
         """Forward pass of the model
 
         Tokens are processed through the model components, which were defined in the create method.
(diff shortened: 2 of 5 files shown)


evenmn:
Fixes #1914

config/default_config.yml
```diff
@@ -220,6 +220,14 @@ validation_config:
...
     # output streams to write; default all
     streams: null,
+    # streaming output configuration for inference
+    streaming: {
+      # enable streaming mode: write output after each N forecast steps instead of all at once
+      enabled: False,
+      # number of forecast steps to process before writing to disk
+      # if null, writes after each step (num_steps=1)
+      num_steps: null,
+    }
```
grassesi: I dont think we need an entirely new segment in the config for streaming: We can simply add a new option: `training_config/validation_config.output.chunk_size`. 
If `chunk_size` is `None` => no streaming / everything processed at once

src/weathergen/model/model.py
```diff
@@ -646,6 +653,10 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
             # latent predictions (raw and with SSL heads)
             output = self.predict_latent(model_params, step, tokens, batch, output)
 
+            # invoke callback for streaming output if provided
+            if step_callback is not None:
+                step_callback(step, output)
+
         return output
```
grassesi, line 649: I like this design: having a hook into the forecasting loop is super flexible and can be also used in the future to facilitate all kinds of interactions.

src/weathergen/train/trainer.py
grassesi, line 592: I dont think we should make a distinction here: I would consider writing all at once / no streaming a special case of the streaming writer.

src/weathergen/utils/validation_io.py
```diff
@@ -131,43 +42,277 @@ def write_output(
...
+    assert len(outputs_physical) == 1
+    target_aux_out_physical = target_aux_out[outputs_physical[0]]
+
+    # Check if streaming is enabled
+    streaming_cfg = val_cfg.get("output", {}).get("streaming", {})
+    streaming_enabled = streaming_cfg.get("enabled", False)
+
+    timestep_idxs = [0] if len(batch.get_output_idxs()) == 0 else batch.get_output_idxs()
+
+    if streaming_enabled:
```
grassesi: I would not make a distinction here, and consider "standart non-streaming" writing a special case (where the number of steps to write in chunk == number of forecast steps). This can be then handled within the StreamingOutputWriter

src/weathergen/utils/validation_io.py
```diff
@@ -131,43 +42,277 @@ def write_output(
...
+            self.geoinfo_channels,
+            self.sample_start,
+        ) = _prepare_batch_data(cf, val_cfg, batch_size, batch_idx, batch)
+
+        # Streaming config
+        self.streaming_cfg = val_cfg.get("output", {}).get("streaming", {})
+        self.write_freq = (
+            self.streaming_cfg.get("num_steps", 1) if self.streaming_cfg.get("num_steps") else 1
+        )
+
```
grassesi: if streaming_cfg.num_steps == validation_cfg.output.num_steps, the non-streaming case is already handled.

grassesi (changes requested):
Thanks for this implementation, I really like the general design however I have two things I think will streamline this PR alot:
- Lets consider the non-streaming case a special case of streaming where n_forecast_steps == n_steps_to_accumulate_before_write. This way we get rid of the decision branches in `write_output` and `Trainer.validate`, reducing the overall complexity.
- Lets simplify the configuration: Adding one new option (the number of fsteps to accumulate before writing) to `validation_config.output` / `training_config.output` should suffice:
  - `output.fstep_chunk_size: 1` => current default behaviour
  - `output.fstep_chunk_size: n` => accumulate n chunks before writing
  - `output.fstep_chunk_size: $training_config.forecast.num_steps` => no streaming
- Finally I think the default for this new argument should be None and if it is None then we dont do streaming (set `output.fstep_chunk_size = training_config.forecast.num_steps`)

If you dont mind I would already take this design and provide a callback function for `model.forward` in the new io (#1908 )

clessig (commented):
The model forward function should not have callbacks. The control flow should be with the trainer, and the model forward just executes. Having an IO operation (which potentially is very slow, with a very large variance per worker) in forward will likely cause all kinds of havoc with model parallelism because it leads to highly asynchronous execution.

evenmn:
> The model forward function should not have callbacks. The control flow should be with the trainer, and the model forward just executes. Having an IO operation (which potentially is very slow, with a very large variance per worker) in forward will likely cause all kinds of havoc with model parallelism because it leads to highly asynchronous execution.

I didn't think about that, but you're totally right. I will move the logics to trainer.py and push a new version

config/default_config.yml
evenmn: I agree, this is a more elegant solution

grassesi:
> The model forward function should not have callbacks. The control flow should be with the trainer, and the model forward just executes. Having an IO operation (which potentially is very slow, with a very large variance per worker) in forward will likely cause all kinds of havoc with model parallelism because it leads to highly asynchronous execution.

The execution of the forecasting loop has to to be interrupted by io, since otherwise otherwise predictions will accumulate in memory. So I dont see a implemetation without asynchronous-parallel io, that does not suffer from the same concerns regarding model parallelism/asynchronous execution. Having a callback is a nice and simple interface, alternatively the forecasting-loop would need to moved outside the forward function, or the forward function would need to be turned in some kind of yielding Generator (but this would probably require more logic in the loss calculation/optimization )

evenmn:
I implemented this functionality in a fundamentally different way from the one proposed here. I therefore closed this PR, and opened a new one here: #2076

PR #1949, opened 2026-02-27
added new stream_directory (config_forecasting_inference) to set max_…

ankitpatnala:
New config.yml where num_target is set to -1

### DIFF
--- a/config/config_forecasting_finetuning.yml
+++ b/config/config_forecasting_finetuning.yml
@@ -90,7 +90,6 @@ norm_type: "LayerNorm"
 
 #####################################
 
-streams_directory: "./config/streams/era5_1deg_forecasting/"
 streams: ???
 
 # type of zarr_store
--- a/config/streams/era5_1deg_forecasting_inference/era5.yml
+++ b/config/streams/era5_1deg_forecasting_inference/era5.yml
@@ -1,0 +1,105 @@
+# (C) Copyright 2024 WeatherGenerator contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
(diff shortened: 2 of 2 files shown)


MatKbauer (changes requested):
The era5 config looks good. Let's also add a `config_forecasting_inference.yml`, which specifies 40 rollout steps per default and sets the `streams_directory` to `./config/streams/era5_1deg_forecasting_inference`. Then we can launch inference conveniently with, e.g.,
```
uv run inference --from-run-id {RUN_ID} --samples 16 --overwrite_config config/config_forecasting_inference.yml
```

Furthermore, can you modify the `config_forecasting_finetuning.yml` to not contain a `streams_directory` field anymore?

ankitpatnala:
> The era5 config looks good. Let's also add a `config_forecasting_inference.yml`, which specifies 40 rollout steps per default and sets the `streams_directory` to `./config/streams/era5_1deg_forecasting_inference`. Then we can launch inference conveniently with, e.g.,
> 
> ```
> uv run inference --from-run-id {RUN_ID} --samples 16 --overwrite_config config/config_forecasting_inference.yml
> ```
> 
> Furthermore, can you modify the `config_forecasting_finetuning.yml` to not contain a `streams_directory` field anymore?

I feel the suggestion by Christian ws nice. We can launch inference now with 
`uv run inference --from-run-id {RUN_ID} --samples 16 --options streams_directory=./config/streams/era5_1deg_forecasting_inference training_config.forecast_num_steps=40`

MatKbauer (approved):
Agree, we don't necessarily need a dedicated `config_forecast_inference.yml`. Thanks for tackling this, @ankitpatnala.

PR #1951, opened 2026-02-27
Enable per stream masking config override

### LINKED ISSUE #1950: Enable per-stream masking

Enable per-stream masking but overwriting the masking strategy config on a per-stream basis.

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation

shmh40:
## Description

Enables per-stream masking config override.

## Issue Number

Closes #1950 


src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -245,12 +247,55 @@ def __init__(
...
 
+        self._effective_masking_cfgs = self._build_effective_masking_cfgs()
+
         self.mini_epoch = 0
 
         self.rng = None
         self.perms = None
         self.perms_num_forecast_steps = None
 
+    def _build_effective_masking_cfgs(self) -> dict[StreamName, Config]:
```
clessig: Shouldn't this be in Masker?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -245,12 +247,55 @@ def __init__(
...
+        deep-copy ``model_input`` and ``target_input`` from ``mode_cfg`` and
+        merge the overrides into each strategy entry's ``masking_strategy``
+        and ``masking_strategy_config``.
+
+        Only ``masking_strategy`` and ``masking_strategy_config`` are
+        overridable — structural keys (``num_samples``, ``num_steps_input``)
+        stay consistent across streams for batch assembly.
+        """
+        cfgs: dict[StreamName, Config] = {}
+        for stream_info in self.streams:
```
clessig: I think the logic here should be: 

for each stream:
  stream_merged_masking_config = config.merge( masking_config, stream_config.masking_override)

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -575,21 +620,26 @@ def _get_data_windows(self, base_idx, num_forecast_steps, num_steps_input_max, s
 
     def _get_source_target_masks(self, training_mode):
         """
-        Generate source and target masks for all streams
+        Generate source and target masks for all streams.
+
+        Each stream uses its own effective masking config (which may include
```
clessig: I don't think this should be here (would make sense if the masking_config would be passed as an argument, which is a sensible possibility). Can be in line 631.

shmh40:
Thanks for the comments @clessig. Agreed it should be handled in masking and I have tried to do this with only minor changes to msds. Hopefully better now. We can think if we want it to be more general but I am ok with this for now.

I directly merged your PR in too. I haven't come up with a neater solution to the source/target distinction but can revisit it next week if helpful.

src/weathergen/datasets/masking.py
```diff
@@ -129,6 +130,48 @@ def reset_rng(self, rng) -> None:
...
+    def merge_masking_config(mode_cfg, override):
+        """Merge a stream's masking override into the base mode config.
+
+        Only masking strategy fields are overridden. Structural keys like
+        ``num_samples`` and ``num_steps_input`` remain unchanged.
+        """
+        if override is None:
+            return mode_cfg
+
+        effective = copy.deepcopy(mode_cfg)
```
clessig: Can we use a better variable name than `effective`, e.g. `stream_cfg_masking` seem appropriate

src/weathergen/datasets/masking.py
```diff
@@ -129,6 +130,48 @@ def reset_rng(self, rng) -> None:
...
+        for section_key in ("model_input", "target_input"):
```
clessig, line 180: Could you explain the logic for this part

clessig: Also, how do you expect the overrides in the stream config to look like?

src/weathergen/datasets/masking.py
```diff
@@ -288,10 +331,15 @@ def build_samples_for_stream(
...
                 else:
+                    # prevent target from being dropped
+                    stream_cfg_target = copy.deepcopy(stream_cfg)
+                    stream_cfg_target["randomly_drop_as_source_rate"] = 0.0
+                    # get
                     target_mask, mask_params = self._get_mask(
                         num_cells=num_cells,
                         strategy=target_cfg.get("masking_strategy"),
                         masking_strategy_config=target_cfg.get("masking_strategy_config", {}),
+                        stream_cfg=stream_cfg_target,
```
clessig: The `masking_strategy_config` that is passed here should be the consolidated one. Then `stream_cfg` should not be needed here. See also the comment above.

src/weathergen/datasets/masking.py
```diff
@@ -344,6 +392,7 @@ def build_samples_for_stream(
                         num_cells=num_cells,
                         strategy=source_cfg.get("masking_strategy"),
                         masking_strategy_config=masking_config,
+                        stream_cfg=stream_cfg,
```
clessig: See above. Do we really need it?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -575,16 +580,18 @@ def _get_data_windows(self, base_idx, num_forecast_steps, num_steps_input_max, s
...
         """
-        Generate source and target masks for all streams
+        Generate source and target masks for all streams.
         """
-
         masks = {}
         for stream_info in self.streams:
+            # Each stream uses its own effective masking config (which may include
+            # per-stream ``masking_override`` merged on top of the global config).
+            stage_cfg = self._effective_masking_cfgs[stream_info["name"]]
```
clessig: stage_cfg -> stream_cfg -- nothing related to stage here.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -243,7 +243,12 @@ def __init__(
             else cf.data_loading.rng_seed * 97
         )
 
-        self.tokenizer = TokenizerMasking(cf.healpix_level, Masker(cf.healpix_level, stage))
+        self.masker = Masker(cf.healpix_level, stage)
+        self.tokenizer = TokenizerMasking(cf.healpix_level, self.masker)
+
+        self._effective_masking_cfgs = self.masker.build_effective_masking_cfgs(
```
clessig: I am wondering if `_effective_masking_cfgs` cannot be kept in the Masker and constructed in the constructor?

clessig (commented):
Left some more comments. We need an example how the overrides look like but for documentation and to ensure things are handled correctly. I also made some more suggestions to further improve encapsulation.

src/weathergen/datasets/masking.py
shmh40: Yes I think so too. Just checking -- this is done currently because we need to pass through the randomly_drop_as_source_rate? So we should put the randomly_drop in build_samples_from_stream, and then we can remove stream_cfg and stream_cfg_target deepcopy?

src/weathergen/datasets/multi_stream_data_sampler.py
shmh40: This is possible, I don't think it makes too much difference though?

src/weathergen/datasets/masking.py
clessig: Where do we have `build_samples_from_stream`? `randomly_drop_as_source_rate` should be in the consolidated masking config.

src/weathergen/datasets/multi_stream_data_sampler.py
clessig: It helps encapsulation. MultiStreamDataSampler is complex enough.

src/weathergen/datasets/masking.py
```diff
@@ -336,9 +417,12 @@ def build_samples_for_stream(
                 # target is specified)
                 target_idx += i_sample % target_num_samples[target_cfg_idx].item()
 
-                # determine if forcing dataset => mask is empty
+                # determine if diagnostic dataset => mask is empty
                 if is_stream_diagnostic(stream_cfg, self.stage):
                     source_mask, mask_params = torch.zeros(num_cells, dtype=torch.bool), {}
+                elif randomly_drop_rate > 0.0 and self.rng.uniform() < randomly_drop_rate:
```
clessig: The two branches are identical so they should be merged. To make the code more readable, one can compute the condition first, e.g.

```
is_stream_dropped = randomly_drop_rate > 0.0 and self.rng.uniform() < randomly_drop_rate

if is_stream_diagnostic(stream_cfg, self.stage) or is_stream_dropped:
    source_mask, mask_params = torch.zeros(num_cells, dtype=torch.bool), {}
```

src/weathergen/datasets/masking.py
```diff
@@ -427,7 +511,10 @@ def _get_mask(
         return (mask, params)
 
     def _generate_cell_mask(
-        self, num_cells: int, strategy: str, masking_strategy_config: dict
+        self,
```
clessig, line 518: This should fit again in one line (linters are stupid and they only add lines, never remove :))

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -575,16 +576,18 @@ def _get_data_windows(self, base_idx, num_forecast_steps, num_steps_input_max, s
...
         """
-        Generate source and target masks for all streams
+        Generate source and target masks for all streams.
         """
-
         masks = {}
         for stream_info in self.streams:
+            # Each stream uses its own effective masking config (which may include
+            # per-stream ``masking_override`` merged on top of the global config).
+            stream_cfg = self.masker._effective_masking_cfgs[stream_info["name"]]
```
clessig: Why do we still need this here? Is it not directly available in the masker where it is needed?

Also `stream_cfg` is a bit misleading and I would call it `stream_masking_cfg`

src/weathergen/datasets/masking.py
```diff
@@ -123,12 +124,88 @@ def __init__(self, healpix_level: int, stage: Stage):
...
+            self._effective_masking_cfgs = {}
+
     def reset_rng(self, rng) -> None:
         """
         Reset rng after mini_epoch to ensure proper randomization
         """
         self.rng = rng
 
+    @staticmethod
+    def merge_masking_config(mode_cfg, override):
```
clessig: I don't think this should be static.

src/weathergen/datasets/masking.py
```diff
@@ -123,12 +124,88 @@ def __init__(self, healpix_level: int, stage: Stage):
...
+                    strategy_cfg["masking_strategy"] = override_values["masking_strategy"]
+                if "masking_strategy_config" in override_values:
+                    strategy_cfg["masking_strategy_config"] = omegaconf.OmegaConf.merge(
+                        strategy_cfg.get("masking_strategy_config", omegaconf.OmegaConf.create({})),
+                        override_values["masking_strategy_config"],
+                    )
+
+        return stream_cfg_masking
+
+    @staticmethod
```
clessig: I don't think this should be static.

clessig (approved):
Minor comments but please address before merging and let me know if we should disuss anything.

shmh40:
To activate randomly_drop_as_source_rate, or to override the masking_strategy_config for a particular stream, an example snippet to be included in the stream config is below:

Example usage:

```
ERA5_smooth :
  type : anemoi
  filenames : ['aifs-ea-an-oper-0001-mars-o96-1979-2023-6h-v8.zarr']
  stream_id : 0
  source : ['2d', '2t', 'msl', 'skt', 'sp']
  target : ['2d', '2t', 'msl', 'skt', 'sp']
  ...
  masking_override :
    randomly_drop_as_source_rate: 0.9
    target_input :
      masking_strategy_config :
        hl_mask : 3
  embed:
     ...
```

shmh40:
NB: if multiple strategies are included, e.g. for JEPA with JEPA loss *and* a physical loss, the random dropping of streams is applied independently per source sample.

src/weathergen/datasets/masking.py
shmh40: Linter didn't want one line! :(

clessig (approved):
Much cleaner now. Thanks for cleaning up.

PR #1965, opened 2026-03-02
Export IMERG in evaluation

### LINKED ISSUE #1944: Enable exporting IMERG data

We would like to export and regrid some IMERG precipitation data for further analysis. So far, the export sub-package in evaluation can only deal with ERA5. Adding this channel should be straightforward since it is also on an O96 grid. 

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation

s6sebusc:
## Description

Some small changes needed to enable exporting O96 IMERG_ANEMOI data. Also includes
- a slight change in the regridder that should speed it up considerably
- ~a small fix to the reshape function that should enable exporting single pressure levels~ *we'll do it in another issue*

## Issue Number

Closes #1944


packages/evaluate/src/weathergen/evaluate/export/parsers/netcdf_parser.py
```diff
@@ -128,7 +130,7 @@ def reshape(self, data: xr.DataArray) -> xr.Dataset:
         data_vars = {}
 
         for new_var, old_vars in var_dict.items():
-            if len(old_vars) > 1:
+            if len(pl) > 0:
```
enssow: i think this is the line causing the problem with the ordering of the ipoint and pressure_level. without checking the old_vars, it is using pl which is common between all data sets, and making surface variables with 1 pressure_level e.g. 
```Dimensions without coordinates: ipoint, pressure_level, np.str_('2t'): <xarray.DataArray (ipoint: 40320, pressure_level: 1)> Size: 161kB
array([[298.],
       [298.],
       [298.],
       ...,
       [300.],
       [302.],
       [300.]], shape=(40320, 1), dtype=float32)
```

enssow: I guess this was the change to fix the bug about single levels - i think to get this in quicker, perhaps better to revert this change and merge the PR before opening a seperate issue to fix the single levels? let me know what you think

enssow (changes requested):
There are some new changes to be merged from develop but looks great (plus comment about the pressure_levels in reshape)

Tested on both IMERG and ERA5 inference runs. Regridding functionality not able to test at the moment due to disk issues on JUWELS (eathkit needs to download matrices from a cloud database)

packages/evaluate/src/weathergen/evaluate/export/parsers/netcdf_parser.py
s6sebusc: rolled it back, I'll open a new issue for that soon

enssow (approved):
new changes approves, un requesting review hopefully?

PR #1997, opened 2026-03-06
Add changes for pushforward trick

### LINKED ISSUE #1740: Run pushforward trick experiments

The idea is to apply loss function on the last forecast step rather applying to all intermediate steps
The experiments will be conducted using both [fe-param-search](https://github.com/ecmwf/WeatherGenerator/tree/mk/develop/fe-param-search)  branch and `develop` branch.

_No response_

### Area

- [ ] model
- [x] science
- [ ] documentation

SavvasMel:
## Description

This PR adds the necessary changes for the pushforward trick.

## Issue Number

Fixes https://github.com/ecmwf/WeatherGenerator/issues/1740

See Hedgedoc: https://gitlab.jsc.fz-juelich.de/hedgedoc/bGdOXZgDTTyX074RRvliWg?both


src/weathergen/model/model_interface.py
```diff
@@ -111,7 +111,10 @@ def init_model_and_shard(
 
         for module in model.forecast_engine.fe_blocks.modules():
             if isinstance(module, modules_to_shard):
-                fully_shard(module, **fsdp_kwargs)
+                # reshard_after_forward=False keeps FE parameters unsharded
+                # during the multi-step rollout loop.
+                # Needed for pushforward trick.
+                fully_shard(module, reshard_after_forward=False, **fsdp_kwargs)
```
clessig, line 113: @sophie-xhonneux : is this maybe related to the problem we are seeing with the EMATeacher where we need to reshard?

src/weathergen/model/model.py
```diff
@@ -695,15 +695,34 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
...
+                    tokens = self.forecast_engine(tokens, step, coords=model_params.rope_coords)
+                    output = self.predict_latent(model_params, step, tokens, batch, output)
+                    output = self.predict_decoders(model_params, step, tokens, batch, output)
+                else:
+                    # Pushforward mode: only add empty predictions (no grad, no engine call)
+                    with torch.no_grad():
+                        tokens = self.forecast_engine(tokens, step, model_params.rope_coords)
+
+                    # Add empty predictions for all streams (vectorized / batched if possible)
+                    for stream_name in self.stream_names:
```
clessig: We can avoid this when we create `output` with the correct length. I thought we do this anyway, and hence the `step` argument.

src/weathergen/model/model.py
```diff
@@ -695,15 +695,34 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
...
-            # latent predictions (raw and with SSL heads)
-            output = self.predict_latent(model_params, step, tokens, batch, output)
+                needs_full_prediction = (
+                    not pushforward_trick
+                    or not self.training
+                    or step == max(batch.get_output_idxs())
+                )
+
+                if needs_full_prediction:
+                    tokens = self.forecast_engine(tokens, step, coords=model_params.rope_coords)
```
clessig: Remove `coords=` for consistency

src/weathergen/model/model.py
```diff
@@ -695,15 +695,34 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
...
-            output = self.predict_decoders(model_params, step, tokens, batch, output)
-            # latent predictions (raw and with SSL heads)
-            output = self.predict_latent(model_params, step, tokens, batch, output)
+                needs_full_prediction = (
+                    not pushforward_trick
+                    or not self.training
+                    or step == max(batch.get_output_idxs())
+                )
+
+                if needs_full_prediction:
```
clessig: We already call the forecast engine in l702. Don't we call it twice then in one iteration. 

Also, if we set `self.forecast_engine` to Identity if the number of blocks is 0 then we avoid the condition above.

src/weathergen/model/model.py
```diff
@@ -695,15 +695,34 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
...
         for step in batch.get_output_idxs():
             # apply forecasting engine (if present)
             if self.forecast_engine:
                 tokens = self.forecast_engine(tokens, step, coords=model_params.rope_coords)
 
-            # decoder predictions
-            output = self.predict_decoders(model_params, step, tokens, batch, output)
-            # latent predictions (raw and with SSL heads)
-            output = self.predict_latent(model_params, step, tokens, batch, output)
+                needs_full_prediction = (
```
clessig: If you choose a bit more compact variable names then we can fit this in one line and it's more readable.

src/weathergen/model/model.py
```diff
@@ -695,15 +695,34 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
...
+                    not pushforward_trick
+                    or not self.training
+                    or step == max(batch.get_output_idxs())
```
clessig: You use this to determine the number of forecast steps?

src/weathergen/model/model.py
clessig: Eg. `pushforward` instead of `pushforward_trick`

src/weathergen/model/model.py
SavvasMel: I use this to determine the last step for which we will take gradients.

src/weathergen/model/model.py
SavvasMel: 1) Sorry the one in l702 must have been introduced due to wrong merge, I will delete it.

2) Many thanks for the heads-up, I will correct this too.

SavvasMel: @clessig just to make sure that I understand, if blocks are not >0 then the forecasting engine turns to None. Should I introduce an identity function instead in this case?

clessig: Yes exactly

src/weathergen/model/engines.py
```diff
@@ -390,6 +390,17 @@ def forward(self, tokens, coords=None):
         return tokens
 
 
+class IdentityEngine(torch.nn.Module):
+    """Identity engine that passes tokens through unchanged."""
+
+    def __init__(self):
+        super().__init__()
+        self.fe_blocks = torch.nn.ModuleList()
```
clessig, line 537: `self.blocks = ..`. Also do we need this as all? If anything I would expect `self.blocks = torch.nn.Identity()` but since we implement forward it might not be needed

src/weathergen/model/model.py
```diff
@@ -697,16 +697,20 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
         # collapse along input step dimension
         tokens = tokens.reshape(shape).sum(axis=1)
 
+        # Allow for pushforward trick
+        p_fwd = self.cf.get("pushforward_trick", False)
```
clessig: I would expect that the push-forward trick config is part of forecast, i.e.

```
training_config : 

  .... 

  forecast: 
    num_steps: 3
    push_forward: True
```

clessig (commented):
Essentially ready to merge, just some minor details.

src/weathergen/model/engines.py
SavvasMel: I would like to keep it consistent:

```
class ForecastingEngine(torch.nn.Module):
    name: "ForecastingEngine"

    def __init__(self, cf: Config, mode_cfg, num_healpix_cells: int, dim_aux: int = None) -> None:
        """
        Initialize the ForecastingEngine with the configuration.

        :param cf: Configuration object containing parameters for the engine.
        :param num_healpix_cells: Number of healpix cells used for local queries.
        """
        super(ForecastingEngine, self).__init__()
        self.cf = cf
        self.num_healpix_cells = num_healpix_cells
        self.fe_blocks = torch.nn.ModuleList()        <-------------------
```

src/weathergen/model/model.py
SavvasMel: Done.

src/weathergen/model/engines.py
SavvasMel: Also it allows self.fe_blocks to be iterable.

MatKbauer: Looks clean to me, but also wonder whether we need it. Are there cases where we want to iterate over a non-existing forecast engine?

src/weathergen/model/model.py
```diff
@@ -704,16 +704,20 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
...
+            all_grad = not p_fwd or not self.training or step == max(batch.get_output_idxs())
+
+            if all_grad:
+                tokens = self.forecast_engine(tokens, step, model_params.rope_coords)
+                output = self.predict_latent(model_params, step, tokens, batch, output)
+                output = self.predict_decoders(model_params, step, tokens, batch, output)
+            else:
+                # Pushforward mode: advance tokens without grad
+                with torch.no_grad():
+                    tokens = self.forecast_engine(tokens, step, model_params.rope_coords)
```
MatKbauer:
```suggestion
            if self.forecast_engine:
                without_grad = not p_fwd or not self.training or step == max(batch.get_output_idxs())
                if without_grad:
                    # Pushforward mode: advance tokens without grad
                    with torch.no_grad():
                        tokens = self.forecast_engine(tokens, step, coords=model_params.rope_coords)
                        continue  # no decoding in pushforward mode
                tokens = self.forecast_engine(tokens, step, coords=model_params.rope_coords)

            # decoder predictions
            output = self.predict_decoders(model_params, step, tokens, batch, output)
            # latent predictions (raw and with SSL heads)
            output = self.predict_latent(model_params, step, tokens, batch, output)
```

MatKbauer: I'd suggest to leave the forecast loop unchanged and call a `continue` statement in pushforward mode to not decode.

MatKbauer (changes requested):
Pushforward logic is clean, but I advocate for keeping the forecast loop unchanged; see suggestion in the `forward()` function.

Also, I think we won't need the `IdentityEngine` and suggest to revert to the former code structure. With the suggested implementation of the pushforward logic, we would not need the explicit `IdentityEngine`.

src/weathergen/model/engines.py
SavvasMel: It might be needed in the case we want to export the latent vectors even if FE is not present?

src/weathergen/model/model.py
SavvasMel: Since it has been decided to include an identity engine, the condition statement does not seem necessary. Also for the steps where we do not take gradients the output variable has automatically distributed empty lists for those steps in the initialization. I think the current state is more clean.

Any other suggestions/thoughts @clessig @MatKbauer ?

MatKbauer: Agree on the identity engine. Concerning the order of calling the forecast engine, though, I found it more intuitive to leave the original order of `forecast_engine`, `predict_decoders`, `predict_latent` and add the push-forward mode before. Such as

```python
            for step in batch.get_output_idxs():

                without_grad = not p_fwd or not self.training or step == max(batch.get_output_idxs())
                if without_grad:
                    # Pushforward mode: advance tokens without grad
                    with torch.no_grad():
                        tokens = self.forecast_engine(tokens, step, coords=model_params.rope_coords)
                        continue  # no decoding in pushforward mode

                # Original code follows here, i.e.,
                tokens = self.forecast_engine(tokens, step, coords=model_params.rope_coords)
                # decoder predictions
                output = self.predict_decoders(model_params, step, tokens, batch, output)
                # latent predictions (raw and with SSL heads)
                output = self.predict_latent(model_params, step, tokens, batch, output)
```

MatKbauer: This flow seems more approachable and readable to me.

SavvasMel: Ok I will do the change, in that case, it should be:

`without_grad = p_fwd and self.training and step != max(batch.get_output_idxs())`

src/weathergen/model/model.py
```diff
@@ -704,15 +704,20 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
...
+        p_fwd = self.cf.training_config.get("forecast", {}).get("pushforward", False)
         # roll-out in latent space, iterate and generate output over requested output steps
         for step in batch.get_output_idxs():
-            # apply forecasting engine (if present)
-            if self.forecast_engine:
-                tokens = self.forecast_engine(tokens, step, coords=model_params.rope_coords)
+            without_grad = p_fwd and self.training and step != max(batch.get_output_idxs())
+            if without_grad:
+                # Pushforward mode: advance tokens without grad
+                # Pushforward mode: no decoding in this mode
```
MatKbauer: Let's reduce this to a single comment. Something like `# Pushforward mode: advance tokens without grad; no decoding in this mode`

src/weathergen/model/model.py
```diff
@@ -704,15 +704,20 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
...
+                with torch.no_grad():
+                    tokens = self.forecast_engine(tokens, step, model_params.rope_coords)
+                    continue
 
-            # decoder predictions
+            tokens = self.forecast_engine(tokens, step, model_params.rope_coords)
             output = self.predict_decoders(model_params, step, tokens, batch, output)
-            # latent predictions (raw and with SSL heads)
             output = self.predict_latent(model_params, step, tokens, batch, output)
```
MatKbauer, line 721: I find the two comments for the decoder (physical) and latent predictions useful. Can we keep them in?

MatKbauer (approved):
This looks great, thanks, Savvas! If we can still iterate once more over the comments in the forecast function, I'd be grateful.

PR #2076, opened 2026-03-18
[1914][1596] Incremental Model forward function

### LINKED ISSUE #1596: Forecast output needs to be written incrementally

### LINKED ISSUE #1914: Incremental forward function for long-term roll-out

evenmn:
## Description

Added support for running the Model forward method incrementally. This allows for writing to file during inference rollouts, which again makes long inference rollouts possible. 

Changes include:
- `Model.forward()` can now take a `ModelOutput` object as batch input, and ignore the encoding for that type of objects
- `Trainer.validate()` sets up a chunk list, where the elements are the number of rollout steps in each increment. We loop through this chunk list and potentially call Model.forward() multiple times

The chunking behavior is controlled by a config option `forecast.fstep_chunk_size`, which determines the maximum chunk size. 

Example: is `forecast.fstep_chunk_size: 3` and `forecast.num_steps: 10`, `Trainer` will create the chunks `[3,3,3,1]` and call `Model.forward()` 4 times incrementally.

## Issue Number

Closes #1914
Closes #1596


packages/common/src/weathergen/common/config.py
```diff
@@ -647,6 +647,11 @@ def load_streams(streams_directory: Path) -> list[Config]:
 
 def get_path_run(config: Config) -> Path:
     """Get the current runs results_path for storing run results and logs."""
+    if config is not None and config.get("results_path", None):
+        base = Path(config.results_path)
+        if not base.is_absolute():
+            base = (Path.cwd() / base).resolve()
+        return base / get_run_id_from_config(config)
```
MatKbauer: Can you explain why we need this conversion? Don't we get the right path from calling `_get_shared_wg_path()` below?

packages/common/src/weathergen/common/io.py
MatKbauer, line 1: @iluise or @SavvasMel, could either of you please verify the integrity of the changes in `io.py`? I'm not familiar with this part.

src/weathergen/train/trainer.py
```diff
@@ -591,32 +611,64 @@ def validate(self, mini_epoch, mode_cfg, batch_size):
...
-                            self.cf,
-                            mode_cfg,
-                            batch_size,
-                            mini_epoch,
-                            bidx,
-                            denormalize_data_fct,
-                            batch,
-                            preds,
-                            targets_and_auxs,
+                        for chunk_idx, chunk_size in enumerate(chunks):
```
MatKbauer: Can we modularise the changes introduced here to improve readability? Some function like `writed_chunked_output()` or such would be great. This should save some indentations and make function calls fit into single lines.

src/weathergen/train/trainer.py
```diff
@@ -438,9 +459,11 @@ def train(self, mini_epoch):
                 dtype=self.mixed_precision_dtype,
                 enabled=cf.with_mixed_precision,
             ):
+                x = batch.get_source_samples()
```
MatKbauer: Let's please rename `x` to `source_samples` (applies to various parts also below)

src/weathergen/utils/validation_io.py
```diff
@@ -56,8 +68,9 @@ def write_output(
 
             # handle spoof data: do not write since it might corrupt validation (spoofing invisible
             # there)
+            pred_idx = t_idx - fstep_offset
```
MatKbauer: Let's use `t_chunk_idx` instead of `pred_idx`

src/weathergen/model/model.py
```diff
@@ -21,7 +21,7 @@
 from torch.utils.checkpoint import checkpoint
 
 from weathergen.common.config import Config
-from weathergen.datasets.batch import ModelBatch
+from weathergen.datasets.batch import BatchSamples, ModelBatch
```
MatKbauer, line 25: Can we use `SampleBatch` instead of `BatchSamples` to be consistent with `ModelBatch`?

MatKbauer (changes requested):
Thanks, @evenmn, this looks good to me! I have added some suggestions to modularise the code and change namings to be more explicit. Once this is in place, let's merge this and afterwards proceed with [@clessig's suggestions](https://github.com/ecmwf/WeatherGenerator/issues/2105#issuecomment-4133213894) in @ankitpatnala's issue #2105. But these will come with a separate PR.

evenmn:
Thanks for your feedback @MatKbauer, I will rename the variables as suggested :) The implementation hasn't been properly tested yet. I have run the small integration test with 500 rollout steps and the test passed. The next step will be to do a full scale training for 64 epochs and then run inference for 1000 rollout steps (or 1 year) and verify the results

MatKbauer:
Rolling out to 1000 steps or one year would be great. Not sure whether we need a model that was trained for 64 epochs, though. We can use an existing model, e.g., try the `atmofs02` release model or `he7ut6vx` (which is the pre-training of the current forecast model candidate on JUPITER).

src/weathergen/model/model.py
```diff
@@ -685,26 +718,35 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
...
-        shape = (len(batch), batch.get_num_steps(), *tokens.shape[1:])
-        # collapse along input step dimension
-        tokens = tokens.reshape(shape).sum(axis=1)
+            # recover batch dimension and separate input_steps
+            shape = (len(batch_ctx), batch_ctx.get_num_steps(), *tokens.shape[1:])
+            # collapse along input step dimension
+            tokens = tokens.reshape(shape).sum(axis=1)
+        else:
+            if latent_state is None or latent_state.z_pre_norm is None:
+                raise ValueError("LatentState.z_pre_norm must be provided to run the model.")
```
grassesi: `assert latent_state and latent_state.z_pre_norm, "LatentState.z_pre_norm must be provided to run the model."` is more concise and more in line with the rest of the code here.

packages/common/src/weathergen/common/config.py
grassesi: `"results_path"` option is as far as I know legacy/deprecated. There should no code be added to explicitly support this. Please remove.

config/default_config.yml
```diff
@@ -225,6 +225,8 @@ training_config:
       num_steps: 2
       offset: 1
       policy: "fixed"
+      # number of forecast steps to accumulate before writing to disk
+      fstep_chunk_size: 1
```
grassesi: Should be optional, see my comment on `config.py`. For documentations sake it can be set to `null`.

src/weathergen/train/trainer.py
```diff
@@ -176,6 +182,21 @@ def get_target_aux_calculators(self, mode_cfg):
...
+        return [chunk_size] * (rollout_steps // chunk_size) + (
+            [rollout_steps % chunk_size] if rollout_steps % chunk_size else []
+        )
+
+    def _get_fcst_chunks(self, mode_cfg):
+        if mode_cfg is self.test_cfg:
+            return self.fcst_chunks_by_mode["test"]
+        if mode_cfg is self.training_cfg:
+            return self.fcst_chunks_by_mode["training"]
+        return self.fcst_chunks_by_mode["validation"]
```
grassesi: The code gets much simpler if you dont store the chunk sizes as a member variable. Recomputing is not a big effort.

```python
_get_forecast_chunks(self, forecast_cfg):
    n_full_chunks = forecast_cfg.num_steps // forecast_cfg.chunk_size
    remainder_fsteps = forecast_cfg.num_stpes % forecast_cfg.chunk_size
    return [forecast_cfg.chunk_size] * n_full_chunks + ([remainder_fsteps] if remainder_fsteps else [])
```

src/weathergen/train/trainer.py
```diff
@@ -571,17 +594,14 @@ def validate(self, mini_epoch, mode_cfg, batch_size):
...
-                                self.model_params,
-                                batch.get_source_samples(),
-                            )
-                        else:
-                            preds = self.ema_model.forward_eval(
-                                self.model_params,
-                                batch.get_source_samples(),
-                            )
-
+                        chunks = self._get_fcst_chunks(mode_cfg)
```
grassesi: `self._get_forecast_chunks(mode_cfg.forecast)` (See above)

src/weathergen/train/trainer.py
```diff
@@ -157,6 +158,11 @@ def init(self, cf: Config, devices):
         # Initialize collapse monitor for SSL training
         collapse_config = cf.train_logging.get("collapse_monitoring", {})
         self.collapse_monitor = CollapseMonitor(collapse_config, None)  # device set later in run()
+        self.fcst_chunks_by_mode = {
+            "training": self._init_fcst_chunks(self.training_cfg),
+            "validation": self._init_fcst_chunks(self.validation_cfg),
+            "test": self._init_fcst_chunks(self.test_cfg),
+        }
```
grassesi: Please remove (see below)

packages/common/src/weathergen/common/config.py
```diff
@@ -709,6 +714,11 @@ def validate_forecast_policy_and_steps(forecast_cfg: OmegaConf, mode: str):
     if len(forecast_cfg) == 0:
         return
 
+    if "fstep_chunk_size" in forecast_cfg:
+        fstep_chunk_size = forecast_cfg.get("fstep_chunk_size")
+        if not isinstance(fstep_chunk_size, int) or fstep_chunk_size <= 0:
+            raise TypeError(f"'{mode}.forecast.fstep_chunk_size' must be a positive integer.")
+
```
grassesi: I think `fstep_chunk_size` should be an optional parameter: If None the default behaviour should be the previous one (do all forecast steps in one chunk). I think it also suffices if the parameter is only called `chunk_size` since it clear from the context (beeing part of `forecast_cfg`) that it relates to forecasting/rollout. Also do the bounds checking within a assert statement to keep stylistic consistency with the rest of the code.

```python
forecast_cfg.num_steps = forecast_cfg.get("num_steps", 1)
forecast_cfg.chunk_size = forecast_cfg.get("chunk_size", forecast_cfg.num_steps)
assert (
    isinstance(forecast_cfg.chunk_size, int) and forecast_cfg.chunk_size <= 0,
    f"'{mode}.forecast.fstep_chunk_size' must be a positive integer."
)
```

src/weathergen/train/trainer.py
```diff
@@ -591,32 +611,64 @@ def validate(self, mini_epoch, mode_cfg, batch_size):
...
+                                    fstep_offset=chunk_step_offset,
+                                )
+                            if preds_full is not None:
+                                for step_idx in range(chunk_size):
+                                    preds_full.physical[x.step_offset + step_idx].update(
+                                        x.physical[step_idx]
+                                    )
+                                    preds_full.latent[x.step_offset + step_idx].update(
+                                        x.latent[step_idx]
+                                    )
```
grassesi: I dont see how this solves the problem of accumulating data during long rollout: Even if the data is written incrementally to disk it still accumulates in `preds_full` (It is never cleared out).

src/weathergen/train/trainer.py
```diff
@@ -591,32 +611,64 @@ def validate(self, mini_epoch, mode_cfg, batch_size):
...
+                                    raise ValueError(
+                                        "Missing latent_state for chunked forecast continuation."
+                                    )
+
+                    if preds_full is not None:
+                        _ = self.loss_calculator_val.compute_loss(
+                            preds=preds_full,
+                            targets_and_aux=targets_and_auxs,
+                            metadata=extract_batch_metadata(batch),
                         )
```
grassesi: The loss calculator must be adapted so that it can operate on a per chunk basis, so that prediction data can actually be cleared out. The loss calculator might internally keep a running sum of the loss for all chunks.

src/weathergen/train/trainer.py
```diff
@@ -571,17 +594,14 @@ def validate(self, mini_epoch, mode_cfg, batch_size):
...
-                            )
-
+                        chunks = self._get_fcst_chunks(mode_cfg)
+                        x = batch.get_source_samples()
+                        total_steps = batch.get_output_len()
+                        preds_full = (
+                            ModelOutput(total_steps, batch=batch.get_source_samples())
+                            if total_steps > 0
+                            else None
+                        )
```
grassesi: From what I see `preds_full` stores the predictions of all chunks for the purpose of loss computation. This is an issue since the predictions dont get cleared out and accumualate in memory. Instead the loss computation should be adapted (see comments below), making `preds_full` needless here.

src/weathergen/train/loss_modules/loss_module_physical.py
```diff
@@ -220,8 +220,10 @@ def compute_loss(self, preds: dict, targets: dict, metadata) -> LossValues:
 
             # TODO: make nicer
             output_step_loss_weights = self._get_output_step_weights(len(targets.output_idxs))
-            if len(targets.physical) - len(targets.output_idxs) > 0:
-                output_step_loss_weights.insert(0, None)
+            # Pad with Nones for forecast offsets so indices align with targets.physical
+            output_offset = len(targets.physical) - len(targets.output_idxs)
+            if output_offset > 0:
+                output_step_loss_weights = [None] * output_offset + list(output_step_loss_weights)
```
grassesi, line 306: This is just a refactoring of the previous behavior right? If so nicely done :) (only `output_step_loss_weights` does not need to be converted to `list` => it is already list)

src/weathergen/model/model.py
```diff
@@ -685,26 +718,35 @@ def forward(self, model_params: ModelParams, batch: ModelBatch) -> ModelOutput:
...
+            tokens = latent_state.z_pre_norm
+            output.add_latent_prediction(0, "posteriors", latent_state)
```
grassesi: I dont like seeing BatchSamples and ModelOutput mixed together like this here. I find it highly confusing and fail to see the need for this: In this example I pass BatchSamples and ModelOutput separately for each chunk. It should be equivalent to your code and is shorter and simpler. Please tell me if I am overlooking something here 😅.

```python
def forward(self, model_params: ModelParams, batch: BatchSamples, rollout_steps: int, preds: ModelOutput | None = None):
    if preds is None:
        tokens, posteriors = self.encoder(model_params, batch)
        output.add_latent_prediction(0, "posteriors", posteriors)

        # recover batch dimension and separate input_steps
        shape = (len(batch.sources), batch.get_num_steps(), *tokens.shape[1:])
        # collapse along input step dimension
        tokens = tokens.reshape(shape).sum(axis=1)
        step_offset = 0
    else:
        posteriors = preds.get_last_latent_state()
        tokens = posteriors.z_pre_norm
        step_offset = preds.step_offset + len(preds.physical)

    output = ModelOutput(rollout_steps, step_offset=step_offset)
    output.add_latent_prediction(0, "posteriors", posteriors)

    ...

    return output
```

src/weathergen/model/model.py
```diff
@@ -80,6 +86,20 @@ def get_physical_prediction(
     def get_latent_prediction(self, fstep: int):
         return self.latent[fstep]
 
+    def get_last_latent_state(self) -> LatentState:
+        for latent_pred in reversed(self.latent):
+            latent_state = latent_pred.get("latent_state", None)
+            if latent_state is not None:
+                return latent_state
+
+        raise ValueError("ModelOutput does not contain a latent_state prediction to continue from.")
```
grassesi, line 94: nice

src/weathergen/model/model.py
```diff
@@ -80,6 +86,20 @@ def get_physical_prediction(
...
+
+    def get_batch_size(self) -> int:
+        if self.batch is not None:
+            return len(self.batch)
+
+        return self.get_last_latent_state().z_pre_norm.shape[0]
```
grassesi, line 100: I think this no longer needed if ModelOutput and BatchSamples are more clearly separated.

src/weathergen/train/trainer.py
```diff
@@ -591,32 +611,64 @@ def validate(self, mini_epoch, mode_cfg, batch_size):
...
+                                    self.model_params,
+                                    x,
+                                    chunk_size,
+                                )
+                            else:
+                                x = self.ema_model.forward_eval(
+                                    self.model_params,
+                                    x,
+                                    chunk_size,
+                                )
```
grassesi: See my comment on `Model.forward()`
```python
preds = None
for chunk_idx, chunk_size in enumerate(chunks):
    if self.ema_model is None:
        preds = self.model(
            self.model_params,
            source_samples,
            chunk_size,
            preds,
        )
    else:
        preds = self.ema_model.forward_eval(
            self.model_params,
            source_samples,
            chunk_size,
            preds,
        )
    
    ...
```

src/weathergen/model/model.py
```diff
@@ -53,10 +53,16 @@ class ModelOutput:
 
     physical: list[dict[StreamName, torch.Tensor]]
     latent: list[dict[str, torch.Tensor | LatentState]]
+    batch: BatchSamples | None
```
grassesi: Please Remove

packages/common/src/weathergen/common/io.py
```diff
@@ -483,20 +483,33 @@ def _create_dataset(self, group: zarr.Group, name: str, array: NDArray):
...
     @functools.cached_property
     def example_key(self) -> ItemKey:
         try:
             sample, example_sample = next(self.data_root.groups())
             stream, example_stream = next(example_sample.groups())
-            fstep = 0
+            forecast_steps = list(example_stream.group_keys())
+            if len(forecast_steps) == 0:
+                raise StopIteration
+            fstep = int(forecast_steps[0])
```
grassesi: `fstep = int(next(example_stream.group_keys()))` should achieve the same thing but is more consistent with the rest of the code.

packages/common/src/weathergen/common/io.py
```diff
@@ -483,20 +483,33 @@ def _create_dataset(self, group: zarr.Group, name: str, array: NDArray):
...
+            sample, example_sample = next(self.data_root.groups())
+            stream, example_stream = next(example_sample.groups())
+        except StopIteration as e:
+            msg = f"Data store at: {self._store_path} is empty."
+            raise FileNotFoundError(msg) from e
+
+        fstep0_key = ItemKey(int(sample), 0, stream)
+        if self.data_root.get(fstep0_key.path) is None:
+            return 1
+        fstep0_datasets = self._get_datasets(fstep0_key)
```
grassesi: This code works against the semantics of the current output, please remove:
- fstep 0 should **always** exist, it contains the sources
- if fstep 0 also contains target/prediction data we know forecast_offset was 0 => else 0 (this is what _infer_forecast_offset does)
- if fstep 0 is missing, this does not mean forecast offset is 1, but that something went wrong.

packages/common/src/weathergen/common/io.py
grassesi: The modification of `OutputBatchData` looks fine, the modification of `ZarrIO` is not. @evenmn Please verify the code by running once with forecast_offset=0 (masked token training) and forecast_offset=1.
The output should be structured like this:
For forecast_offset=0
- sample
  - stream
    - 0
      - prediction
      - target
      - source
    - 1
      - prediction
      - target
    - ...

For forecast_offset=1
- sample
  - stream
    - 0
      - source
    - 1
      - prediction
      - target
    - ...

grassesi (changes requested):
There is some cleanup needed. I provided some suggestions below. Please also carefully check the output again if it follows the semantics I explained.

src/weathergen/model/model.py
evenmn: The idea here was to let the model forward function take either a raw batch or its own output as input. Just one of them is provided at a time, so I don't think the solution proposed here will work. I see the benefits of not mixing batch and pred, but here I think it's the neatest solution :)

src/weathergen/model/model.py
evenmn: @MatKbauer  I agree that would be a more natural name here. However, since I just used the `BatchSamples` class already implemented in src/datasets/batch.py, I think it's better to open a separate PR for this renaming

matisaar:
Need quality training data? My financial datasets are used by quant traders & ML engineers.

**$5** (limited): https://tinyurl.com/2axt4k9b
More info: https://matisaar.github.io/dataforge-sales/

src/weathergen/train/trainer.py
```diff
@@ -160,6 +161,11 @@ def init(self, cf: Config, devices):
         # Initialize collapse monitor for SSL training
         collapse_config = cf.train_logging.get("collapse_monitoring", {})
         self.collapse_monitor = CollapseMonitor(collapse_config, None)  # device set later in run()
+        # self.fcst_chunks_by_mode = {
+        #    "training": self._init_fcst_chunks(self.training_cfg),
+        #    "validation": self._init_fcst_chunks(self.validation_cfg),
+        #    "test": self._init_fcst_chunks(self.test_cfg),
+        # }
```
MatKbauer: Please remove

MatKbauer (commented):
That's a huge PR. I will reach out to arange a meeting such that we can discuss the changes.

When testing the PR with the inference command below, I'm still running the forecast in a single chunk, without sub-chunking being applied.

```
uv run inference --from-run-id zc6sx9jr --options training_config.forecast.fstep_chunk_size=4 training_config.forecast.num_steps=40 test_config.samples_per_mini_epoch=1 test_config.output.num_samples=1
```

Is `training_config.forecast.fstep_chunk_size` the correct parameter?  I could not find it throughout the code elsewhere. The same occurs with `training_config.forecast.chunk_size`.

packages/common/src/weathergen/common/config.py
```diff
@@ -731,6 +731,13 @@ def validate_forecast_policy_and_steps(forecast_cfg: OmegaConf, mode: str):
     if len(forecast_cfg) == 0:
         return
 
+    forecast_cfg.num_steps = forecast_cfg.get("num_steps", 1)
```
clessig: Remove. We don't use the validation check and it is has led to as many problems as it solves.

packages/common/src/weathergen/common/io.py
clessig: How this been implemented?

src/weathergen/model/attention.py
```diff
@@ -102,9 +102,7 @@ def forward(self, x, x_lens, ada_ln_aux=None, coords=None):
         ks = self.lnorm_k(self.proj_heads_k(x).reshape(s)).to(self.dtype)
         vs = self.proj_heads_v(x).reshape(s)
 
-        qs, ks = apply_rope(
-            qs, ks, coords, self.rope_mode, 1
-        )
+        qs, ks = apply_rope(qs, ks, coords, self.rope_mode, 1)
```
clessig: Revert

src/weathergen/model/engines.py
```diff
@@ -80,56 +80,92 @@ def __init__(self, cf: Config, sources_size) -> None:
...
+        chunk_size = self.cf.get("embed_batch_chunk_size", None)
+        if chunk_size is not None:
+            return chunk_size if chunk_size > 0 else None
 
-        num_tokens = torch.sum(batch.tokens_lens, 2).flatten().sum().item()
-        tokens_all = torch.empty(
-            (num_tokens, self.cf.ae_local_dim_embed), dtype=self.dtype, device=batch.get_device()
-        )
+        # Validation / inference favors lower peak memory over throughput.
+        return 1024 if not self.training else None
```
clessig: There should be no magic constants hard-coded in the code.

src/weathergen/model/engines.py
```diff
@@ -80,56 +80,92 @@ def __init__(self, cf: Config, sources_size) -> None:
             else:
                 raise ValueError("Unsupported embedding network type")
 
-    def forward(self, batch, pe_embed):
-        num_steps_input = batch.get_num_steps()
+    def _get_embed_batch_chunk_size(self) -> int | None:
```
clessig: Why is there any chunking here?

clessig (commented):
Didn't finish but here some comments.

PR #2085, opened 2026-03-19
Ensure _latest.chkpt and _latest.json exist and are up to date

### LINKED ISSUE #2084: Checkpoint _latest.chkpt and _latest.json missing or not latest

The `_latest.chkpt` and `_latest.json` are sometimes missing (e.g. when `samples_per_mini_epoch < train_logging.checkpoint`) or not always corresponding to the latest checkpoint (e.g. when `samples_per_mini_epoch` not a multiple of `train_logging.checkpoint`). 

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation

Jubeku:
## Description

Always update _latest entries after saving checkpoint after completing epoch.

## Issue Number

Closes #2084 

### DIFF
--- a/src/weathergen/train/trainer.py
+++ b/src/weathergen/train/trainer.py
@@ -10,6 +10,7 @@
 # nor does it submit to any jurisdiction.
 import copy
 import logging
+import shutil
 import time
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


TillHae (approved):

src/weathergen/train/trainer.py
```diff
@@ -704,6 +705,19 @@ def save_model(self, mini_epoch: int, name=None):
             # save config
             config.save(self.cf, mini_epoch)
 
+            # keep _latest always pointing to the most recent checkpoint
```
clessig, line 708: The solution seems overly complicated for me. What we mean by `latest` is that this are the inter-epoch checkpoints which are saved during the epochs. In most cases, the difference between this and the epoch checkpoint should be scientifically negligible.

One situation that was identified where `latest` is missing is when the epoch is smaller than `train_logging.checkpoint_freq`. Are there other cases?

Jubeku: If `latest` is only for inter-epoch checkpoints, then `train_continue` shouldn’t be dependent on it and rather use completed mini-epoch checkpoints. 
Otherwise I agree the current solution makes the code only more complicated..

clessig: train_continue is not dependent on it, `latest` is just the default since this is what most users currently want usually. There's the `--mini-epoch` argument that can be used to select/specify a specific epoch

Jubeku: When you chain-runs you are dependent on `latest`, aren’t you?

Jubeku: Can we simply do a `self.save_model(-1)` outside the training loop in line 541?
Maybe under following condition to not do the same operation twice in a row:
```
if bidx % self.train_logging.checkpoint != 0: 
    self.save_model(-1)
```

Issue #1925, opened 2026-02-25
Chain inference and evaluation to training


tjhunter:
### Describe the task. Describe the task. It can be a feature, a set of experiments, documentation, etc.

Precious time is lost between the end of a training job and the start of inference and then again evaluation, because each of these require manual steps. As a result, a full cycle that could be done in 24 hours now takes usually 2-3 business days. 

The goal of this task: modify launch-slurm so that it can schedule training, inference and eval

General guidelines:
- it should not change the existing workflows (the current commands should still work, it is just adding new options)
- no resilience to failures
- it should not attempt to unify the configuration between the various steps. It should not require changes to the existing config system in terms of layout or behavior of options
- it should rely on slurm only (no ecflow, airflow, homegrown scheduler etc.)

The goal is to show at least a proof of concept on santis and juwels. Once this is debugged, we can add the other HPCs.

It should likely required a different slurm script for evaluation because the compute needs are very different.

### Area

- [ ] datasets, data readers, data preparation and transfer
- [ ] model
- [ ] science
- [x] infrastructure and engineering
- [x] evaluation, export and visualization
- [ ] documentation

iluise:
This is a first step in this direction: https://github.com/ecmwf/WeatherGenerator/pull/1747

Now the eval parameters are hardcoded in the script, but we probably need to add the eval config parameters to the "main" config so people can choose what to run in advance. 

NB. The comparison _across_ run_ids will still be separate, but if the scores are already computed and stored in json that will take no time and the evaluation will be definitely much faster.

clessig:
If we have a fine-tuning stage (which we currently have for all major lines of work), then this requires an overwrite config. Specifying this as well in the base-config for the pre-training seem complex (although not impossible if we agree that we have zero-level keys `pretraining`, `finetuning`, `evalution`; what when we have multiple fine-tuning stages?)

tjhunter:
Do people have strong feelings if we start with [`training`|`pretraining`|`finetuning`] -> `inference` -> `evaluation`? 

To @clessig  's point:
- we do not have yet a well-defined full cycle in the config for training/pretraining/finetuning
- can someone advocate the case of automatically chaining `training` -> `finetuning` _right now_? I have heard clear pain points for inference and evaluation, but not so much for before yet.

clessig:
> * d full cycle in the config for training/pretraining/finetuning

[...]

PR #2088, opened 2026-03-20
fix: default inference args overwrite start_date and end_date

### LINKED ISSUE #2087: Test config periods are overwritten by default values unless directly handed over through CLI
### What happened?

Parameters `test_config.start_date` and `test_config.end_date` in the config are not used during inference and instead overwritten by default values specified in cli.py here: https://github.com/ecmwf/WeatherGenerator/blob/develop/src/weathergen/utils/cli.py#L76-L107 
The latter ones can be overwritten during submission from CLI, but if no CLI options are specified, the test config options should be used _**when explicitly specified**_. The last point is important, because if they are not explicitly specified, the test config will be a copy of the training config and hence falling back to the default values in cli.py is preferred. 

Specify `test_config.start_date` and `test_config.end_date` different from the default values in cli.py and run inference which will fall back to the default values.

_No response_

simone99n:
## Description

- part of #1925

## Issue Number

- closes #2087

### DIFF
--- a/src/weathergen/run_train.py
+++ b/src/weathergen/run_train.py
@@ -98,8 +98,8 @@ def run_inference(args):
         args.mini_epoch,
         args.base_config,
         *args.config,
-        inference_overwrite,
         cli_overwrite,
+        inference_overwrite,
     )
     cf = config.set_run_id(cf, args.run_id, args.reuse_run_id)
 
--- a/src/weathergen/run_train.py
+++ b/src/weathergen/run_train.py
@@ -103,7 +103,6 @@ def run_inference(args):
     )
     cf = config.set_run_id(cf, args.run_id, args.reuse_run_id)
 
-    logger.info(f"Running inference with config: {cf}")
     devices = Trainer.init_torch()
     cf = Trainer.init_ddp(cf)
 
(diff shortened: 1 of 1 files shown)


MatKbauer (commented):
Thanks for tackling, @simone99n! I have two questions:
1. Don't we also have to remove the default dates in [cli.py](https://github.com/ecmwf/WeatherGenerator/blob/develop/src/weathergen/utils/cli.py#L84C1-L84C30) to prevent them from overwriting the dates?
2. What happened to `samples_per_mini_epoch=args.samples` and `output=dict(num_samples=args.samples if args.save_samples else 0)` in your updated `run_train.py`?

clessig:
> Thanks for tackling, @simone99n! I have two questions:
> 
> 1. Don't we also have to remove the default dates in [cli.py](https://github.com/ecmwf/WeatherGenerator/blob/develop/src/weathergen/utils/cli.py#L84C1-L84C30) to prevent them from overwriting the dates?
> 2. What happened to `samples_per_mini_epoch=args.samples` and `output=dict(num_samples=args.samples if args.save_samples else 0)` in your updated `run_train.py`?

@simone99n @MatKbauer : Fixed by #2098 . Agreed?

PR #2123, opened 2026-03-27
Rolfhm/dev/mutable config

### LINKED ISSUE #1528: config should be immutable, dynamic parameters should be tracked in run_params

The config is currently changed on the fly, both at run initialization time (e.g. paths are resolved, stream configs are resolved with channels) and during a run (e.g. istep). Parameters that change during a run should be kept in a new structure outside the config, potentially also all things resolved at initalization time. 

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation

rolfhm:
## Description

Beginning of a mutable config class for values that change during a run.

## Issue Number

Closes #1528 

-   [X] I have informed and aligned with people impacted by my change:


src/weathergen/model/model_interface.py
```diff
@@ -46,16 +46,16 @@
...
     run_id_contd,
     mini_epoch_contd,
     training_mode,
     device,
-    with_ddp,
     with_fsdp,
     overrides={},
 ):
-    model_creation_device = "meta" if with_ddp and with_fsdp else "cuda"
+    model_creation_device = "meta" if mcf.with_ddp and with_fsdp else "cuda"
```
clessig: This is inconsistent since with_ddp and with_fsdp are specified in the config. I think we should remove with_ddp in the config--DDP will always be used when possible and this is standard and correct behaviour. 

We need to remove it in all configs then.

src/weathergen/model/model_interface.py
```diff
@@ -320,10 +326,13 @@ def get_target_aux_calculator(
             meta_ema_model,
             halflife_steps=target_and_aux_calc_params.get("ema_halflife_in_thousands", 1e-3),
             rampup_ratio=target_and_aux_calc_params.get("ema_ramp_up_ratio", 0.09),
-            is_model_sharded=(cf.with_ddp and cf.with_fsdp),
+            is_model_sharded=(with_ddp and cf.with_fsdp),
```
clessig: We re-determine this multiple times. I think we should introduce a flag in the `mcf`--this is more robust.

packages/common/src/weathergen/common/mutable_config.py
```diff
@@ -0,0 +1,31 @@
...
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+from omegaconf import DictConfig
+
+import logging
+
+_logger = logging.getLogger(__name__)
+
+class MutableConfig():
```
clessig: I don't think this is the right name. It's not mutable during the run (except for istep) but it's determined at run initialization time. If I start from `istep` then this is more the run state.

packages/common/src/weathergen/common/run_state.py
```diff
@@ -0,0 +1,31 @@
...
+
+from omegaconf import DictConfig
+
+import logging
+
+_logger = logging.getLogger(__name__)
+
+class MutableConfig():
+    """
+    lalala
```
clessig: Update

packages/common/src/weathergen/common/run_state.py
```diff
@@ -0,0 +1,31 @@
...
+
+        self.istep = None
+        self.world_size = None
+        self.world_size_original = None
+        self.rank = None
+        self.local_rank = None
+        self.with_ddp = None
+        self.run_history = []
+
+        print("Trolololo 1")
```
clessig: Remove

packages/common/src/weathergen/common/run_state.py
```diff
@@ -0,0 +1,31 @@
...
+import logging
+
+_logger = logging.getLogger(__name__)
+
+class MutableConfig():
+    """
+    lalala
+    """
+
+    def __init__(self):
```
clessig: Shouldn't this class have serialization and deserialization code? Also: a lot of code around the regular config is to ensure it is loaded and stored correctly. This has not been touched?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -219,7 +221,7 @@ def __init__(
             self.len = self.samples_per_mini_epoch
 
         # adjust len to split loading across all workers and ensure it is multiple of batch_size
-        len_chunk = ((self.len // cf.world_size) // self.batch_size) * self.batch_size
+        len_chunk = ((self.len // mcf.world_size) // self.batch_size) * self.batch_size
```
clessig: self.world_size

src/weathergen/train/trainer_base.py
```diff
@@ -136,9 +138,9 @@ def init_ddp(cf):
...
-        cf.rank = rank
-        cf.local_rank = local_rank
-        cf.with_ddp = world_size > 1
+        mcf.world_size = world_size
+        mcf.rank = rank
+        mcf.local_rank = local_rank
+        mcf.with_ddp = world_size > 1
 
-        return cf
+        return cf, mcf
```
clessig: We don't need to return cf

clessig (commented):
Thanks for the work. We probably need some discussion of how to write it and to make sure we don't need to modify too much code related to reading/writing config.

src/weathergen/train/trainer_base.py
rolfhm: general.run_id and data_loader_rng_seed are also computed in this routine. Should they also be moved to runstate?

packages/common/src/weathergen/common/run_state.py
rolfhm: cf is just a DictConfig from OmegaConf as far as I can tell. Is there any built in code in DictConfig to handle it?
Wondering if it would be easier to just make RunState (former MutableConfig) also a DictConfig? @tjhunter

src/weathergen/train/trainer_base.py
clessig: There is the decision if we consider run_id as a mutable object or not. It's (usually) determined at init time but then always constant. I hence would prefer if it's part of the immutable config.

packages/common/src/weathergen/common/run_state.py
clessig: OmegaConf.DictConfig makes a lot of sense to me: it has some useful functionality (like merge()) that we want to have.

src/weathergen/model/model_interface.py
rolfhm: runstate now has an `is_sharded` variable

src/weathergen/model/model_interface.py
rolfhm: `with_ddp` is determined in `init_ddp` and is set to `world_size > 1`

packages/common/src/weathergen/common/mutable_config.py
rolfhm: changed to `RunState`

packages/common/src/weathergen/common/run_state.py
rolfhm: done

packages/common/src/weathergen/common/run_state.py
rolfhm: done

packages/common/src/weathergen/common/run_state.py
rolfhm: Chenged it to an OmegaConf DictConf object

src/weathergen/datasets/multi_stream_data_sampler.py
rolfhm: fixed

src/weathergen/datasets/multi_stream_data_sampler.py
grassesi, line 1: No need to pass runstate to MultiStreamDataSampler: It is only needed for rank/world_size, both can be looked up globally using `weathergen.utils.distributed.get_rank()` / `.get_world_size()`

packages/common/src/weathergen/common/config.py
grassesi, line 1: This function should stay here, since all other functions that look up paths are also defined in config.py

src/weathergen/train/trainer_base.py
```diff
@@ -136,9 +138,10 @@ def init_ddp(cf):
...
-        cf.local_rank = local_rank
-        cf.with_ddp = world_size > 1
+        runstate.world_size = world_size
+        runstate.rank = rank
+        runstate.local_rank = local_rank
+        runstate.with_ddp = world_size > 1
+        runstate.is_sharded = runstate.with_ddp and cf.with_fsdp
 
-        return cf
+        return cf, runstate
```
grassesi: If this always returns a freshly instantiated RunState and loading the a old runstate only overwrites the needed attributes (`istep`, `run_history`, `world_size_original`), the initialization program flow is simplified.

src/weathergen/utils/train_logger.py
grassesi, line 1: This will break things, please revert.

src/weathergen/run_train.py
```diff
@@ -97,22 +98,23 @@ def run_inference(args):
...
     trainer = Trainer(cf.train_logging)
     try:
-        trainer.inference(cf, devices, args.from_run_id, args.mini_epoch)
+        trainer.inference(cf, runstate, devices, args.from_run_id, args.mini_epoch)
     except Exception:
         extype, value, tb = sys.exc_info()
         traceback.print_exc()
-        if cf.world_size == 1:
+        if runstate.world_size == 1:
             pdb.post_mortem(tb)
```
grassesi, line 115: Program flow should be as follows (see comments on `TrainerBase.init_ddp`, `run_state.load_runstate`):
1. `TrainerBase.init_ddp` returns fresh runstate (with `world_size`, etc. set)
2. loading overwrites `istep`, `run_history`, `world_size_original`

src/weathergen/run_train.py
```diff
@@ -135,23 +137,24 @@ def run_continue(args):
     )
     cf = config.set_run_id(cf, args.run_id, args.reuse_run_id)
 
+    runstate = load_runstate(run_id=args.from_run_id, mini_epoch=args.mini_epoch, model_path=None)
     mp_method = cf.general.get("multiprocessing_method", "fork")
     devices = Trainer.init_torch(multiprocessing_method=mp_method)
-    cf = Trainer.init_ddp(cf)
+    cf, runstate = Trainer.init_ddp(cf, runstate)
```
grassesi: see above.

src/weathergen/run_train.py
```diff
@@ -169,16 +172,19 @@ def run_train(args):
...
     cf = config.set_run_id(cf, args.run_id, False)
 
+    runstate = init_runstate()
+
     cf.data_loading.rng_seed = int(time.time())
     mp_method = cf.general.get("multiprocessing_method", "fork")
     devices = Trainer.init_torch(multiprocessing_method=mp_method)
-    cf = Trainer.init_ddp(cf)
+
+    cf, runstate = Trainer.init_ddp(cf, runstate)
```
grassesi, line 172: see above.

packages/common/src/weathergen/common/config.py
grassesi, line 1: A compatibility mechanism is needed, so that training/inference can be continued from configs that dont implement runstate yet.

grassesi, line 1: `_get_model_config_file_write_name`, `_get_model_config_file_read_name` as well as `load_run_config` can be simplified, since the file name for the config should now be unambiguous: `f"model_{run_id}.json"` (It is now independent from mini epoch.)

packages/common/src/weathergen/common/run_state.py
grassesi, line 1: This code is very much inspired by `weathergen.common.config`, but it does not need to be. Since we are introducing something new here, the opportunity should be taken to do things properly:
- A dataclass for the RunState => we dont need to be super flexible like we need to be for the general config
- The least amount of complexity needed in terms of serialization logic => only -1 (`"_latest"`) or n (`f"_chkpt{n:05d}"`) as valid suffixes. No loading from full path or fallback logic for loading (like it used to be in `config.load_run_config`)

grassesi (changes requested):
Thanks for the implementation. There are some fixes required and this PR would also benefit from some structural improvements. I will link a PR with suggestions later today.

grassesi:
Suggestion for improvements can be found [here](https://github.com/rolfhm/WeatherGenerator/pull/3). Please have a look to see if anything is missing, feel free to merge it if you want to.

PR #2140, opened 2026-03-30
Parallel I/O, scoring and plotting for FastEvaluation 

### LINKED ISSUE #2095: speed up export package

The export to grib is extremely slow (95h for 360 samples and 40 forecast steps). 
With the help of Claude we need to get it back to around 2-3h.  

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation

iluise:
## Description

End-to-end refactor of the evaluation pipeline for speed, correctness, and maintainability.
The primary goal was to make evaluation runs significantly faster on HPC (Jupiter) by
parallelising I/O, scoring, and plotting at multiple levels.

---

### Parallel raw zarr I/O (`wegen_reader.py`)

- Bypasses `ZarrIO` / `dask` / `as_xarray` entirely with direct zarr array access via
  `_read_sample_raw`
- One loky worker per sample; each worker opens its own store handle (safe for both
  `ZipStore` and `LocalStore`)
- `ZipStore` path has its own parallel pool with graceful sequential fallback on failure
- Sub-step-aware: fstep keys are expanded when `tokenize_spacetime` is active
- `_resolve_num_io_workers`: auto-detects safe parallelism from `ulimit -u` and current
  process count; formula `min(headroom // 4, cpu_count, 16)`

### Parallel scoring (`utils.py`)

- `_score_single_fstep`: stateless, thread-safe worker — xarray/numpy release the GIL
  during C-level computation
- All fsteps scored concurrently with `ThreadPoolExecutor(num_scoring_threads)`
- Results reassembled in deterministic fstep order after all futures complete

### Parallel score-map plotting (`utils.py`)

- All `(region, fstep)` pairs dispatched in parallel via `Parallel(loky, n_jobs=n_plot_workers)`
- `_score_map_fstep_worker`: picklable module-level function — no `Reader` object sent to
  workers (only plain dicts and DataArrays)
- Inner metric computation parallelised with `ThreadPoolExecutor` (threads safe inside loky
  workers)
- `_resolve_num_plot_workers` called **once** per stream, not once per fstep

### Parallel map plotting (`utils.py`)

- `_plot_single_sample`: module-level loky worker, one task per `(fstep, sample)`
- Each worker creates its own `Plotter` / matplotlib state — no shared state

### Parallel animations (`plotter.py`)

- `_build_single_animation`: module-level function dispatched via `ThreadPoolExecutor`
- One task per `(region, sample, variable)` — pure I/O + Pillow, no matplotlib state

---

### Data loaded once, shared everywhere (`run_evaluation.py`)

- `reader.get_data()` / `reader.get_data_raw()` called once per stream in `_process_stream`
- The resulting `output_data` object is passed to `plot_data`, `calc_scores_per_stream`,
  and `plot_score_maps_per_stream` — eliminates the previous double (or triple) zarr read

### Score maps separated from scoring (`utils.py`)

- `calc_scores_per_stream`: pure scoring, no plotting responsibility
- `plot_score_maps_per_stream`: new public API for spatial score maps (aggregates over the
  `sample` dimension, keeps `ipoint` for 2D display)
- `run_evaluation.py` calls them sequentially with the shared `output_data`

### Early channel selection (`wegen_reader.py`)

- Channel index filtering applied before reading zarr arrays, not after
- Only the requested channels are read from disk

### Substep-aware plotting filter (`utils.py`)

- Detects sub-step key expansion (zarr fsteps × substeps → output dict keys)
- Expands the plotting fstep filter to include all output keys when `"all"` was requested,
  so sub-steps are not silently dropped

## Issue Number
Closes https://github.com/ecmwf/WeatherGenerator/issues/2095


iluise:
Checked that the jsons are the same between this branch and the current `develop`: 

Run ID           Stream             Region  Metric Status           Detail
========================================================================================================================
sbi5gmcj         ERA5               global  fbi    ✅ OK             max_rel=0.00e+00, max_abs=0.00e+00
sbi5gmcj         ERA5               global  froct  ✅ OK             max_rel=7.02e-05, max_abs=2.35e-04
sbi5gmcj         ERA5               global  mae    ✅ OK             max_rel=6.72e-06, max_abs=1.56e-04
sbi5gmcj         ERA5               global  rmse   ✅ OK             max_rel=5.64e-06, max_abs=2.75e-04
sbi5gmcj         ERA5               global  troct  ✅ OK             max_rel=2.60e-05, max_abs=7.44e-05
sbi5gmcj         ERA5               nhem    fbi    ✅ OK             max_rel=0.00e+00, max_abs=0.00e+00
sbi5gmcj         ERA5               nhem    froct  ✅ OK             max_rel=2.81e-05, max_abs=1.92e-04
sbi5gmcj         ERA5               nhem    mae    ✅ OK             max_rel=5.11e-06, max_abs=2.82e-04
sbi5gmcj         ERA5               nhem    rmse   ✅ OK             max_rel=4.21e-06, max_abs=3.28e-04
sbi5gmcj         ERA5               nhem    troct  ✅ OK             max_rel=3.73e-05, max_abs=4.86e-05
sbi5gmcj         NPPATMS            global  fbi    ✅ OK             max_rel=0.00e+00, max_abs=0.00e+00
sbi5gmcj         NPPATMS            global  froct  ✅ BOTH-NaN       all values NaN on both branches
sbi5gmcj         NPPATMS            global  mae    ✅ OK             max_rel=3.82e-06, max_abs=8.01e-05
sbi5gmcj         NPPATMS            global  rmse   ✅ OK             max_rel=1.76e-06, max_abs=5.34e-05
sbi5gmcj         NPPATMS            global  troct  ✅ BOTH-NaN       all values NaN on both branches
sbi5gmcj         NPPATMS            nhem    fbi    ✅ OK             max_rel=0.00e+00, max_abs=0.00e+00
sbi5gmcj         NPPATMS            nhem    froct  ✅ BOTH-NaN       all values NaN on both branches
sbi5gmcj         NPPATMS            nhem    mae    ✅ OK             max_rel=3.46e-06, max_abs=1.01e-04
sbi5gmcj         NPPATMS            nhem    rmse   ✅ OK             max_rel=1.32e-06, max_abs=4.20e-05
sbi5gmcj         NPPATMS            nhem    troct  ✅ BOTH-NaN       all values NaN on both branches
sbi5gmcj         SurfaceCombined    global  fbi    ✅ OK             max_rel=0.00e+00, max_abs=0.00e+00
sbi5gmcj         SurfaceCombined    global  froct  ✅ BOTH-NaN       all values NaN on both branches
sbi5gmcj         SurfaceCombined    global  mae    ✅ OK             max_rel=2.67e-06, max_abs=1.00e-05
sbi5gmcj         SurfaceCombined    global  rmse   ✅ OK             max_rel=2.26e-06, max_abs=1.38e-05
sbi5gmcj         SurfaceCombined    global  troct  ✅ BOTH-NaN       all values NaN on both branches
sbi5gmcj         SurfaceCombined    nhem    fbi    ✅ OK             max_rel=0.00e+00, max_abs=0.00e+00
sbi5gmcj         SurfaceCombined    nhem    froct  ✅ BOTH-NaN       all values NaN on both branches
sbi5gmcj         SurfaceCombined    nhem    mae    ✅ OK             max_rel=3.97e-06, max_abs=1.24e-05
sbi5gmcj         SurfaceCombined    nhem    rmse   ✅ OK             max_rel=1.82e-06, max_abs=1.19e-05
sbi5gmcj         SurfaceCombined    nhem    troct  ✅ BOTH-NaN       all values NaN on both branches

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -26,9 +27,23 @@
...
+    get_data_raw_impl,
+    get_data_raw_zip_impl,
+    resolve_num_io_workers,
+)
 from weathergen.evaluate.io.io_reader import Reader, ReaderOutput
 from weathergen.evaluate.scores.score_utils import to_list
-from weathergen.evaluate.utils.derived_channels import DeriveChannels
+
+_logger = logging.getLogger(__name__)
+_logger.setLevel(logging.INFO)
```
florianscheidl, line 39: Duplicate

florianscheidl (commented):
Finding it hard to review due to the volume of changes. I suggest you build a few unit tests on simple cases we understand to validate the newly added functions. 

Also, if there are any particular sections you would like to have reviewed, let me know :)

packages/evaluate/src/weathergen/evaluate/io/data/__init__.py
```diff
@@ -0,0 +1,8 @@
+# (C) Copyright 2025 WeatherGenerator contributors.
```
clessig: Leave empty

packages/evaluate/src/weathergen/evaluate/io/data/dataarray_builders.py
```diff
@@ -0,0 +1,301 @@
...
+    tars_list: list[NDArray],
+    preds_list: list[NDArray],
+    samples: list[int],
+    read_channels: list[str],
+    lat: NDArray,
+    lon: NDArray,
+    per_sample_valid_times: list[np.datetime64],
+    source_interval_starts: NDArray,
+    forecast_step_val: int,
+    ensemble: list[str],
```
clessig: Use `ensemble` and `all_ens` to resolve the indices so that one does not have to pass both variables around.

packages/evaluate/src/weathergen/evaluate/io/data/dataarray_builders.py
```diff
@@ -0,0 +1,301 @@
...
+    preds_stacked : np.ndarray
+        Shape (n_samples, n_ipoints, n_channels[, n_ens]).
+    base_coords : dict
+        Coordinate dict (without ens).
+    ensemble : list[str]
+        Requested ensemble members or ["mean"].
+    all_ens : list[str] | None
+        All ensemble member names from zarr (needed for index mapping).
+    """
+    if preds_stacked.ndim == 4:
```
clessig: These branches have a lot of overlap. By collecting things probably one can reduce the code by half and make it also more readable; one should have one return statement at the end of the function and the code before should collect the args to xr.DataArray()

packages/evaluate/src/weathergen/evaluate/io/data/dataarray_builders.py
```diff
@@ -0,0 +1,301 @@
...
+        "ipoint": np.arange(n_ipoints),
+        "channel": read_channels,
+        "lat": ("ipoint", sub_lat),
+        "lon": ("ipoint", sub_lon),
+        "valid_time": (("sample", "ipoint"), valid_time_2d),
+        "source_interval_start": ("sample", source_interval_starts.copy()),
+        "forecast_step": forecast_step_val,
+    }
+
+    da_tar = xr.DataArray(
```
clessig: This motiv also appears in build_pred_datarray() and could be refactored--the target case is identical to no ensemble (up to squeeze, which however is the identity() if nothing is to be squeezed)

packages/evaluate/src/weathergen/evaluate/io/data/dataarray_builders.py
```diff
@@ -0,0 +1,301 @@
...
+
+    for si, sample in enumerate(samples):
+        n_ip = tars_list[si].shape[0]
+        tar_data = tars_list[si]  # (n_ip, n_channels)
+        pred_data = preds_list[si]  # (n_ip, n_channels[, n_ens])
+
+        # Use per-sample coords if available, otherwise fall back to reference
+        sc = per_sample_coords[si] if si < len(per_sample_coords) else None
+        if sc is not None and len(sc) >= n_ip:
+            sample_lat = sc[:n_ip, 0].astype(np.float64)
```
clessig: Why do we need fp64 here?

packages/evaluate/src/weathergen/evaluate/io/data/dataarray_builders.py
```diff
@@ -0,0 +1,301 @@
...
+            if ensemble == ["mean"]:
+                pred_data = pred_data.mean(axis=-1)
+                pred_coords = dict(sample_coords)
+                da_p = xr.DataArray(
+                    pred_data,
+                    dims=["ipoint", "channel"],
+                    coords=pred_coords,
+                )
+            else:
+                ens_idxs = (
```
clessig: This is the same code as in `build_pred_dataarray()` below for the ens case.

packages/evaluate/src/weathergen/evaluate/io/data/dataarray_builders.py
```diff
@@ -0,0 +1,301 @@
...
+
+        da_t = xr.DataArray(
+            tar_data,
+            dims=["ipoint", "channel"],
+            coords=sample_coords,
+        )
+        per_sample_tars.append(da_t)
+
+        # Handle ensemble for predictions
+        if pred_data.ndim == 3:
```
clessig: Same comment as below in `build_pred_dataarray()` too much code duplication in the three branches.

clessig: Also, is this code really different from `build_pred_dataarray()` and couldn't we re-use the function?

packages/evaluate/src/weathergen/evaluate/io/data/get_data_orchestration.py
```diff
@@ -0,0 +1,216 @@
...
+    _logger.info(
+        f"RUN {run_id} - {stream}: Loading {len(samples)} samples × "
+        f"{len(fsteps)} fsteps = {n_total} items "
+        f"(workers={effective_threads}, zip={is_zip_store})..."
+    )
+
+    results_by_fstep: dict[int, list] = {f: [] for f in fsteps}
+
+    if effective_threads > 1:
+        if is_zip_store:
```
clessig: Why is a different parallelization mode used depending on the zarr store type?

packages/evaluate/src/weathergen/evaluate/io/data/get_data_orchestration.py
```diff
@@ -0,0 +1,216 @@
...
+                    if result is not None:
+                        fstep_r, sample_r, target, pred, vt = result
+                        results_by_fstep[fstep_r].append((sample_r, target, pred, vt))
+        except (RuntimeError, OSError) as pool_err:
+            _logger.warning(
+                f"Parallel pool failed ({pool_err}). Falling back to sequential loading."
+            )
+            effective_threads = 0  # signal caller to run sequential
+
+    if effective_threads <= 1:
```
clessig: This variable should be parallel_failed or similar. In terms of code structure, it would also be more logical if we have two functions: read_parallel and read_serial and read_serial is also used as a fallback.

packages/evaluate/src/weathergen/evaluate/io/data/get_data_orchestration.py
```diff
@@ -0,0 +1,216 @@
...
+                zio,
+                s,
+                stream,
+                f,
+                ensemble,
+                is_gridded,
+            )
+
+        try:
+            with pool_cls(**pool_kwargs) as executor:
```
clessig: This looks much more complex than the code I have based on `joblib`.

packages/evaluate/src/weathergen/evaluate/io/data/get_data_orchestration.py
```diff
@@ -0,0 +1,216 @@
...
+                result = _load_single_sample(zio, s, stream, f, ensemble, is_gridded)
+                if result is None:
+                    continue
+                _, sample_r, target, pred, vt = result
+                results_by_fstep[f].append((sample_r, target, pred, vt))
+
+    return results_by_fstep, effective_threads
+
+
+def _reassemble_fsteps(
```
clessig: Does the parallel dispatch write the result in random order?

packages/evaluate/src/weathergen/evaluate/io/data/get_data_orchestration.py
```diff
@@ -0,0 +1,216 @@
...
+def _apply_postprocessing(
+    fsteps_final: list,
+    da_tars: list,
+    da_preds: list,
+    stream: str,
+    channels: list[str],
+    stream_cfg: dict,
+    is_gridded: bool,
+) -> ReaderOutput:
+    """Phase 3: channel selection, scaling, and lead_time coord assignment."""
```
clessig: Was it tested that it's not faster to only read required channels?

packages/evaluate/src/weathergen/evaluate/io/data/raw_io_orchestration.py
```diff
@@ -0,0 +1,518 @@
...
+    n_workers: int
+    backend: str = "loky"
+
+
+# ---------------------------------------------------------------------------
+# Shared helpers
+# ---------------------------------------------------------------------------
+
+
+def resolve_num_io_workers(requested: int) -> int:
```
clessig: We should primarily use `$SLURM_CPUS_PER_TASK` and/or `$SLURM_CPUS_ON_NODE`.

clessig: I am not sure how much code we should invest into a fallback

packages/evaluate/src/weathergen/evaluate/io/data/raw_io_orchestration.py
```diff
@@ -0,0 +1,518 @@
...
+        try:
+            return Parallel(n_jobs=n_workers, backend=backend, verbose=5)(
+                delayed(_read_sample_raw)(sample=s, **kwargs) for s in samples
+            )
+        except (RuntimeError, OSError) as err:
+            _logger.warning(f"{label}: parallel pool failed ({err}). Falling back to sequential.")
+            with contextlib.suppress(Exception):
+                get_reusable_executor().shutdown(wait=True)
+
+    return [_read_sample_raw(sample=s, **kwargs) for s in samples]
```
clessig: We should have one return statement and run sequentially in the except block if necessary.

packages/evaluate/src/weathergen/evaluate/io/data/raw_io_orchestration.py
```diff
@@ -0,0 +1,518 @@
...
+        fsteps=fsteps_arg,
+        channel_idxs=channel_idxs,
+        is_zip=is_zip,
+        read_coords=need_coords,
+        is_gridded=is_gridded,
+    )
+
+    if n_workers > 1:
+        try:
+            return Parallel(n_jobs=n_workers, backend=backend, verbose=5)(
```
clessig: I am confused here: how does this parallelization differ from those in `_dispatch_reads()`?

clessig (commented):
First partial review.

packages/evaluate/src/weathergen/evaluate/io/data/xarray_utils.py
```diff
@@ -0,0 +1,326 @@
...
+            )
+            channels = [ch for ch in channels if ch in all_channels]
+
+        da_tar = da_tar.sel(channel=channels)
+        da_pred = da_pred.sel(channel=channels)
+
+    return da_tar, da_pred
+
+
+def _scale_z_channels(data: xr.DataArray, stream: str) -> xr.DataArray:
```
clessig: I think we should have a more generic function that applies transformations to channels. E.g. CERRA has wind speed and wind direction and it is hence difficult to compare to ERA5. And for ERA5 having wind speed and wind direction would be more useful than raw u and v values.

packages/evaluate/src/weathergen/evaluate/io/data/xarray_utils.py
```diff
@@ -0,0 +1,326 @@
...
+    factor = 9.80665
+
+    if channels_z:
+        channels = data.channel.astype(str)
+        mask = channels.str.startswith("z_")
+        data = data.where(~mask, data / factor)
+    return data
+
+
+def _split_by_valid_time(arrays: list[xr.DataArray]) -> list[xr.DataArray]:
```
clessig: The data is stored by time (forecast step) in the zarr. Why is this function needed? I also remember that it's quite slow?

packages/evaluate/src/weathergen/evaluate/io/data/xarray_utils.py
```diff
@@ -0,0 +1,326 @@
...
+        raise ValueError(
+            "Multiple lead_time values across samples for "
+            f"forecast_step {da.forecast_step.values}: {unique_lead}"
+        )
+
+    da = da.assign_coords(lead_time=unique_lead[0])
+    return da
+
+
+def _force_consistent_grids(ref: list[xr.DataArray]) -> xr.DataArray:
```
clessig: Do we still need this function? It is enforced on the level of the model now.

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -366,155 +395,101 @@ def get_data(
...
-                    if ensemble == ["mean"]:
-                        _logger.debug("Averaging over ensemble members.")
-                        pred = pred.mean("ens", keepdims=True)
-                    else:
-                        _logger.debug(f"Selecting ensemble members {ensemble}.")
-                        pred = pred.sel(ens=ensemble)
-
-                    pred = pred.squeeze()
-                    target = target.squeeze()
+            effective_threads = min(self._num_io_threads, len(samples) * len(fsteps))
```
clessig: This should be determined in `_dispatch_reads()`

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -366,155 +395,101 @@ def get_data(
...
+            )
 
-                    if is_gridded_data:
-                        vt_list = np.unique(target.valid_time.values).tolist()
-                        if len(vt_list) > 1:
-                            valid_times_fs.append(vt_list)
+        fsteps_final, da_tars, da_preds = _reassemble_fsteps(
+            results_by_fstep, fsteps, is_gridded, self.run_id, stream
+        )
+        return _apply_postprocessing(
```
clessig: Does it logically really make sense to have this in the return statememnt?

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -366,155 +395,101 @@ def get_data(
...
+        ensemble: list[str] | None = None,
+    ) -> ReaderOutput:
+        """Fast-path data loading using direct zarr array access and joblib parallelism.
 
-                if not da_tars_fs:
-                    _logger.info(
-                        f"[{self.run_id} - {stream}] No valid data found for fstep {fstep}."
-                    )
-                    continue
+        Bypasses the ZarrIO → OutputDataset → as_xarray → dask pipeline for
```
clessig: Do we still need the old pipeline?

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -366,155 +395,101 @@ def get_data(
...
+                resolved_ensemble,
+                self._num_io_workers,
+                self.get_ensemble(stream) if resolved_ensemble != ["mean"] else None,
+            )
+            if state.is_zip:
+                return get_data_raw_zip_impl(state)
+            return get_data_raw_impl(state)
+        except Exception as e:
+            _logger.warning(
+                f"Fast I/O failed for {self.run_id} - {stream}: {e}. "
```
clessig: Can this really happen? The code itself has already a serial reader as fallback

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -366,155 +395,101 @@ def get_data(
...
+                f"Falling back to standard get_data()."
+            )
+            with contextlib.suppress(Exception):
+                get_reusable_executor().shutdown(wait=True)
+            return self.get_data(stream, samples, fsteps, channels, ensemble)
 
     ######## reader utils ########
```
clessig: Remove non-standard comment

packages/evaluate/src/weathergen/evaluate/plotting/plotter.py
```diff
@@ -620,47 +626,46 @@ def animation(self, samples, fsteps, variables, select, tag) -> list[str]:
...
+                "duration_ms": duration_ms,
+            }
+            for region in self.regions
+            for sa in samples
+            for var in variables
+        ]
 
-                    else:
-                        _logger.warning(f"No images found for animation {var} sample {sa}")
+        n_workers = min(8, len(tasks))
```
clessig: Reuse the code that is used to determine number of threads for reading the data (using SLURM vars as default option).

packages/evaluate/src/weathergen/evaluate/plotting/plotter.py
```diff
@@ -620,47 +626,46 @@ def animation(self, samples, fsteps, variables, select, tag) -> list[str]:
...
+                futures = {executor.submit(_build_single_animation, **t): t for t in tasks}
+                for future in tqdm(
+                    as_completed(futures),
+                    total=len(futures),
+                    desc=f"Creating animations {self.stream} {tag}",
+                ):
+                    result = future.result()
+                    if result:
+                        all_image_paths.extend(result)
+        else:
```
clessig: Does the above code really fail with n_workers=1 and len(tasks)=1?

packages/evaluate/src/weathergen/evaluate/plotting/plotter.py
```diff
@@ -620,47 +626,46 @@ def animation(self, samples, fsteps, variables, select, tag) -> list[str]:
...
+            for var in variables
+        ]
 
-                    else:
-                        _logger.warning(f"No images found for animation {var} sample {sa}")
+        n_workers = min(8, len(tasks))
+        all_image_paths: list[str] = []
+
+        if n_workers > 1 and len(tasks) > 1:
+            with ThreadPoolExecutor(max_workers=n_workers) as executor:
```
clessig: What's the advantage of ThreadPoolExecutor compared to joblib? Having only one approach for parallelization in the eval code would be much better (ie simpler to read and to maintain).

packages/evaluate/src/weathergen/evaluate/plotting/plotter.py
```diff
@@ -682,6 +687,66 @@ def get_map_title(self, var, valid_time, data):
...
+    images = [Image.open(p) for p in image_paths]
+    out_path = f"{map_output_dir}/animation_{run_id}_{tag}_{sa}_{stream}_{region}_{var}.gif"
+    images[0].save(
+        out_path,
+        save_all=True,
+        append_images=images[1:],
+        duration=duration_ms,
+        loop=0,
+    )
+    # Close frames to free file handles
```
clessig: The main point of plt.close() is not to clear the file handles but the matplotlib internal resources associated with the plot

packages/evaluate/src/weathergen/evaluate/plotting/plotter.py
```diff
@@ -682,6 +687,66 @@ def get_map_title(self, var, valid_time, data):
...
+    for img in images:
+        img.close()
+
+    _logger.debug(f"Saved animation → {out_path}")
```
clessig: "->" -> "to"

packages/evaluate/src/weathergen/evaluate/plotting/plotter.py
```diff
@@ -1487,8 +1552,10 @@ def plot(
         skill_models = []
         for run_index in range(1, n_runs):
             skill_model = 0.0
+            data0_channels = [str(x) for x in np.atleast_1d(data[0].channel.values)]
+            datai_channels = [str(x) for x in np.atleast_1d(data[run_index].channel.values)]
```
clessig: datai_channels -> data_idx_channels

packages/evaluate/src/weathergen/evaluate/plotting/plotter.py
```diff
@@ -1578,8 +1645,10 @@ def plot(
     def extract_common_channels(self, data, channels, n_runs):
         common_channels = []
         for run_index in range(1, n_runs):
+            data0_channels = [str(x) for x in np.atleast_1d(data[0].channel.values)]
+            datai_channels = [str(x) for x in np.atleast_1d(data[run_index].channel.values)]
```
clessig: Same as above

packages/evaluate/src/weathergen/evaluate/utils/plotting.py
```diff
@@ -0,0 +1,681 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Process headroom helper
+# ---------------------------------------------------------------------------
+
+
+def _resolve_num_plot_workers(requested: int = 0) -> int:
```
clessig: Third time we have the same code for the functionality. Again, SLURM env var should be used by default.

packages/evaluate/src/weathergen/evaluate/utils/plotting.py
```diff
@@ -0,0 +1,681 @@
...
+
+    n_tasks = len(fstep_tasks)
+    effective = min(n_plot_workers, n_tasks)
+    _logger.info(
+        f"RUN {run_id} - {stream}: Plotting {n_tasks} score-map tasks "
+        f"({len(regions)} region(s) × {len(fsteps)} fstep(s)) "
+        f"with {effective} worker(s)."
+    )
+
+    if effective > 1 and n_tasks > 1:
```
clessig: Why do we have this code here but also parallel plotting code in plotter.py?

packages/evaluate/src/weathergen/evaluate/utils/plotting.py
```diff
@@ -0,0 +1,681 @@
...
+        try:
+            Parallel(n_jobs=effective, backend="loky", verbose=2)(
```
clessig: Unless there are strong reasons otherwise, we should use one approach/library for parallelization

packages/evaluate/src/weathergen/evaluate/utils/plotting.py
```diff
@@ -0,0 +1,681 @@
...
+            with contextlib.suppress(Exception):
+                get_reusable_executor().shutdown(wait=True)
+            for t in tqdm(fstep_tasks, desc=f"Score maps {stream} (sequential)"):
+                _score_map_fstep_worker(**t)
+    else:
+        for t in tqdm(fstep_tasks, desc=f"Score maps {stream}"):
+            _score_map_fstep_worker(**t)
+
+
+def _score_map_fstep_worker(
```
clessig: Does this function have a purpose? It seems to just contain a single function call with the same args

packages/evaluate/src/weathergen/evaluate/utils/plotting.py
```diff
@@ -0,0 +1,681 @@
...
+                    "plot_target": plot_target,
+                    "plot_histograms": plot_histograms,
+                    "maps_config": maps_config_dict,
+                    "bias_config": bias_config_dict,
+                }
+            )
+
+    effective_workers = min(num_plot_workers, len(tasks))
+
+    if effective_workers > 1 and len(tasks) > 1:
```
clessig: Again: do we really need this branch?

packages/evaluate/src/weathergen/evaluate/utils/plotting.py
```diff
@@ -0,0 +1,681 @@
...
+        _logger.info(
+            f"Parallel plotting: dispatching {len(tasks)} (fstep, sample) tasks "
+            f"across {effective_workers} loky workers."
+        )
+        try:
+            Parallel(n_jobs=effective_workers, backend="loky", verbose=2)(
```
clessig: How does this parallelization differ from the one in plotter and above, i.e. does it make sense to have both and we don't have nested parallelism?

packages/evaluate/src/weathergen/evaluate/scores/score_orchestration.py
```diff
@@ -0,0 +1,406 @@
+# (C) Copyright 2025 WeatherGenerator contributors.
```
clessig, line 1: Why is this file all new?

packages/evaluate/src/weathergen/evaluate/run_evaluation.py
```diff
@@ -271,39 +265,86 @@ def _process_stream(
         _logger.info(f"No evaluation config for {run_id} - {stream}. Skipping.")
         return run_id, stream, {}
 
-    # Parallel plotting
-    if stream_dict.get("plotting") and type_ == "zarr":
-        plot_data(reader, stream, global_plotting_opts)
+    # -----------------------------------------------------
```
clessig: Standard comment

clessig (commented):
Review part II. 

From the code it seems to me we have nested parallelism (although it's hard to tell for sure). This cannot be used since the nested threads fight over resources. We should also use one approach/lib for parallelization unless there are very strong reasons otherwise.

I can have another, more fine-grained look once these comments have been addressed.

packages/evaluate/src/weathergen/evaluate/io/data/get_data_orchestration.py
iluise: yes, they are in random order, but this line` per_fstep.sort(key=lambda x: x[0])` within the `_reassemble_fsteps` function compensates for it

packages/evaluate/src/weathergen/evaluate/io/data/get_data_orchestration.py
iluise: this is old code based on the current `zarr_reader` in `common`. I removed all this part

packages/evaluate/src/weathergen/evaluate/io/data/raw_io_orchestration.py
iluise: that was an older backup version based on reading data from `get_data` in `common`. removed `_dispatch_reads()` now

packages/evaluate/src/weathergen/evaluate/io/data/xarray_utils.py
iluise: this is already implemented in DeriveChannels, one just need to add "10ff" to the list of variables. I moved the scaling function there as well so they are all in the same place, good point.

packages/evaluate/src/weathergen/evaluate/io/data/xarray_utils.py
iluise: still from the backup of the old get_data. removed.

packages/evaluate/src/weathergen/evaluate/io/data/xarray_utils.py
iluise: same as above. removed

packages/evaluate/src/weathergen/evaluate/scores/score_orchestration.py
iluise: I removed `utils.py` and moved the scoring functions to a dedicated file. 
`utils.py` contained functions for scoring and plotting all merged tougher and it was > 1000 lines. Now the structure is:

```
scores:
   -> scores_orchestration.py --> parallelisation and high level functions
   -> scores_utils.py -> helpers
   -> scores.py -> underlying scoring class
```

same for plotting:

```
plotting:
   -> plot_orchestration.py --> parallelisation and high level functions
   -> plot_utils.py -> helpers
   -> plotter.py -> underlying plotting class
```

iluise:
it is ready for review, this is the logic: 

architecture_diagram


- I removed all the functions related to the current `zarr_io` reader class in `common` so the code is less redundant now. 
- plotting/scoring and I/O now use the same parallelisation strategy
- I played around with the num workers it is indeed an advantage, but we reach a plateau and having 36 workers seems faster than 48 or 72 in terms of I/O. Is it ok to keep it at 36 max or I should dig up more on the cause?

clessig:
> it is ready for review, this is the logic:
> 
> architecture_diagram
> * I removed all the functions related to the current `zarr_io` reader class in `common` so the code is less redundant now.
> * plotting/scoring and I/O now use the same parallelisation strategy
> * I played around with the num workers it is indeed an advantage, but we reach a plateau and having 36 workers seems faster than 48 or 72 in terms of I/O. Is it ok to keep it at 36 max or I should dig up more on the cause?

Thanks for the work and the overview diagram. 

36 workers is likely a good choice for all GH200 machines but probably too high for many others (eg Leonardo). It also makes sense that there is a degradation if the number is too high (contention about IO resources). I am fine to fix 36 in the code as long as there is an option to overwrite it from the command line.

packages/evaluate/src/weathergen/evaluate/utils/derived_channels.py
```diff
@@ -121,25 +126,41 @@ def get_derived_channels(
...
-        - data_tars: Target dataset
-        - data_preds: Prediction dataset
-
-        Returns:
-        --------
-        - data_tars: Updated target dataset (if channel can be added)
-        - data_preds:  Updated prediction dataset (if channel can be added)
-        - self.channels: all the channels of interest
-
+        Derive channels from available channels in the data.
```
clessig, line 129: We ultimately would like to have a more generic form that takes required channels and a conversion function as arguments. This would eg allow to also convert between wind and wind direction (or vice versa, as needed for CERRA) and specific and relative humidity. Maybe we can open an issue for that

packages/evaluate/src/weathergen/evaluate/scores/score_orchestration.py
```diff
@@ -0,0 +1,369 @@
...
+                        data_dict = {"scores": [data_dict]}
+                    scores = data_dict.get("scores")
+                    for i, existing_score in enumerate(scores):
+                        if existing_score["attrs"] == metric_data.attrs:
+                            _logger.warning("Metric with same parameters found, replacing")
+                            scores[i] = metric_data_dict
+                            break
+                    else:
+                        scores.append(metric_data_dict)
+                        _logger.info(f"Appending results to {save_path}")
```
clessig: Should this be debug?

packages/evaluate/src/weathergen/evaluate/scores/score_orchestration.py
```diff
@@ -0,0 +1,369 @@
...
+                else:
+                    _logger.info(f"Saving results to new file {save_path}")
```
clessig: Should this be debug?

packages/evaluate/src/weathergen/evaluate/scores/score_orchestration.py
```diff
@@ -0,0 +1,369 @@
...
+        for region in regions:
+            for run_id, metric_data in metric_stream[region][stream].items():
+                save_path = (
+                    reader.metrics_dir
+                    / f"{run_id}_{stream}_{region}_{metric}_chkpt{reader.mini_epoch:05d}.json"
+                )
+                metric_data_dict = metric_data.to_dict()
+
+                if save_path.exists():
+                    _logger.info(f"{save_path} already present")
```
clessig: debug

packages/evaluate/src/weathergen/evaluate/scores/score_orchestration.py
```diff
@@ -0,0 +1,369 @@
...
+            calls,
+            n_workers=n_workers,
+            backend="threading",
+            desc=f"Scoring {reader.run_id} - {stream} {region}",
+        )
+        fstep_results = [r for r in all_results if r is not None]
+
+        fstep_results.sort(key=lambda r: r[0])
+
+        for fstep, combined_metrics, fstep_attrs in fstep_results:
```
clessig: This is not doing actual computations but just populating the result data structure?

clessig: It seems it would be good to split the contents of this function up as `compute_scores_for_region()` and `store_metrics_for_region()`

packages/evaluate/src/weathergen/evaluate/scores/score_orchestration.py
```diff
@@ -0,0 +1,369 @@
...
+    preds_next: xr.DataArray | None,
+    tars_next: xr.DataArray | None,
+    climatology: xr.DataArray | None,
+    bbox: "RegionBoundingBox",
+    metrics: dict,
+    group_by_coord: str | None,
+) -> tuple[int, xr.DataArray, dict[tuple[int, str], dict]] | None:
+    """Score all metrics for one fstep in one region. Stateless, thread-safe.
+
+    All inputs are immutable xarray DataArrays. All numpy/xarray operations
```
clessig: Remove comment about GIL

packages/evaluate/src/weathergen/evaluate/scores/score_orchestration.py
```diff
@@ -0,0 +1,369 @@
...
+from weathergen.evaluate.scores.score import VerifiedData, get_score
+from weathergen.evaluate.utils.array_utils import scalar_coord_to_dim
+from weathergen.evaluate.utils.clim_utils import get_climatology
+from weathergen.evaluate.utils.regions import RegionBoundingBox
+
+_logger = logging.getLogger(__name__)
+
+
+def get_next_data(fstep, da_preds, da_tars, fsteps):
+    """Get the next forecast step data for the given forecast step."""
```
clessig, line 30: Then the function should also be called `get_next_fstep_data()`

packages/evaluate/src/weathergen/evaluate/plotting/plotter.py
```diff
@@ -611,75 +614,16 @@ def healpixlines(
...
-            Selection to be applied to the DataArray
-        tag: str
-            Any tag you want to add to the plot
-
-        Returns
-        -------
-            List of plot names for the saved animations.
+        """Return the task descriptors for building GIF animations.
 
+        .. deprecated::
```
clessig: Do we really want to commit this?

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
```diff
@@ -0,0 +1,701 @@
...
+    output_data : ReaderOutput | None
+        Pre-loaded data; when provided ``reader.get_data()`` is skipped.
+    """
+    if not reader.is_gridded_data(stream):
+        _logger.debug(f"RUN {reader.run_id} - {stream}: Skipping score maps (non-gridded data).")
+        return
+
+    map_dir = reader.runplot_dir / "plots" / stream / "score_maps"
+    map_dir.mkdir(parents=True, exist_ok=True)
+    _logger.info(f"RUN {reader.run_id} - {stream}: Plotting score maps to {map_dir}")
```
clessig: ... Saving score maps to ...

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
```diff
@@ -0,0 +1,701 @@
...
+            stream, fsteps=fsteps, samples=samples, channels=channels, ensemble=ensemble
+        )
+
+    da_preds = output_data.prediction
+    da_tars = output_data.target
+    fsteps = sorted(da_preds.keys())
+    aligned_clim_data = get_climatology(reader, da_tars, stream)
+
+    n_plot_workers = resolve_num_workers(
+        int(reader.eval_cfg.get("num_plot_workers", 0)), check_process_headroom=True
```
clessig: Do we really need `num_plot_workers`? When would the number of parallel workers for plotting and other tasks (say IO) differ?

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -589,6 +496,12 @@ def is_gridded_data(self, stream: str) -> bool:
         -------
             True if the stream is regularly spaced. False otherwise.
         """
+        if stream not in self._cached_is_gridded:
+            self._cached_is_gridded[stream] = self._compute_is_gridded(stream)
+        return self._cached_is_gridded[stream]
+
+    def _compute_is_gridded(self, stream: str) -> bool:
+        """Original is_gridded_data logic, called once per stream and cached."""
```
clessig: Is "Original" indeed useful as a comment

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -358,165 +366,57 @@ def __init__(self, eval_cfg: dict, run_id: str, private_paths: dict | None = Non
...
+            sorted(int(f) for f in (fsteps or self.get_forecast_steps())),
+            sorted(int(s) for s in (samples or self.get_samples())),
+            to_list(channels or self.get_stream(stream).get("channels", self.get_channels(stream))),
+            resolved_ensemble,
+            self._num_io_workers,
+            ens_select,
+        )
+        if state.is_zip:
+            return get_data_zip_impl(state)
+        return get_data_impl(state)
```
clessig: Get data, then one return statement. The function `get_data_zipstore()` and `get_data_dirstore()` seem also clearer.

packages/evaluate/src/weathergen/evaluate/scores/score_orchestration.py
```diff
@@ -0,0 +1,369 @@
...
+"""Score orchestration: per-fstep scoring, stream aggregation, and JSON output."""
+
+import json
+import logging
+
+import numpy as np
+import xarray as xr
+from joblib import delayed
+
+from weathergen.evaluate.io.data.io_orchestration import dispatch_parallel, resolve_num_workers
```
clessig: `get_num_workers()` seems to be a better name

packages/evaluate/src/weathergen/evaluate/io/data/io_orchestration.py
```diff
@@ -0,0 +1,623 @@
...
+    n_workers: int
+    backend: str = "loky"
+
+
+# ---------------------------------------------------------------------------
+# Shared helpers
+# ---------------------------------------------------------------------------
+
+
+def resolve_num_workers(requested: int = 0, *, check_process_headroom: bool = False) -> int:
```
clessig: get_num_workers

packages/evaluate/src/weathergen/evaluate/io/data/io_orchestration.py
```diff
@@ -0,0 +1,623 @@
...
+    * When *n_workers* ≤ 1 **and** the backend is ``"loky"`` we also skip the
+      ``Parallel`` call to avoid any pool-creation overhead.
+    """
+    n_tasks = len(calls)
+    if n_tasks == 0:
+        return []
+
+    effective = min(n_workers, n_tasks)
+
+    # --- fast path: skip Parallel entirely when sequential ----------------
```
clessig: Can we remove the `---` etc

packages/evaluate/src/weathergen/evaluate/io/data/io_orchestration.py
```diff
@@ -0,0 +1,623 @@
...
+    if effective <= 1 and backend == "loky":
+        return [c[0](*c[1], **c[2]) for c in calls]
+
+    # --- threading or loky-with-multiple-workers --------------------------
+    if backend != "loky":
```
clessig: Why do we need this separate branch? What is the harm of using the try/except statement also with threading

packages/evaluate/src/weathergen/evaluate/io/data/io_orchestration.py
```diff
@@ -0,0 +1,623 @@
...
+            get_reusable_executor().shutdown(wait=True)
+        return results
+    except Exception as exc:
+        _logger.warning(
+            f"{desc}: parallel pool failed ({type(exc).__name__}: {exc}). "
+            f"Falling back to sequential."
+        )
+        with contextlib.suppress(Exception):
+            get_reusable_executor().shutdown(wait=True)
+        return [c[0](*c[1], **c[2]) for c in calls]
```
clessig: Again, `results = ...` and have one return statement

packages/evaluate/src/weathergen/evaluate/io/data/io_orchestration.py
```diff
@@ -0,0 +1,623 @@
...
+            is_zip=state.is_zip,
+            need_coords=not state.is_gridded,
+            is_gridded=state.is_gridded,
+            n_workers=n_workers,
+            backend=state.backend,
+            label=f"RUN {state.run_id} - {state.stream} fstep {fs}",
+        )
+        # If _parallel_read fell back to sequential, honour that for the rest
+        if n_workers > 1 and len(results) == len(state.samples):
+            # check if it actually ran sequentially (no exception path exposed,
```
clessig: But why don't you pass a return flag from `_read_samples()` so that you have the information

packages/evaluate/src/weathergen/evaluate/io/data/io_orchestration.py
```diff
@@ -0,0 +1,623 @@
...
+def get_data_zip_impl(state: IOState) -> ReaderOutput:
+    """ZipStore fast-path: dispatch *all* (sample, fstep) pairs in parallel.
+
+    Each worker opens its own ZipStore handle, so the zip central-directory
+    is parsed once per worker (not once per task) thanks to loky worker
+    reuse.  With ``samples × fsteps`` tasks the pool utilises all available
+    workers instead of being capped at ``len(samples)``.
+    """
+    n_total = len(state.samples) * len(state.fsteps)
+    _logger.info(
```
clessig, line 548: Should this really be info. Do people really care about the information when they run eval?

packages/evaluate/src/weathergen/evaluate/io/data/io_workers.py
```diff
@@ -0,0 +1,257 @@
...
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""Module-level loky worker functions and direct zarr I/O helpers.
+
+These are module-level (not class-bound) so they are picklable and can be
```
clessig: picklable -> pickable

packages/evaluate/src/weathergen/evaluate/io/data/dataarray_postprocessing.py
```diff
@@ -0,0 +1,139 @@
+# (C) Copyright 2025 WeatherGenerator contributors.
```
clessig, line 1: Is the file name really appropriate? This aren't generic utils for xarray but utility function for our data representation (that happens to use xarray)

clessig (approved):
Minor comments to improve clarity of code. Someone from the eval team should also review and test (which I didn't do). 

On the overview figure. Could we add a 1-sentence description to each of the colored blocks.

iluise:
architecture_diagram

I updated the code after the latest round of comments. This is the new architecture. 

I will open a follow up issue to fix the json and the re-computation of the scores for streams with sub-steps (e.g. hourly era5).

PR #2146, opened 2026-03-31
Feature/visualize masking

### LINKED ISSUE #2145: Initiative: Visualize masking before pretraining/finetuning

As many team members requested to have this feature of previewing masked source and target before starting any pretraining or fine-tuning, I am adding a script to provide a preview of how the data would look like before submitting any job. Easy to use script with one line.

_No response_

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation
- [ ] performance

wael-mika:
## Description

Implementation of the masking visualization script

## Issue Number
Closes #2145 


src/weathergen/utils/visualize_masking.py
```diff
@@ -0,0 +1,1133 @@
...
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""
+Visualize masked source/target samples using the training data pipeline.
+
+This script loads a config, builds a MultiStreamDataSampler, extracts one batch,
+and plots a single variable for source and target with masking/cropping applied.
+This script can run on a cpu or logging node without GPUs.
+Please activate your .venv before running.
```
tjhunter, line 16: this is wrong. You should not have to do that. `uv run` should handle running without activating the environment.

src/weathergen/utils/visualize_masking.py
```diff
@@ -0,0 +1,1133 @@
...
+
+    test_cfg = wg_config.merge_configs(training_cfg, cf.get("validation_config", {}))
+    test_cfg = wg_config.merge_configs(test_cfg, cf.get("test_config", {}))
+    test_cfg = filter_config_by_enabled(test_cfg, keys_to_filter)
+    return test_cfg, VAL
+
+
+def _to_numpy(arr):
+    if isinstance(arr, torch.Tensor):
+        return arr.detach().cpu().numpy()
```
TillHae, line 297: Why do we have to move this to cpu?

src/weathergen/utils/visualize_masking.py
```diff
@@ -0,0 +1,1133 @@
...
+        Boolean mask over HEALPix cells (True = visible).
+    healpix_level : int
+        HEALPix resolution level (nside = 2^level).
+
+    Returns
+    -------
+    np.ndarray
+        Boolean array of same length as lats/lons indicating visibility.
+    """
+    if len(lats) == 0:
```
TillHae, line 347: Is len(lats) == len(lons) all the time?

src/weathergen/utils/visualize_masking.py
```diff
@@ -0,0 +1,1133 @@
...
+
+    Returns
+    -------
+    tuple[np.ndarray, np.ndarray, np.ndarray]
+        Downsampled latitude, longitude, and value arrays.
+    """
+    if max_points is None or len(vals) <= max_points:
+        return lats, lons, vals
+    idxs = rng.choice(len(vals), size=max_points, replace=False)
+    return lats[idxs], lons[idxs], vals[idxs]
```
TillHae, line 589: This might be dangerous to do as it generates a discontinuous array that looks continuous

src/weathergen/utils/visualize_masking.py
```diff
@@ -0,0 +1,1133 @@
...
+
+    include_full = len(vals_full) > 0
+    ncols = 3 if include_full else 2
+
+    if vmin is None or vmax is None:
+        vals_ref = vals_full
+        if len(vals_ref) == 0:
+            vals_ref = vals_src if len(vals_src) else vals_tgt
+        vmin = np.nanpercentile(vals_ref, 2)
+        vmax = np.nanpercentile(vals_ref, 98)
```
TillHae, line 663: It doesn't make a big difference in most cases, but I think cutting the outliers can be unwanted sometimes, where we are interested in them and want to see it in our plot. Especially when evaluating masking this can be interesting.

src/weathergen/utils/visualize_masking.py
```diff
@@ -0,0 +1,1133 @@
...
+        if target_view == "target_values":
+            target_var_idx, target_var_name = _resolve_var_idx(
+                ds.target_channels,
+                target_var_name,
+                "target",
+                strict=parsed.target_variable is not None,
+            )
+        else:
+            target_var_idx, target_var_name = _resolve_var_idx(
+                ds.source_channels,
```
TillHae, line 893: Why do we have to use `ds.source_channels` when we have no target_tokens in our target stream?

src/weathergen/utils/visualize_masking.py
```diff
@@ -0,0 +1,1133 @@
...
+        )
+        tgt_label = (
+            f"Target ({target_var_name}, {target_view})\n"
+            f"{_format_mask_params(_to_dict(target_meta.params))}\n"
+            f"points={len(vals_tgt)}"
+        )
+
+        pairs_data.append(
+            {
+                "pair_idx": source_idx,
```
TillHae, line 970: Shouldn't the key "pair_idx" point to the value of `pair_indices` defined in l.829

TillHae (commented):
Looks good, I just have a few concerns with your script.

**I didn't check parsing, docstrings, plotting, meta info, instance checks and didn't test the script.**

src/weathergen/utils/visualize_masking.py
wael-mika: I agree with you, but after dealing with different systems, there are different Python versions running on each of those, and I found out that using uv run is the easiest way to handle this situation. I can add a command using python as well if you agree

src/weathergen/utils/visualize_masking.py
wael-mika: This script can be run completely on cpu, we do not need GPU for it

src/weathergen/utils/visualize_masking.py
wael-mika: Yes, always. Both lats and lons are derived from the same coords array via coords[:, 0] and coords[:, 1], so they are guaranteed to have equal length at every call site.

src/weathergen/utils/visualize_masking.py
wael-mika: I agree, but this is not an evaluation plot, so extreme events are prevented here to keep the colorscale consistent.

src/weathergen/utils/visualize_masking.py
wael-mika: When visualizing masking, I am choosing only one variable, which is enough for this purpose. Source and target here represent the student and teacher view which are different and can be set up in the config

src/weathergen/utils/visualize_masking.py
wael-mika: source_idx already is a value from pair_indices (the loop iterates for source_idx in pair_indices). So "pair_idx": source_idx stores the source sample index, which is meaningful (it identifies which batch sample this is). Storing the enumeration position (0, 1, 2…) would be less informative and would make the filename pair0_, pair1_ regardless of which samples were selected.
Though the naming pair_idx vs source_idx is admittedly redundant — pair_idx could be renamed to make it clearer.

src/weathergen/utils/visualize_masking.py
wael-mika: Disagree. For scatter plots. Unlike line plots, scatter does not connect adjacent points. Each point is rendered independently, so random downsampling only reduces density. It doesn't create false continuity. The points=N label already shows how many points are displayed. 
_downsample is only used when you set up "--max-points" for high resolution data and it is not used in individual cases if not specificaly sat up

src/weathergen/utils/visualize_masking.py
TillHae: But if I have a gpu on my local machine I might want to use it for performance

TillHae:
Thanks for clarifying. Now everything sounds logical to me.

src/weathergen/utils/visualize_masking.py
wael-mika: I believe most of users will be using HPCs as the data is stored there. That said, it is only retrieving one sample so cpu is perfectly suited even on any machine. The performance is not important here as the script is used to generate one plot before training.

TillHae:
We should decide if this is useful enough to others to merge or too niche to add to the repo.

clessig:
> We should decide if this is useful enough to others to merge or too niche to add to the repo.

It's useful but it should no go into the main repo but into the private one. It would also be important to have a mechansim to ensure that the code is kept functional. The masking code is relatively stable but we should know when the visualization code breaks.

PR #2212, opened 2026-04-14
Add attention base class

### LINKED ISSUE #2130: Refactor attention code

The attention code is currently scattered in many different functions with a lot of code duplication. This should be refactored to avoid the duplication as much as possible.

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation
- [ ] performance

TillHae:
## Description

Adds base class for attention mechanism that all other attention classes inherit from.
This avoids duplications as much as possible.

CC @clessig, @csjfwang

## Issue Number

Closes #2130 


TillHae:
Maybe we should create a separate PR for just ruffing the repository again. The failed `cscs/test-single` test seems to already fail at creating the environment.

TillHae:
@csjfwang do you have some time in the upcoming weeks to review it? You can ignore all changes that are not in `attention.py`. They are just due to linting.

TillHae:
@clessig is the refactoring of the attention class still needed or do you want to close the issue and PR?

clessig:
> @clessig is the refactoring of the attention class still needed or do you want to close the issue and PR?

Yes, that would still be good

TillHae:
Here is the UML-diagram showing how the structure was before and is currently:

attention_before_PR2212

Here is the UML-diagram showing how the structure would be with this PR:

attention_after_PR2212

src/weathergen/model/attention.py
```diff
@@ -121,62 +147,36 @@ def forward(self, x, x_lens, ada_ln_aux=None, coords=None):
...
+        self.dim_head_proj = dim_embed // self.num_heads if dim_head_proj is None else dim_head_proj
 
-        if norm_type == "LayerNorm":
-            norm = partial(torch.nn.LayerNorm, elementwise_affine=False, eps=norm_eps)
-        else:
-            norm = RMSNorm
+        self._make_qk_lnorms()
 
-        self.lnorm = norm(dim_embed, eps=norm_eps)
+        self.lnorm = self.norm(dim_embed, eps=self.norm_eps)
```
MatKbauer: Let's remove the `eps=self.norm_eps` argument from the `self.norm` initialisations, since `self.norm_eps` is anyway defined in the `AttentionBase` class and can be used there directly instead of passing it at each `self.norm` initialisation.

src/weathergen/model/attention.py
```diff
@@ -121,62 +147,36 @@ def forward(self, x, x_lens, ada_ln_aux=None, coords=None):
...
         self.proj_heads_q = torch.nn.Linear(dim_embed, num_heads * self.dim_head_proj, bias=False)
         self.proj_heads_k = torch.nn.Linear(dim_embed, num_heads * self.dim_head_proj, bias=False)
         self.proj_heads_v = torch.nn.Linear(dim_embed, num_heads * self.dim_head_proj, bias=False)
         self.proj_out = torch.nn.Linear(dim_embed, dim_embed, bias=False)
```
MatKbauer: Can't these be modularised in the `BaseAttention` class too?

TillHae: I can add a method in `BaseAttention` that gets the current `dim_embeds` and creates the projection heads based upon this

MatKbauer (approved):
Looks good to me and tested successfully. Similar convergence behaviour as on `develop`. Thanks for taking this on, Till, it reads much cleaner now. Ready to be merged from my perspective.

@clessig, would you like to have another look, or can I go forward with merging?

TillHae:
> Looks good to me and tested successfully. Similar convergence behaviour as on `develop`. Thanks for taking this on, Till, it reads much cleaner now. Ready to be merged from my perspective.
> 
> @clessig, would you like to have another look, or can I go forward with merging?

Great to hear, thanks for taking the time to test and review it!

sophie-xhonneux:
Hey @TillHae! Thanks for all the hard work already, very exciting to see!

I think in some way the PR doesn't go far enough and doesn't solve the original problem, we still have a lot classes to manage for basic changes to the attention, e.g. how RoPE is applied.

Let's maybe have a discussion about this? A couple points I am thinking about:

- Let's see if we can separate the attention kernel variant (flex, FA, torch) away from attention mechanism design decisions like Global vs Local
- I suspect the difference between self and cross attention is also not needed, both use the same attention(q, k, v) with different inputs for q,k,v
- Let's write a clean base class that easily captures the flow from input -> projection -> RoPE-like stuff -> attention variant (global, local, etc) -> dropout? -> XSA -> residual connection
- There are a couple things to consider like Muon interacts with how you store the weight matrices of multi-head attention
- We should easily be able to allow for GQA attention
- Also varlen or not should ideally use the same code with an Optional argument for the varlen that I think only needs to be passed through to the attention kernel variant API. (I forget what the sliced Q is used for)

Maybe you can start with some pseudo-code and we can continue the discussion from there?

sophie-xhonneux (changes requested):
See previous comment

TillHae:
@sophie-xhonneux sorry for the late reply but I needed to look into some stuff to address your comments as I am not so experienced with the science behind the attention. I created a hedgedoc report to keep it more organised:

https://gitlab.jsc.fz-juelich.de/hedgedoc/sJCn1pUMRPS3Ytfj0qsNtQ

I propose a new 4 class architecture that should be feasible to implement and goes beyond the current version. If you want I can also add pseudo-code later but please first let me know if you think the new proposal is a valid option for us.

sophie-xhonneux:
@TillHae Could you allow me to edit the hedgedoc please!

TillHae:
> @TillHae Could you allow me to edit the hedgedoc please!

Done

PR #2208, opened 2026-04-13
Fix issue 2205

sophie-xhonneux:
## Issue Number

Closes #2205 #2222 

Is this PR a draft? Mark it as draft.

## Checklist before asking for review

-   [x] I have performed a self-review of my code
-   [x] My changes comply with basic sanity checks:
      - I have fixed formatting issues with `./scripts/actions.sh lint`
      - I have run unit tests with `./scripts/actions.sh unit-test`
      - I have documented my code and I have updated the docstrings.
      - I have added unit tests, if relevant
-   [ ] I have tried my changes with data and code:
      - I have run the integration tests with `./scripts/actions.sh integration-test`
      - (bigger changes) I have run a full training and I have written in the comment the run_id(s): `launch-slurm.py --time 60`
      - (bigger changes and experiments) I have shared a hegdedoc in the github issue with all the configurations and runs for this experiments
-   [ ] I have informed and aligned with people impacted by my change:
    - for config changes: the MatterMost channels and/or a design doc
    - for changes of dependencies: the MatterMost software development channel

PR #2228, opened 2026-04-17
Sgrasse/sophiex/2205

grassesi:
Small patch that patches up all config files containing `norm_type` with also `qk_norm_type` and applies fixes to all base config specified with `--base-config` flag.


clessig:
Should we really fix all the configs? I think this is the better solution: #2229

grassesi:
> Should we really fix all the configs? I think this is the better solution: #2229
Ok I will close this PR then. Should I also go ahead and close #2208 since it every functionality it implements would be redundant?

PR #2262, opened 2026-04-23
Fix multi-stream training

### LINKED ISSUE #2261: Multi-stream training broken
### What happened?

Multi-stream training broken due to broken scatter index computation

_No response_

_No response_

clessig:
## Description

Fix multi-stream training, i.e. scatter index computation

## Issue Number

Closes #2261 


clessig:
Convergence for broken and fixed code (see legend):

image

src/weathergen/model/engines.py
```diff
@@ -106,27 +110,81 @@ def forward(self, batch, pe_embed):
...
-        repeat = torch.repeat_interleave
-        scatter_idxs = repeat(
-            torch.ones(len(tok_counts), dtype=torch.int64, device=tok_counts.device), tok_counts
-        )
-        scatter_idxs = scatter_idxs.cumsum(0) - 1
-        # scatter index must exist for each element and not just per row
-        scatter_idxs = scatter_idxs.unsqueeze(1).repeat((1, self.cf.ae_local_dim_embed))
+        if batch.tokens_lens.shape[2] == 1:
+            # trivial with one stream
+            tokens_all = peh(torch.cat(x_embeds))
```
shmh40: Shouldn't we do the same thing here that we do below for the multiple stream case?

Do the reordering again and then

    pe_idxs = rows[rows < tok_counts.unsqueeze(1)]
    tokens_all = torch.cat(x_embeds) + pe_embed[pe_idxs]

?

src/weathergen/model/model.py
```diff
@@ -180,11 +180,15 @@ def reset_parameters(self, cf: Config) -> "ModelParams":
         # positional encodings
 
         dim_embed = cf.ae_local_dim_embed
-        len_token_seq = 1024
+        len_token_seq = 32
+        token_idx_bias = 16
```
shmh40, line 185: These are new changes that may separately slightly affect model performance? Should in principle be a separate PR, but I am fine with this for now, I don't expect the change to be much.

src/weathergen/model/engines.py
```diff
@@ -96,6 +97,9 @@ def forward(self, batch, pe_embed):
                 for sample in batch.get_samples():
                     sdata += [sample.streams_data[stream_name].source_tokens_cells[istep]]
 
+            if all(s is None for s in sdata):
```
shmh40, line 101: Great

src/weathergen/model/model.py
```diff
@@ -93,7 +93,7 @@ def __init__(self, cf) -> None:
         self.dtype = get_dtype(cf.attention_dtype)
 
         # Positional embeddings
```
shmh40, line 96: I suppose we do not expect to ever have more than 32 tokens per cell, so I think this makes sense. But this may fail for more than 32 tokens per cell with an indexerror...so if someone makes a mistake/poor choice in the stream config this will fail? Maybe we should have an error + warning rather than uninformative index error. Pretty unlikely but might be worthwhile

src/weathergen/model/engines.py
```diff
@@ -106,27 +110,81 @@ def forward(self, batch, pe_embed):
...
+    def get_scatter_idxs(self, batch):
+        """
+        Compute reordering index so that tokens from different streams but same cell are
+        continguous
+
+        Simple version (reference implementation)
+        """
+
+        dev = batch.get_device()
+        # batch.tokens_lens : (num_samples, num_steps_input, num_streams, num_cells)
```
shmh40: **num_steps**, num_samples, num_streams, num_cells ?

src/weathergen/model/engines.py
```diff
@@ -106,27 +110,81 @@ def forward(self, batch, pe_embed):
...
-        # actual scatter operation
-        tokens_all.scatter_(0, scatter_idxs, torch.cat(x_embeds) + pe_embed[pe_idxs])
+            # per cell indices into positional encoding
+            tok_counts = batch.tokens_lens.permute([2, 0, 1, 3]).sum(0).flatten()
+            rows = torch.arange(tok_counts.max(), device=tok_counts.device).unsqueeze(0)
+            rows = rows.expand(tok_counts.shape[0], -1)
+            pe_idxs = rows[rows < tok_counts.unsqueeze(1)]
+
+            # actual scatter operation
+            tokens_all.scatter_(0, scatter_idxs, torch.cat(x_embeds) + pe_embed[pe_idxs])
```
shmh40: I am not certain this is right. We are adding the pe (computed from per-cell) to the x_embeds (still stream first), and *then* applying the scatter?

So the pe_embed should just be added after we do the scatter, so that the x_embeds are now tokens_all and they are in the appropriate per cell order?

shmh40 (approved):
There are comments. Approving because I am not sure how the internet will be, but please have a look at the comments before merging.

src/weathergen/model/engines.py
clessig: This is required when spoofing is disabled/removed

src/weathergen/model/model.py
clessig: We might hit for GEOs or CERRA. Will check and add an error message

clessig: The retuning was helpful. The blue line above is with the old, black with new tuning. The dark blue line is quite a bit more noisy

clessig: CERRA (with token_size 128) leads to 53 tokens. I raised the limit to 64, which also works for the GEOs and refactored to make it cleaner

clessig:
Improved CERRA convergence:

s1k4nwsf_hk50o93c_tjlso3p4_w9f6d82s_r4zhnc5q_cdrsq7xp_wfpk9lgw_mlpgfesd_train_fs_2_mse_ERA5_2t

PR #2286, opened 2026-04-29
tp fixed for zarr_dt = 0

### LINKED ISSUE #2285: export: tp not working for forecast_step = 0
### What happened?

For the `verif_parser` the `tp` variable is calculated across a whole time step available. If only one forecast step is provided and the time step is equal to zero it is not possible to accumlate across a time value. In this scenario we should return only the values available. 

In the future the handling of inference of 1/2h time steps (or time steps less than 1h) will need to be reviewed (for now NaN values will be used but will see how requirements evolve)

_No response_

_No response_

### LINKED ISSUE #2395: one file saved per batch in verif_parser
### What happened?

* more investigation needed
* seems that number of valid_times in data = number of batches rather than true number of samples
* investigate if save being called after processing sample is causing issues for verif

_No response_

_No response_

enssow:
## Description

Adds handling for when zarr_dt (time between forecast_steps) is 0, which happens when there is only one forecast step with 0h lead time

## Issue Number

Closes #2285 
Closes #2395


enssow:
This PR now contains some initially unrelated changes but they are important to merge as they relate to verif and computing the `tp` variable correctly, for ease of review:

- In the situation where there is only one forecast step and the forecast offset is 0, it means that the forecast step calculated from the difference between source_interval_start and source_interval_end is 0. The code then did not use the data available for the across the forecast step window which resulted in incorrect comparisons. The code now looks for the default forecast step from the inference config. If the inference has been copied over from an HPC this will not be found so the CLI argument `fstep_hours` is used to provide the hours of the forecast window (default is 6h as this is in the default config itself). If a CLI arg is used a warning will be provided. 
- The case where the forecast step might be less than 1 h is not comprehensively tested as there is not a suitable inference available and thus a warning is provided. It should create an array that has hourly values made from the aggregate of intervals within the previous hour, and NaNs for the rest. 
- The `tp` and `tcw` variables have been added to the `config_zarr2cf.yaml`
- The MEPS unit conversion for `config_zarr2verif.yaml` has been rectified
- `get_model_results` is now more flexible to allow the retrieval of the inferences made from multiple ranks and epochs. The default is None for both which then retrieves all the zarr/zip files available.


Testing commands on JUPITER:
(commands testing forecast step 0)
`uv run export --run-id ksvzuzr0 --output-dir ../test_tp2 --format verif --obs /e/home/jusers/owens1/jupiter/WeatherGen/metno_observations_v3.nc --samples 0 1 --fsteps 1`
(examining files)
```
import xarray as xr
import matplotlib.pyplot as plt

verif_ds = xr.open_dataset("/e/home/jusers/owens1/jupiter/WeatherGen/test_tp2/verif/ERA5/tp/ksvzuzr0_ERA5_tp_2d_prediction.nc")
verif_ds

i =  1
x = verif_ds.sel(leadtime= verif_ds["leadtime"].values[0], time=verif_ds["time"].values[i])["lon"].values
y = verif_ds.sel(leadtime= verif_ds["leadtime"].values[0], time=verif_ds["time"].values[i])["lat"].values
z = verif_ds.sel(leadtime= verif_ds["leadtime"].values[0], time=verif_ds["time"].values[i])["fcst"].values
plt.scatter(x, y, c=z, cmap='viridis')
plt.colorbar()
```
image

(commands testing epoch/rank)
`uv run export --run-id ksvzuzr0 --output-dir ../test_tp2 --format netcdf  --samples 0 1 --fsteps 2 3 --epoch 0`
`uv run export --run-id ksvzuzr0 --output-dir ../test_tp2 --format netcdf --rank 1 2 --samples 0 1 --fsteps 2 3`

packages/evaluate/src/weathergen/evaluate/export/parsers/netcdf_parser.py
```diff
@@ -55,6 +55,7 @@ def process_sample(
         self,
         fstep_iterator_results: iter,
         ref_time: np.datetime64,
+        default_fstep: int,
```
iluise, line 58: can you add a description of what this fstep is doing? Why do we need it? For hourly models with substeps this might not work though right? Because the fstep will be 6 but the steps would be hourly spaced. (unless I am missing something)

packages/evaluate/src/weathergen/evaluate/export/export_core.py
```diff
@@ -277,6 +280,29 @@ def get_streams(stream, fname_zarr):
     return streams
 
 
+def get_default_fstep(kwargs, epoch):
```
iluise, line 291: I am not fully understanding why we actually need this, can't we compute the step from the position in the list and the lead_time as valid_time - source_end ?

Like here: https://github.com/ecmwf/WeatherGenerator/blob/develop/packages/evaluate/src/weathergen/evaluate/io/data/io_orchestration.py#L353-L375

packages/evaluate/src/weathergen/evaluate/export/export_core.py
```diff
@@ -304,119 +330,134 @@ def export_model_outputs(data_type: str, config: OmegaConf, **kwargs) -> None:
...
-                    f"Batch {batch_idx + 1}/{n_batches}: "
-                    f"samples {batch_start}–{batch_end - 1} "
-                    f"({len(batch_samples)} samples, {len(batch_tasks)} tasks)"
-                )
+    current_epoch = None
+    fname_zarr_list, epochs = get_model_results(run_id, epoch, rank)
+    processed_samples = []
+    for epoch, fname_zarr in zip(epochs, fname_zarr_list, strict=False):
+        fsteps = get_fsteps(fsteps, fname_zarr)
+        samples = get_samples(samples, fname_zarr)
```
iluise: Did you check that this gives the correct number of samples with multiple ranks right?

iluise (commented):
The quaver parser fails with this error:

```
  File "<string>", line 4, in <module>
  File "/users/iluise/WeatherGenerator/packages/evaluate/src/weathergen/evaluate/export/export_inference.py", line 307, in export_from_args
    export_model_outputs(dtype, config, **kwargs)
  File "/users/iluise/WeatherGenerator/packages/evaluate/src/weathergen/evaluate/export/export_core.py", line 433, in export_model_outputs
    processed = parser.process_sample(
                ^^^^^^^^^^^^^^^^^^^^^^
TypeError: QuaverParser.process_sample() got an unexpected keyword argument 'default_fstep'
```

Can you take a look? Thanks!

packages/evaluate/src/weathergen/evaluate/export/parsers/netcdf_parser.py
enssow: I thought if the user wants to change the hours for the steps of the forecast they add the option `training_config.forecast.time_step=1` for example to do am hourly forecast. If I am wrong happy to be corrected.

packages/evaluate/src/weathergen/evaluate/export/export_core.py
enssow: I wasn't aware of this function - i'll have a look through and see if it will work for the edge case that happened for verif

packages/evaluate/src/weathergen/evaluate/export/export_core.py
enssow: yep, tested with `ksvzuzr0` there were 12 samples across the 4 ranks :)

packages/evaluate/src/weathergen/evaluate/export/export_core.py
enssow: I had a look through at the function - it performs the same role that the offset performs for init time
```
offset = stream_cfg.get("offset")
    if offset is not None:
        offset = np.timedelta64(pd.Timedelta(offset).value, "h")
```
It's a fall back in the sitaution where valid_time - source_interval_start =0 but I still want to save the precip for the preciding 6h (or however long the default_fstep/offset hours is)

Do you suggest to infer it from the offset rather than the training_config.forecast.time_step ? let me know what is better :)

packages/evaluate/src/weathergen/evaluate/export/parsers/netcdf_parser.py
TillHae: The config value `default_fstep` is overwritten with `fstep_hours`, when it isn't given. I think the existence of both is really weird and we should keep only one. If we want the user to be able to give the fstep time as a config parameter and as a command line argument they should at least have the same name and be integrated into one single variable.

@enssow can you fix this as part of this PR?

enssow:
split the PR into sections. parts that fixed the multi zips and multi stream have been merged. Closing this PR until further information is found to indicate the the issue consistently occurs

packages/evaluate/src/weathergen/evaluate/export/parsers/netcdf_parser.py
enssow: Hi Till, sorry this PR is quite outdated and there were no plans to merge so I'm closing it. If we need something similar I'll make sure that there isn't overlap between the variables though. Thank you for taking the time to look into it !

PR #2344, opened 2026-05-11
Fix per stream masking rate

### LINKED ISSUE #2341: Per stream masking rates broken
### What happened?

Per stream masking rates are currently ignored

_No response_

_No response_

clessig:
## Description

Fix per stream masking rate

## Issue Number

Closes #2341 


src/weathergen/datasets/masking.py
```diff
@@ -166,24 +166,22 @@ def merge_masking_config(self, mode_cfg, override):
...
-            stream_cfg_masking["randomly_drop_as_source_rate"] = override[
-                "randomly_drop_as_source_rate"
-            ]
+            stream_cfg["randomly_drop_as_source_rate"] = override["randomly_drop_as_source_rate"]
+
+        # target and source are assumed identical when target is not specified
+        if len(stream_cfg.get("target_input", {})) == 0:
+            stream_cfg["target_input"] = copy.deepcopy(stream_cfg.get("model_input", {}))
 
         for section_key in ("model_input", "target_input"):
```
shmh40, line 175: This enforces that the names of the sections in the stream config are model_input and target_input. I suppose we have to make a choice like this at the moment, but even for this merge the streams configs have source_input, so it is a no-op. 

If we are hard-coding these strings, we should advertise/warn that they must be called this.

config/streams/era5_1deg/era5.yml
```diff
@@ -15,11 +15,9 @@ ERA5 :
   target_exclude : ['w_', 'slor', 'sdor', 'tcw', 'cp', 'tp']
```
shmh40, line 15: There is a test from the original PR tests/test_per_stream_masking_override.py which is now outdated. I am happy to delete this, or it should be updated.

src/weathergen/datasets/masking.py
```diff
@@ -193,16 +191,16 @@ def merge_masking_config(self, mode_cfg, override):
...
         """Build effective masking configs for all streams."""
         cfgs = {}
         for stream_info in streams:
             name = stream_info["name"]
-            override = stream_info.get("masking_override", None)
+            override = stream_info.get("masking_override", {})
             cfgs[name] = self.merge_masking_config(mode_cfg, override)
-            if override is not None and is_root():
+            if override != {} and is_root():
                 logger.info(f"Stream '{name}' using masking override: {override}")
```
shmh40: This will still be logged falsely if there is a mismatch between model_input and here source_input and the override is not actually used.

config/streams/era5_georing_avhrr/avhrr.yml
```diff
@@ -13,7 +13,10 @@ METOP_ABC_AVHRR_IASI :
   filenames : ['observations-ea-ofb-0001-2007-2021-metop-a-iasi-radiances-v1.zarr', 'observations-ea-ofb-0001-2013-2023-metop-b-iasi-radiances-v1.zarr', 'observations-ea-ofb-0001-2019-2023-metop-c-iasi-radiances-v1.zarr']
   geoinfo_channels : ['cos_local_time', 'sin_local_time', 'cos_julian_day', 'sin_julian_day']
   loss_weight : 1.0
-  masking_rate : 0.7
+  masking_override :
+    source_input :
```
shmh40: Need to rename if we keep current logic below

config/streams/era5_georing_avhrr/era5.yml
```diff
@@ -16,7 +16,10 @@ ERA5_in :
   geoinfo_channels : ['lsm', 'slor', 'sdor', 'insolation', 'cos_local_time', 'sin_local_time', 'cos_julian_day', 'sin_julian_day']
   loss_weight : 1.
   location_weight : cosine_latitude
-  masking_rate : 0.6
+  masking_override :
+    source_input :
```
shmh40: Ditto

src/weathergen/datasets/masking.py
```diff
@@ -166,24 +166,22 @@ def merge_masking_config(self, mode_cfg, override):
...
 
         # Copy top-level masking keys from override
         if "randomly_drop_as_source_rate" in override:
-            stream_cfg_masking["randomly_drop_as_source_rate"] = override[
-                "randomly_drop_as_source_rate"
-            ]
+            stream_cfg["randomly_drop_as_source_rate"] = override["randomly_drop_as_source_rate"]
+
+        # target and source are assumed identical when target is not specified
+        if len(stream_cfg.get("target_input", {})) == 0:
```
shmh40: Not sure, but here we are making the target_input identical to the model_input *before* we update the model_input with the per-stream override, so the per-stream target_input will always be the same as the global model_input. Is this desired behaviour?

Perhaps we should do this after we have updated the model_input for the stream? Whether this matters depends on the strategy being used, I think...

config/streams/era5_georing_avhrr/avhrr.yml
clessig: Why?

config/streams/era5_georing_avhrr/era5.yml
clessig: Why?

src/weathergen/datasets/masking.py
clessig: We really in many places that these are the keys used in the config.

src/weathergen/datasets/masking.py
clessig: The issue was there before then. But we can remove

config/streams/era5_georing_avhrr/era5.yml
shmh40: Because source_input will not be read by the code? It needs to be model_input?

config/streams/era5_georing_avhrr/avhrr.yml
shmh40: See below

config/streams/era5_georing_avhrr/era5.yml
clessig: Ah yes!

clessig:
@shmh40 : all comments addressed and cleaned up a bit. Could you have another look

shmh40:
Can you update the configs to have model_input, not source_input? Or this will be a separate PR?

src/weathergen/datasets/masking.py
clessig: Fixed

shmh40 (approved):

PR #2348, opened 2026-05-11
call BatchSample method instead of repeating content

### LINKED ISSUE #1791: Explicit definition of `get_num_steps` in `BatchSamples`

Currently `get_num_steps` recalculates the number of steps every time it is called. We should either calculate it only the first time, then store it as a parameter in the object and use the stored value from this point onwards or already pass num_steps as a parameter to the constructor.

This will ensure robustness and enhance performance (slightly)

https://github.com/ecmwf/WeatherGenerator/blob/develop/src/weathergen/datasets/batch.py#L156

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation

TillHae:
## Description

In [`get_num_source_steps` - `ModelBatch`](https://github.com/ecmwf/WeatherGenerator/blob/develop/src/weathergen/datasets/batch.py#L186) everything from [`get_num_steps` - `BatchSample`](https://github.com/ecmwf/WeatherGenerator/blob/develop/src/weathergen/datasets/batch.py#L186) is just repeated. Instead we can also just call the method on the parameter `target_samples`, which is a `BacthSample` object.

CC @clessig 

## Issue Number

Closes #1791 


grassesi (approved):
This does exactly what it says. I also noticed this function is not used throughout the entire codebase. Instead downstream code in the `weathergen.model` subpackage directly uses `BatchSamples.get_num_steps`. Maybe this method should be entirely removed. What do you think @clessig ?

clessig:
The underlying problem is that the function should read:

```
    def get_num_source_steps(self) -> int:
        """
        Get number of input/source steps from smallest of all available streams
        """
        # TODO: define explicitly
        lens = [
            len(stream.source_tokens_cells)
            for _, stream in self.source_samples.samples[0].streams_data.items()
        ]

        return min(lens)
 ```

That is, it should be `for _, stream in self.source_samples.samples[0].streams_data.items()` instead of `for _, stream in self.target_samples.samples[0].streams_data.items()`. Even better would be to also add a function `get_num_source_steps()` to `StreamData` and perhaps to the intermediate classes like  `Sample`.

@TillHae : could you take care of this?

**Review by clessig — CHANGES_REQUESTED** [review]
See last text comment

**Comment by TillHae** [comment]
What I did with commit d79f737:

- added getter for source and target steps to `StreamData`
- added getter for source and target steps, which iterates through all streams, gets the corresponding length from the `StreamData` getter and returns the minimum to `Sample`
- removed `get_num_steps` from `BatchSamples` and added `get_num_source_steps` and `get_num_target_steps`, which both just use the `Sample` getter
- due to the removal of `get_num_steps` I needed to change all occurences to `get_num_source_steps`
- the getter in `ModelBatch` also just uses the getter from `BatchSamples`

File: src/weathergen/datasets/batch.py [hunk 1]
```diff
@@ -134,6 +134,29 @@ def get_stream_data(self, stream_name: str) -> StreamData:
...
+    def get_num_source_steps(self) -> int:
+        """
+        Get number of source steps from smallest of all available streams
+        """
+        lens = [
+            stream.get_num_source_steps()
+            for _, stream in self.streams_data.items()
+            if stream is not None
+        ]
+        return min(lens) if lens else 0
```
clessig: Can we be more explicit here: ... if len(lens) > 0 else 0

src/weathergen/datasets/batch.py
```diff
@@ -134,6 +134,29 @@ def get_stream_data(self, stream_name: str) -> StreamData:
...
+    def get_num_target_steps(self) -> int:
+        """
+        Get number of target steps from smallest of all available streams
+        """
+        lens = [
+            stream.get_num_target_steps()
+            for _, stream in self.streams_data.items()
+            if stream is not None
+        ]
+        return min(lens) if lens else 0
```
clessig: Same as above

clessig (changes requested):
Thanks for cleaning this up. It looks good, just some minor comments

clessig (approved):

PR #2391, opened 2026-05-20
Tropical cyclone casestudy

### LINKED ISSUE #2260: Tropical storm case studies

We have seen that the model can produce realistic-looking tropical cyclones, for example [Typhoon Koinu](https://en.wikipedia.org/wiki/Typhoon_Koinu) here:

Image

We would probably like to look at the model performance for one or a few selected storms in a way that is repeatable and ideally quantitative. This could take the form of an extra package within the fast evaluation, that might read some exported netcdf files and then  look at a pre-selected region and time-span known to contain an interesting storm. 

_How exactly this should be realized technically and how such case-studies should be selected is up for debate._ 

### example output
As a proof-of-concept, I have implemented a simple cyclone detector, that would allow us to look at track positions and quantities like wind-speed and core pressure along the track:

Image
Image

In this example, the model is almost perfect for Koinu, and even follows the intensification (but not the track) of [storm Bolaven](https://en.wikipedia.org/wiki/Typhoon_Bolaven_(2023)) which was initiated around step 30. Note that other examples during the same period were not as good though. 

_No response_

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation
- [ ] performance

s6sebusc:
## Description

Analyzes a single prediction for a single user-selected tropical cyclone by comparing the track location, cyclone pressure and wind speed to the target. Input is expected to be interpolated to 1°x1° and stored in netcdf files produced by the export package. A config file specifies the forecast files and the time and location at which the target cyclone is to be found. The algorithm the detects and tracks all tropical cyclones in target and prediction, selects the tracks matching the chosen storm and produces evaluation plots.

The tracking logic is later intended to be used for systematic evaluation of all tropical cyclones over a longer period of time.

## Issue Number

Closes #2260


s6sebusc:
Example output for Typhoon [Bolaven](https://en.wikipedia.org/wiki/Typhoon_Bolaven_(2023)): 

i0xr7z48_cyclone_2023-10-07T00

Snapshots of wind speed along the track, showing also the detected storm centers:

i0xr7z48_cyclone_2023-10-07T00_snapshots

clessig:
@s6sebusc : thank you very much! Could we run the tracker over the full 3 month evaluation period and compare the number of TCs to the one in IBTRACS?

s6sebusc:
> @s6sebusc : thank you very much! Could we run the tracker over the full 3 month evaluation period and compare the number of TCs to the one in IBTRACS?

yes, something like that would be a next step! If someone has data for three months already we can try it out.

clessig:
> > @s6sebusc : thank you very much! Could we run the tracker over the full 3 month evaluation period and compare the number of TCs to the one in IBTRACS?
> 
> yes, something like that would be a next step! If someone has data for three months already we can try it out.

@MatKbauer ?

packages/evaluate/src/weathergen/evaluate/example_extras/tropical_cyclones/TC_config.yml
```diff
@@ -0,0 +1,17 @@
+runid : "i0xr7z48"
+init_time :  "2023-10-07T00"
+inpath : "/p/project1/weatherai/buschow1/wegen_export/cyclones/"
```
iluise: can we merge relative paths only?

packages/evaluate/src/weathergen/evaluate/example_extras/tropical_cyclones/TC_casestudy_main.py
```diff
@@ -0,0 +1,146 @@
...
+to a user-selected storm of interest and produces diagnostic
+plots for that storm, including the track error, simulated pressure 
+and wind speed.
+
+The tracking functionality is also intended for future use in
+systematical evaluation of all tropical cyclones in the prediction. 
+
+Before running, export 10u, 10v and msl to netcdf, regridded
+to 1°x1° as follows:
+uv run export --run-id <INFERENCE_ID> --stream ERA5 \
```
iluise, line 26: what happens for smaller grids like CERRA?
Can we add an error check and/or a comment for that

packages/evaluate/src/weathergen/evaluate/example_extras/tropical_cyclones/TC_config.yml
```diff
@@ -0,0 +1,17 @@
+runid : "i0xr7z48"
+init_time :  "2023-10-07T00"
+inpath : "/p/project1/weatherai/buschow1/wegen_export/cyclones/" 
+outpath : "./plots/"
+latmin: -30 # TCs are only detected for latmin<=lat<=latmax
+latmax: 30
+selected_storm : # the storm you want to analyze
```
iluise, line 7: Can we add a README with more details on these config entries? Also for the tracking_parameters

iluise (approved):

packages/evaluate/src/weathergen/evaluate/example_extras/tropical_cyclones/TC_config.yml
s6sebusc: done, user has to insert their own path anyways

packages/evaluate/src/weathergen/evaluate/example_extras/tropical_cyclones/TC_casestudy_main.py
s6sebusc: The script runs on the output of the exporter anyways (mentioned in the new readme). If and when CERRA regridding is supported there, it will work here, until then you will get an error in the export. In principle, this implementation can also run on any other grid as long as the data is a matrix. User would typically have to adjust some parameters.

packages/evaluate/src/weathergen/evaluate/example_extras/tropical_cyclones/TC_config.yml
s6sebusc: done.

PR #2402, opened 2026-05-25
Score animations with fixed colorbar

### LINKED ISSUE #1828: Create score map animations with fixed colorbar

_No response_

_No response_

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation

SavvasMel:
## Description

1) This PR allows for score animations with fixed colorbar. See examples in the comments.

2) Some restructuring needed to take place so this PR should be reviewed carefully.

**IMPORTANT:** Since the score animations are mainly animations over forecast steps, the colorbar depicts all the accumulated error growth over those which might lead to the misconception that there is negligible error at lower number of forecast steps. To avoid this, I added the possibility of plotting a colorbar on log scale. See examples below in the comments.

Instructions:

```
global_plotting_options:
  log_colorbar: false

evaluation:
  metrics  : ["rmse", "froct", "mae"]
  regions: ["global", "nhem"]
  plot_score_maps: true
  plot_score_animations: true
```

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/1828

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
+++ b/packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
@@ -278,27 +278,25 @@ def _build_single_animation(
     stream: str,
     region: str | None,
     var: str,
-    sa: object,
+    sample: object,
     fsteps: list,
     image_format: str,
     animation_format: str,
     duration_ms: int,
     prefix: str = "map",
-    score_animation: bool = False,
 ) -> list[str]:
     """Build one animation for a single (region, sample/ens, variable) combination.
 
     All work is I/O + Pillow — no matplotlib state involved.
 
-    When ``score_animation=False`` (default) the function scans ``output_dir`` for
-    per-sample map/histogram frames whose filenames follow::
+    The function scans ``output_dir`` for per-sample map/histogram frames whose filenames follow:
 
-        {prefix}_{run_id}_{tag}_{sa}_{valid_time}_{stream}_{region}_{var}_{fstep:03d}
+        {prefix}_{run_id}_{tag}_{sample}_{valid_time}_{stream}_{region}_{var}_{fstep:03d}
 
     When ``score_animation=True`` filenames are constructed deterministically because
     the fstep is embedded in the tag (``score_maps_{metric}_fstep_{N}``) rather
     than being a zero-padded suffix.  Pass ``tag="score_maps_{metric}"`` and
-    ``sa`` as the ensemble value (or ``None`` for no ensemble).
+    ``sample`` as the ensemble value (or ``None`` for no ensemble).
 
     Returns the list of source frame paths assembled into the animation, or an
     empty list when no (or fewer than two for score maps) frames were found.
@@ -307,8 +305,8 @@ def _build_single_animation(
         return []
 
     region_part = region if region else ""
-    if sa is not None:
-        head = "_".join(filter(None, [prefix, run_id, tag, str(sa)]))
+    if sample is not None:
+        head = "_".join(filter(None, [prefix, run_id, tag, str(sample)]))
     else:
         head = "_".join(filter(None, [prefix, run_id, tag]))
     tail = "_".join(filter(None, [stream, region_part, var]))
@@ -324,8 +322,8 @@ def _build_single_animation(
     )
     if not image_paths:
         return []
-    if sa is not None:
-        anim_parts = ["animation", run_id, tag, str(sa), stream]
+    if sample is not None:
+        anim_parts = ["animation", run_id, tag, str(sample), stream]
     else:
         anim_parts = ["animation", run_id, tag, stream]
     if region:
@@ -395,7 +393,7 @@ def _dispatch_animations(
             "stream": plotter.stream,
             "region": region,
             "var": var,
-            "sa": sa,
+            "sample": sample,
             "fsteps": list(fsteps),
             "image_format": plotter.image_format,
             "animation_format": plotter.animation_format,
@@ -404,7 +402,7 @@ def _dispatch_animations(
         }
         for prefix, output_dir in prefixes
         for region in plotter.regions
-        for sa in samples
+        for sample in samples
         for var in variables
     ]
 
@@ -447,7 +445,7 @@ def _dispatch_score_map_animations(
             stream=stream,
             region=region,
             var=var,
-            sa=None,
+            sample=None,
             fsteps=list(fsteps),
             image_format=plotter_cfg["image_format"],
             animation_format=plotter_cfg["animation_format"],
--- a/packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
+++ b/packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
@@ -25,6 +25,7 @@ from weathergen.evaluate.io.data.io_orchestration import dispatch_parallel, get_
 from weathergen.evaluate.io.io_reader import Reader, ReaderOutput
 from weathergen.evaluate.plotting.bar_plots import BarPlots
 from weathergen.evaluate.plotting.line_plots import LinePlots
+from weathergen.evaluate.plotting.plot_orchestration_utils import _compute_ranges, _compute_scores
 from weathergen.evaluate.plotting.plot_utils import (
     bar_plot_metric_region,
     heat_maps_metric_region,
@@ -36,11 +37,8 @@ from weathergen.evaluate.plotting.plot_utils import (
 from weathergen.evaluate.plotting.plotter import Plotter
 from weathergen.evaluate.plotting.quantile_plots import QuantilePlots
 from weathergen.evaluate.plotting.score_cards import ScoreCards
-from weathergen.evaluate.scores.score import VerifiedData, get_score
-from weathergen.evaluate.scores.score_orchestration import get_next_fstep_data
 from weathergen.evaluate.utils.array_utils import bias_ranges, common_ranges
 from weathergen.evaluate.utils.clim_utils import get_climatology
-from weathergen.evaluate.utils.regions import RegionBoundingBox
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


SavvasMel:
animation_xr4mk0ja_score_maps_rmse_ens_0_ERA5_global_t_850_unlog
animation_xr4mk0ja_score_maps_rmse_ens_0_ERA5_global_t_850

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
```diff
@@ -49,12 +49,113 @@
...
+    Returns ``(region, fstep, score_results, preds_r, metric_names)``.
+    """
+    score_results: list[xr.DataArray | None] = [
+        get_score(score_data, m, agg_dims="sample", parameters=p)
+        for m, p in zip(metric_names, metric_params, strict=False)
+    ]
+    return region, fstep, score_results, preds_r, metric_names
+
+
+def _compute_scores_and_ranges(
```
iluise: I would separate a compute_scores function and a compute_ranges function

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
```diff
@@ -49,12 +49,113 @@
 # ---------------------------------------------------------------------------
 
 
+def _compute_scores_for_fstep(
```
iluise: it's a bit confusing that plot_orchestration starts with _compute_scores. Can we move the whole score maps code into a separate file?
Do you think this new file would better belong to scores or to plotting? I let you decide on that

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
```diff
@@ -151,28 +258,46 @@ def plot_score_maps_per_stream(
...
+        metrics=all_metrics,
+        regions=regions,
+        variables=all_variables,
+        ens_values=ens_values,
+        fsteps=fsteps,
+        n_workers=n_plot_workers,
+    )
+
 
 def _plot_score_maps_per_stream(
```
iluise, line 178: not related to your changes,  but since we are moving this anyway, can we rename this function so we don't have `plot_scores_per_stream` and `_plot_scores_per_stream` ? Thanks!

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
```diff
@@ -151,28 +258,46 @@ def plot_score_maps_per_stream(
...
+    all_metrics = list(score_ranges_dict.keys())
+    all_variables = sorted(
+        {
+            ch
+            for metric_dict in score_ranges_dict.values()
+            for region_dict in metric_dict.values()
+            for ch in region_dict
+        }
+    )
+    first_preds = next(iter(_computed.values()))[1]  # preds_r
```
iluise: first_preds is here only to get the ens_values in the line below. Can't we retrieve ens_values from preds_r or da_preds?

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
```diff
@@ -151,28 +258,46 @@ def plot_score_maps_per_stream(
...
 def _plot_score_maps_per_stream(
     plotter_cfg: dict,
+    score_ranges_dict: dict,
     output_basedir: str,
     map_dir: str,
     stream: str,
     region: str,
-    score_data: "VerifiedData",
-    metrics: dict[str, object],
+    preds: xr.DataArray,
```
iluise: Do we need preds here or we can retrieve the info from `score_results` now? 
Before we needed preds and target to compute the scores, but now we already have them computed in xarray 

It would be cleaner to propagate and get the info from what we actually plot

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
```diff
@@ -262,45 +392,67 @@ def _build_single_animation(
     animation_format: str,
     duration_ms: int,
     prefix: str = "map",
+    score_animation: bool = False,
```
iluise: can we avoid adding this additional flag and better harmonies the titles so in one case one looks for f"{score}" + XXXXX and in the other "map" + XXXXX?
If we do this, then we can remove the flag and the big if/else statement below so this function remains as general as possible!

iluise (commented):
Thanks for solving this! I added a few suggestions to make the code cleaner

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
SavvasMel: I was about to ask that yes, it made things more difficult having different name titles for score maps and other maps (preds, targets etc.).

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
SavvasMel: Done

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
SavvasMel: Done

SavvasMel:
> Thanks for solving this! I added a few suggestions to make the code cleaner

I did all the requested changes. Let me know if more is needed :)

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration_utils.py
```diff
@@ -0,0 +1,124 @@
...
+import xarray as xr
+from joblib import delayed
+
+from weathergen.evaluate.io.data.io_orchestration import dispatch_parallel
+from weathergen.evaluate.scores.score import VerifiedData, get_score
+from weathergen.evaluate.scores.score_orchestration import get_next_fstep_data
+from weathergen.evaluate.utils.regions import RegionBoundingBox
+
+
+def _compute_scores_for_fstep(
```
iluise, line 11: I find very confusing that in this way we would have a  compute_scores function in something that is related to plotting! 

The more I think about it the more I get convinced that we should create a separate file named score_maps.py with run_score_map_pipeline and all the related sub-functions: 
_compute_scores_for_fstep
_compute_scores
_compute_ranges
_scatter_plot_single
_dispatch_score_map_animations

and put it into the scores/ folder instead of plotting. What's your opinion on this?

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
iluise: do we still need the flag? I don't see it used anywhere right?

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
```diff
@@ -262,28 +284,36 @@ def _build_single_animation(
...
 ) -> list[str]:
-    """Build one GIF for a single (region, sample, variable) combination.
+    """Build one animation for a single (region, sample/ens, variable) combination.
 
     All work is I/O + Pillow — no matplotlib state involved.
 
-    Returns the list of source frame paths that were assembled into the GIF
-    (empty list if no frames were found).
+    When ``score_animation=False`` (default) the function scans ``output_dir`` for
```
iluise: remove also comment here about the flag, but add the Parameters and Return description. I am having a hard time understanding the logic of "sa" and why it is forced to None in dispatch_score_maps_animations.

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
```diff
@@ -262,28 +284,36 @@ def _build_single_animation(
...
-    #   {prefix}_{run_id}_{tag}_{sample}_{valid_time}_{stream}_{region}_{var}_{fstep:03d}
-    # For all_samples histograms, valid_time is omitted.
-    # We match files by checking a fixed prefix and suffix, allowing any
-    # valid_time (or none) in between — no glob wildcards needed.
+    if not output_dir.is_dir():
+        return []
+
     region_part = region if region else ""
-    head = "_".join(filter(None, [prefix, run_id, tag, str(sa)]))
+    if sa is not None:
```
iluise: Can you explain me what is "sa" and why we need this if/else for the animations?

iluise:
Thanks for addressing the comments! The more I think about it the more I am getting convinced that we should move all the score_maps functions into a separate class/file (and ito scores/), because it is counter intuitive to compute scores inside the plotting orchestration. opinions/suggestions are more than welcome!

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
SavvasMel: So "sa" is for samples. In the score map case we do not have samples as in the preds or targets. So this needs to be None for the score maps dispatch. The part related to the "ensemble" part is hidden in the "tag" variable.

SavvasMel:
> Thanks for addressing the comments! The more I think about it the more I am getting convinced that we should move all the score_maps functions into a separate class/file (and ito scores/), because it is counter intuitive to compute scores inside the plotting orchestration. opinions/suggestions are more than welcome!

In general I believe the code needs restructuring. I find it bizarre that we calculate scores twice one before the plotting and one during the plotting of the score maps. Also we need to find a better way to "extract" information like metrics, samples, regions in a more unified way. Every time I am searching which python variable has this info :/

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration_utils.py
SavvasMel: Yeah no worries, I was expecting that it would not be preferable. It was more like draft that to check that the transferring of the functions to a separate file works as expected.

 I agree with your point though not 100% if this is the best restructuring the one that you suggest. I need some time to think.

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
SavvasMel: I clarified now in a new commit

iluise (approved):
We will merge like this and restructure the code in a second PR. The related Issue will come soon.

PR #2482, opened 2026-06-10
added cosine matching loss

### LINKED ISSUE #2481: Task: Introduce cosine matching loss

Upon initial experiments, we found the latents of the consecutive tokens have a cosine distance of 0.75.
Hence, we added a cosine loss to the consecutive forecasting latents.

Upon  initial experiments, we show latents are more active and froct is not degrading.

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation
- [ ] performance

ankitpatnala:
## Description

Added cosine matching loss which is used for constraining cosine similarity between consecutive forecasting latents

## Issue Number

Link the Issue number this change addresses:
Closes #2481


ankitpatnala:
@SavvasMel

clessig (changes requested):
Thanks for the implementation. Can we generalise the PR slightly to have a LossLatent (I think it's a regularisation term since it does not involve targets but this is a minor point), with the cosine being a specific implementation (RMSE or other losses are possible as well). We will also need to best structure the part in the model but let's first to this restructuring.

@MatKbauer : (I don't think so since this uses that target_aux_calculator but double-checking)

@sophie-xhonneux : for me this is more a regularisation term since no target are involved. Any opinion on this?

ankitpatnala:
> Thanks for the implementation. Can we generalise the PR slightly to have a LossLatent (I think it's a regularisation term since it does not involve targets but this is a minor point), with the cosine being a specific implementation (RMSE or other losses are possible as well). We will also need to best structure the part in the model but let's first to this restructuring.
> 
> @MatKbauer : (I don't think so since this uses that target_aux_calculator but double-checking)
> 
> @sophie-xhonneux : for me this is more a regularisation term since no target are involved. Any opinion on this?

Thanks for the comments! I have implemented few changes you proposed. Now I have removed targets and changed the naming to LatentLoss. 
With regards to restructuring, I depopulated the forward function by moving the cosine calculation operation to `loss/utils.py`

PR #2486, opened 2026-06-11
Implement Dynamic Channel-Weighting MSE loss

### LINKED ISSUE #2485: Implement dynamically weighted MSE loss function

In Samudra 2 an improved MSE loss was introduced, using EMA to calculate channel weights dynamically during training. In their case it showed significant improvement in case of channels with low temporal variability. 

This loss function is easy to implement and worth having as an option 

https://arxiv.org/abs/2606.02610

### Area

- [x] model
- [ ] science
- [ ] documentation
- [ ] performance

kacpnowak:
## Description

Implements the Dynamic Channel-Weighting Loss Function from the Samudra 2 paper to prevent variance collapse and imprinting artifacts in slow-evolving fields (like the deep ocean).

  Enable it inside  loss_fcts :
```yaml
      losses: {
        "physical": {
            type: LossPhysical,
            loss_fcts: {
                "mse": { },
                "dynamic_loss": { window: 100, L: 20.0 }
            },
        }
      }
```

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/2485


sophie-xhonneux (approved):
Looks good to me, shouldn't affect anybody if they are not using and backwards compatible as far as I can tell. @clessig your final call

clessig (commented):
Thanks, interesting to see how this works for non-ocean but slowly evolving fields, e.g. soil moistures. 

The EMA functionality should be separated out into separate functions or a small class before merging.

clessig (changes requested):

kacpnowak:
@clessig thank your for the review. I've made the requested changes

kacpnowak:
Before and after comparison is quite spectacular with 30 day forecast 
---
Before 

image

---

After
image

clessig:
> ## Before and after comparison is quite spectacular with 30 day forecast
> Before
> 
> image
> After image

@kacpnowak : What does the grey area mean in the plot?

kacpnowak:
Grey is ~7.5 degree Celsius. It's just so over smoothing that it predicts mean value almost everywhere

src/weathergen/train/loss_modules/loss_module_physical.py
```diff
@@ -31,6 +31,55 @@ def get_num_samples(config) -> np.typing.NDArray:
...
+            weights_channels = weights_channels * weights_channels_static
+
+        return weights_channels
+
+    def update(self, stream_name: str, loss_lfct_chs: torch.Tensor):
+        if not self.enabled:
+            return
+
+        with torch.no_grad():
+            mse_per_chan = loss_lfct_chs.detach().clamp(min=1e-8)
```
clessig: 1e-8 is order machine precision in fp32 (which is the highest precision we possibly have here). Shouldn't we clamp to a larger value?

src/weathergen/train/loss_modules/loss_module_physical.py
```diff
@@ -57,6 +106,9 @@ def __init__(
         self.device = device
         self.name = "LossPhysical"
 
+        # Dynamic Loss state (extract it before parsing the actual loss functions)
+        self.dynamic_loss_cfg = loss_fcts.pop("dynamic_loss", None)
```
clessig: Why do we need to pop here?

src/weathergen/train/loss_modules/loss_module_physical.py
```diff
@@ -220,7 +278,18 @@ def compute_loss(self, preds: dict, targets: dict, metadata) -> LossValues:
 
             losses_all[stream_name] = defaultdict(dict)
 
-            stream_loss_weight, weights_channels = self._get_weights(stream_info)
+            stream_loss_weight, weights_channels_static = self._get_weights(stream_info)
+
+            if self.dynamic_loss_ema.enabled:
```
clessig: Can we encapsulate this block also in `_get_weights()`

clessig (commented):
Some minor comments.

src/weathergen/train/loss_modules/loss_module_physical.py
kacpnowak: I just set at lowest possible value with a plan to increase if loss gets unstable, but no such thing happened. Maybe it could be a configurable parameter?

src/weathergen/train/loss_modules/loss_module_physical.py
kacpnowak: Done

src/weathergen/train/loss_modules/loss_module_physical.py
kacpnowak: If was removed so it doesn't mess with other loss functions, but I changed it so it's `get` and it's checked later

clessig (approved):
Thanks.

PR #2231, opened 2026-04-17
[2186] Scaling utils

florianscheidl:
## Description

Contains utility we used for scaling experiments, namely to capture elapsed time metrics in the trainer and run_train.
Small code modifications became necessary as we trained on more than 16 nodes. Specifically:
- Lower bound on warmup, cooldown, and decay steps in lr_scheduler
- Lower bound on beta2 used in Adam optimizer (negative values are illegal). Set to 0.9 now, based on Internet suggestion, but open to suggestions.
- Lower bound on len_per_rank when setting mini_epoch_base (avoiding zero-division error for small configs).

The data extraction and plot generation script are addressed in https://gitlab.jsc.fz-juelich.de/esde/WeatherGenerator-private/-/merge_requests/180.

Example plots and tables: 

jupiter_weak_scaling_table
jupiter_weak_scaling

## Issue Number

Fixes #2186

## Checklist before asking for review

-   [x] I have performed a self-review of my code
-   [x] My changes comply with basic sanity checks:
      - I have fixed formatting issues with `./scripts/actions.sh lint`
      - I have run unit tests with `./scripts/actions.sh unit-test`
      - I have documented my code and I have updated the docstrings.
      - I have added unit tests, if relevant
-   [x] I have tried my changes with data and code:
      - I have run the integration tests with `./scripts/actions.sh integration-test`
      - (bigger changes) I have run a full training and I have written in the comment the run_id(s): `launch-slurm.py --time 60`
      - (bigger changes and experiments) I have shared a hegdedoc in the github issue with all the configurations and runs for this experiments
-   [x] I have informed and aligned with people impacted by my change:
    - for config changes: the MatterMost channels and/or a design doc
    - for changes of dependencies: the MatterMost software development channel

PR #2501, opened 2026-06-12
Plot losses against elapsed training time via --x-axis flag

florianscheidl:
## Summary

Adds `--x-axis` support to `plot_train` so losses can be plotted against `elapsed_training_time_seconds` instead of samples.

## Changes

- **CLI**: Replace `--x_type` ('step'/'reltime') with `--x-axis` accepting column names `samples` (default) or `elapsed_training_time`
- **plot_loss_avg**: Added `x_axis` param (previously hardcoded `num_samples`); xlabel is auto-derived
- **Friendly labels**: When plotting against `elapsed_training_time_seconds`, x-axis labels read "elapsed training time [s]" instead of the raw column name
- All four plotting functions (`plot_lr`, `plot_loss_avg`, `plot_loss_per_stream`, `plot_loss_per_run`) now respect the `x_axis` param
- Removed dead `x_type` parameter from `plot_loss_per_stream`

## Usage

```bash
# Plot against samples (default)
python -m weathergen.utils.plot_training -fy runs.yml -o ./plots

# Plot against elapsed training time
python -m weathergen.utils.plot_training -fy runs.yml -o ./plots --x-axis elapsed_training_time
```


src/weathergen/run_train.py
```diff
@@ -188,7 +189,7 @@ def run_train(args):
     trainer = Trainer(cf.train_logging)
 
     try:
-        trainer.run(cf, devices)
+        trainer.run(cf, devices, t_start=t_start)
```
clessig: Can't we move this to Trainer?

config/streams/era5_georing/era5.yml
```diff
@@ -0,0 +1,36 @@
+# (C) Copyright 2024 WeatherGenerator contributors.
```
clessig: Remove from PR

config/streams/era5_georing/geos.yml
```diff
@@ -0,0 +1,77 @@
+
```
clessig: Remove from PR

config/config_era5_georing.yml
```diff
@@ -0,0 +1,289 @@
+# (C) Copyright 2025 WeatherGenerator contributors.
```
clessig: Remove from PR

src/weathergen/train/trainer.py
```diff
@@ -378,7 +381,14 @@ def run(self, cf, devices, run_id_contd=None, mini_epoch_contd=None):
...
         self.validate_before_training()
 
+        # Log startup time
+        if is_root() and t_start is not None:
+            startup_time = time.time() - t_start
+            self.train_logger.log_metrics("train", {"startup_time_seconds": startup_time})
+            logger.info(f"Startup time: {startup_time:.2f} seconds")
+
         # training loop
+        self.t_training_start = time.time()
```
clessig, line 383: For plot_train, timing should start here. This would also avoid that run_train is modified.

src/weathergen/train/trainer.py
```diff
@@ -378,7 +381,14 @@ def run(self, cf, devices, run_id_contd=None, mini_epoch_contd=None):
         # run validation before training if requested
         self.validate_before_training()
 
+        # Log startup time
+        if is_root() and t_start is not None:
```
clessig: I don't think this is needed.

src/weathergen/train/trainer.py
```diff
@@ -561,6 +571,19 @@ def train(self, mini_epoch):
 
         self.dataset.advance()
 
+        if is_root():
+            total_training_time = time.time() - self.t_training_start
+            self.train_logger.log_metrics(
+                "train",
+                {
+                    "completed_mini_epoch": mini_epoch,
+                    "training_time_after_mini_epoch_seconds": total_training_time,
```
clessig: Do we use this explicitly? Otherwise I would remove it. It can always be recovered by the time elapsed up to step k and the samples_per_mini_epoch (which is available through the config).

clessig (commented):
Thanks for picking this up. It's important to keep Trainer as clean as possible so I would introduce only what is really necessary.

florianscheidl:
> Thanks for picking this up. It's important to keep Trainer as clean as possible so I would introduce only what is really necessary.

Sorry, this should not have been a PR, only a commit. The correct PR is in https://github.com/ecmwf/WeatherGenerator/pull/2231

clessig:
> > Thanks for picking this up. It's important to keep Trainer as clean as possible so I would introduce only what is really necessary.
> 
> Sorry, this should not have been a PR, only a commit. The correct PR is in #2231

But we should have plot_train with elapsed time on the x-axis and it should be a separate PR.

PR #2503, opened 2026-06-14
[2502][evaluation] score bug fixes

### LINKED ISSUE #2502: Ensemble wide scores currently disabled
### What happened?

Currently, the code in `evaluate/scores/score.py` completely skips any ensemble wide score computations (`return None` [here](https://github.com/ecmwf/WeatherGenerator/blob/ee193c2f71b766f3dab71b21a068e4bcc9aa93c2/packages/evaluate/src/weathergen/evaluate/scores/score.py#L263)).

Furthermore, even when this is fixed, there are two small issues with the score plotting/computation:

1. Plots are produced for each ensemble member, even if the metric is identical across members (e.g. CRPS)
2. SSR uses the member RMSE in the denominator, but the ensemble spread in the enumerator. I believe it should be ensemble RMSE in the denominator to match the literature (e.g. GenCast).
3. There is also a small bug if only ensemble metrics are requested:
`xarray.core.coordinates.CoordinateValidationError: coordinate ens has dimensions ('ens',), but these are not a subset of the DataArray dimensions ('metric', 'ipoint', 'channel')`

Try `uv run evaluate --config [...]` with a config that has `metrics  : ["crps"]` and a run id that has ensembles (e.g. `qvim6zb3` on Santis)

_No response_

moritzhauschulz:
## Description

Enables the previously dead probabilistic metrics (spread, ssr, crps, rank_histogram) in the evaluation package, corrects the spread-skill ratio to the standard ensemble-mean definition, and makes the spatial score-map path robust for metrics that collapse the ensemble dimension. This unblocks GenCast-style spread-skill diagnostics over lead time and ensemble-spread maps.

**I am not very familiar with the eval pipeline, so it would be good if this could be reviewed by someone more knowledgeable**

### `scores/score.py`
- **Enable probabilistic-metric dispatch.** Replace the dead `assert self.ens_dim … / return None`
  (undefined `self.ens_dim`, unconditional return) with a real check: warn and skip when the
  ensemble dim `self._ens_dim` is absent from the predictions (e.g. deterministic runs), otherwise
  dispatch to the metric function. This activates `spread`, `ssr`, `crps`, and `rank_histogram`.
- **Correct the spread-skill ratio.** `calc_ssr` now divides the ensemble spread by the **RMSE of
  the ensemble mean** (the "skill", GenCast / WeatherBench2 convention) —
  `calc_spread(p) / calc_rmse(p.mean("ens"), gt)` — instead of the full-ensemble per-member RMSE.
  SSR is now a single value per variable/level/lead-time with the standard calibration
  interpretation (under-/over-dispersion), consistent with the already ensemble-reduced spread
  numerator.

### `plotting/plot_orchestration.py` (`_plot_score_maps_per_stream`)
- **Fix `CoordinateValidationError` crash.** Guard the ensemble-label assignment on
  `"ens" in plot_metrics.dims` (the concatenated result) rather than `preds.dims`. When every
  selected metric reduces the ensemble dim (e.g. `metrics: ["ssr"]`), `plot_metrics` has no `ens`
  dim and the previous unconditional `plot_metrics["ens"] = preds.ens` raised.
- **Avoid redundant per-member maps.** Track which metrics retain a per-member `ens` dim in their
  own result (`ens_metrics`). Iterate ensemble members only for those; ensemble-reduced metrics
  (`spread`, `crps`, `ssr`) get a single map instead of one identical map per member.
- **Collapse broadcast ensemble axis.** When `xr.concat` broadcasts a reduced metric across `ens`,
  select a single member (`isel(ens=0, drop=True)`) so the plotted field is 2-D.

## Issue Number

Closes #2502 

-   [] I have tried my changes with data and code:

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
+++ b/packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
@@ -358,6 +358,8 @@ def _plot_score_maps_per_stream(
     if not valid:
         return
 
+    ens_metrics = {m for m, r in valid if "ens" in r.dims}
+
     plot_metrics = xr.concat(
         [r for _, r in valid],
         dim="metric",
@@ -370,7 +372,7 @@ def _plot_score_maps_per_stream(
         metric=[m for m, _ in valid],
     ).compute()
 
-    if "ens" in preds.dims:
+    if "ens" in plot_metrics.dims:
         plot_metrics["ens"] = preds.ens
 
     has_ens = "ens" in plot_metrics.coords
@@ -378,13 +380,17 @@ def _plot_score_maps_per_stream(
 
     plot_tasks: list[dict] = []
     for metric in plot_metrics.coords["metric"].values:
-        for ens_val in ens_values:
+        metric_ens_values = ens_values if str(metric) in ens_metrics else [None]
+        for ens_val in metric_ens_values:
             tag = "score_maps" + (f"_ens_{ens_val}" if ens_val is not None else "") + f"_{metric}"
             for channel in plot_metrics.coords["channel"].values:
                 sel = {"metric": metric, "channel": channel}
                 if ens_val is not None:
                     sel["ens"] = ens_val
-                data = plot_metrics.sel(**sel).squeeze()
+                data = plot_metrics.sel(**sel)
+                if ens_val is None and "ens" in data.dims:
+                    data = data.isel(ens=0, drop=True)
+                data = data.squeeze()
                 title = f"{metric} - {channel}: fstep {fstep}" + (
                     f", ens {ens_val}" if ens_val is not None else ""
                 )
--- a/packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
+++ b/packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
@@ -373,15 +373,15 @@ def _plot_score_maps_per_stream(
     ).compute()
 
     if "ens" in plot_metrics.dims:
-        plot_metrics["ens"] = preds.ens
+        plot_metrics = plot_metrics.assign_coords(ens=preds.ens.values)
 
-    has_ens = "ens" in plot_metrics.coords
-    ens_values = plot_metrics.coords["ens"].values if has_ens else [None]
+    all_ens = plot_metrics.coords["ens"].values if "ens" in plot_metrics.dims else [None]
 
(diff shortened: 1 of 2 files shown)


clessig:
@jpolz : could you have a look?

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
```diff
@@ -370,21 +372,25 @@ def _plot_score_maps_per_stream(
         metric=[m for m, _ in valid],
     ).compute()
 
-    if "ens" in preds.dims:
-        plot_metrics["ens"] = preds.ens
+    if "ens" in plot_metrics.dims:
```
iluise, line 375: why this is necessary? shouldn't already be the case that the ens has the same coords of preds (since it comes from there?)

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
```diff
@@ -358,6 +358,8 @@ def _plot_score_maps_per_stream(
     if not valid:
         return
 
+    ens_metrics = {m for m, r in valid if "ens" in r.dims}
```
iluise: can we use more descriptive names for m and r?

iluise (approved):
Hi Mortiz, thanks for working on this! I agree that ssr should be as you said. 
I added two style comments to the PR. It looks good otherwise.

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
moritzhauschulz: I think this is necessary for the case where all of the metrics collapse the ensemble dimension (e.g. only plotting CRPS). Then `preds` might still have ensembles, but `plot_metrics` won't. Adding `plot_metrics["ens"] = preds.ens` will then error. 

But full disclosure, I did use Claude for this and might not be familiar with all the intricacies. I reviewed this again today and am pretty confident this makes sense. The old version did cause a bug for me (for the case described above) which is now gone.

packages/evaluate/src/weathergen/evaluate/plotting/plot_orchestration.py
moritzhauschulz: Changed this to `metric`, `result` throughout.

iluise (approved):
Tested and it works. Approving it

PR #2508, opened 2026-06-15
[2507][Evaluation] Implement Adjusted SSR

### LINKED ISSUE #2507: GenCast Style Corrected SSR

[Gencast](https://arxiv.org/pdf/2312.15796) implements SSR slightly different than the current implementation. Both the spread and the SSR (using spread) have a corrective factor that ensures unbiasedness, so that the optimal SSR is ~1. I think this should be implemented for consistency – and it's just nice that the optimal SSR is 1. I may be convinced otherwise. I propose to implement this as 'adjusted SSR'.

Image 

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation
- [ ] performance

moritzhauschulz:
## Description

Adds two probabilistic metrics following GenCast (Price et al., 2024, App. A),
leaving the existing `spread`/`ssr` untouched.

**UPDATE: Also includes fixes to https://github.com/ecmwf/WeatherGenerator/issues/2577 as provided by this [PR](https://github.com/moritzhauschulz/WeatherGenerator/pull/4)**

## Changes
- **`spread_adj`** — unbiased ensemble spread `sqrt(mean Var_ens(ddof=1))` (GenCast Eq. A.6).
- **`ssr_adj`** — adjusted spread-skill ratio `sqrt((M+1)/M) · spread_adj / RMSE(ens_mean)`
  (GenCast Eq. A.9), where `M` is the ensemble size. The `sqrt((M+1)/M)` factor removes the
  finite-ensemble bias so a perfectly calibrated ensemble gives **SSR = 1**.
- Plotting: `ssr_adj` line plots draw a horizontal reference line at the optimal value of 1.

## Notes
- Legacy `spread` (biased `ddof=0`) and `ssr` are unchanged; GenCast has no uncorrected SSR,
  so these remain only as non-standard diagnostics.
- Use via `evaluation.metrics` in the eval config (e.g. add `ssr_adj`, `spread_adj`).

**Should be reviewed by someone from the evaluation team**

On Santis, run with `uv run evaluate --config ./config/evaluate/eval_config_test.yml -run-ids qvim6zb3` using the attached eval config. 

## Issue Number

Closes #2507 

Note this depends on #2503, which should be merged first.

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/scores/score.py
+++ b/packages/evaluate/src/weathergen/evaluate/scores/score.py
@@ -1505,7 +1505,10 @@ class Scores:
         self,
         p: xr.DataArray,
         gt: xr.DataArray,
-        latitude_weights: xr.DataArray | None = None,
+        use_latitude_weights: bool = False,
+        lat_coord_name: str | None = None,
+        min_lat_weight: float = 1e-3,
+        max_lat_weight: float = 1.0,
     ) -> xr.DataArray:
         """
         Calculate the Spread-Skill Ratio (SSR) of the forecast ensemble data w.r.t. reference data
@@ -1516,9 +1519,16 @@ class Scores:
             Forecast data array with ensemble dimension
         gt: xr.DataArray
             Ground truth data array
-        latitude_weights: xr.DataArray | None
-            Optional latitude weights for area-weighted averaging.
-            If provided, both spread and RMSE (skill) will be weighted independently by these values.
+        use_latitude_weights: bool
+            If True, automatically calculate and apply latitude weights to both spread and RMSE
+            (skill) components independently. Default is False.
+        lat_coord_name: str | None
+            Name of the latitude coordinate. If None and use_latitude_weights is True, will search
+            for standard latitude coordinate names ('lat', 'latitude', 'rlat').
+        min_lat_weight: float
+            Minimum latitude weight value (at poles). Default is 1e-3.
+        max_lat_weight: float
+            Maximum latitude weight value (at equator). Default is 1.0.
             
         Returns
         -------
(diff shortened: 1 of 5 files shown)


packages/evaluate/src/weathergen/evaluate/scores/score.py
```diff
@@ -1402,6 +1407,33 @@ def calc_spread(self, p: xr.DataArray, **kwargs) -> xr.DataArray:
...
+        ----------
+        p: xr.DataArray
+            Forecast data array with ensemble dimension
+
+        Returns
+        -------
+        xr.DataArray
+            Unbiased ensemble spread (GenCast convention)
+        """
+        ens_var = p.var(dim=self._ens_dim, ddof=1)
```
iluise: the current `calc_spread` is calculated in this way:

```
ens_std = p.std(dim=self._ens_dim)
return self._mean(np.sqrt(ens_std**2))
```

which is different from `sqrt(mean(var))` that you are calculating. Investigating inside the ECMWF documentation, it seems that in Quaver the spread is computed as follows:

Image

so, IIUC, as you are suggesting to do as well. Can you please harmonize `calc_spread` accordingly (and as I said above ideally merge the two functions adj and non adj into one)?

iluise (commented):
Hi Mortiz, 
Thanks a lot for looking into this!

General comment: why do we need separate metrics? can't we modify the existing `ssr` and `spread` with a flag for the adjusted case?

I checked and the current `calc_spread` seems computed in a convoluted way with `mean(sqrt(std**2))`. We should definitely harmonize the spread computation but I would like to hear @jpolz and @SavvasMel opinions on this. Which one is correct for you?

packages/evaluate/src/weathergen/evaluate/scores/score.py
jpolz: I also found both ways in different packages. @MatKbauer mentioned `spread = sqrt(ens_var)` after internal discussions.

jpolz (approved):
I implemented the lat weighting and unified ssr (tested). I also reviewed and tested the ref line which has worked for ssr. Can be merged IMO

iluise (approved):

PR #2574, opened 2026-06-30
[2558] Make DataReaderGREP compliant with new datasets

### LINKED ISSUE #2558: [Readers Extra] Changes in DataReaderGREP

Modify the DataReaderGREP in order to make it compatible with two new datasets:

- CGLORSv8 ice (cice_cglorsv8_1993_2023.zarr)
- CGLORSv8 nemo (nemo_cglorsv8_subset_1993_2023.zarr)

Changes needed:
- Opening of datasets in zarr version 3
- Different coordinates name
- Masking of land points

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation
- [ ] performance

pierluigicosi:
## Description

Changes:

- Opening of datasets in zarr version 3
- Added different coordinates name (TLAT, TLON)
- Added different time coordinate name (time_counter)
- Handle masking of land points in different ways

CC @simone99n 

## Issue number
Closes #2558 

### DIFF
--- a/packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
+++ b/packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
@@ -15,6 +15,16 @@ import numpy as np
 import xarray as xr
 from numpy.typing import NDArray
 
+# Dask is optional: if it is not installed we transparently fall back to the
+# original eager (non-chunked) behaviour, so functionality never depends on it.
+try:
+    import dask.array as dask_array
+
+    _HAS_DASK = True
+except ImportError:  # pragma: no cover - dask is an optional speed-up dependency
+    dask_array = None
+    _HAS_DASK = False
+
 from weathergen.datasets.data_reader_base import (  # type: ignore
     DataReaderTimestep,
     ReaderData,
@@ -33,6 +43,14 @@ class DataReaderGREP(DataReaderTimestep):
     This reader handles datasets stored as Zarr with dimensions (time, latitude, longitude)
         beta: handling  dimensions (time_centered, nav_lat, nav_lon)
     Converts the gridded data to ReaderData format.
+
+    PERFORMANCE NOTES
+    ------------------
+    This version batches all (channel, timestep) reads for a given window into a single
+    xarray/Dask read instead of issuing one Zarr read per channel per timestep. When Dask
+    is available, the dataset is opened with chunking enabled so that the underlying Zarr
+    chunks can be fetched/decompressed concurrently (multi-threaded) via Dask's scheduler.
+    None of this changes the values, shapes, dtypes or ordering returned by `_get`.
     """
 
     def __init__(
@@ -73,8 +91,14 @@ class DataReaderGREP(DataReaderTimestep):
         self.n_lon: int = 0
         self.n_points: int = 0
 
+        # Cached time coordinate/dimension info (populated during lazy init, reused by _get
+        # on every call instead of being re-resolved each time).
+        self._time_coord_name: str | None = None
+        self._time_dim_name: str | None = None
+        self.time_coord: NDArray | None = None
+
         # debug
-        self.log_debug = False
+        self.log_debug = True
 
         # Set empty defaults so the object is always in a valid state
         self.init_empty()
@@ -97,6 +121,9 @@ class DataReaderGREP(DataReaderTimestep):
         else:
             raise ValueError(f"Cannot determine Zarr format for {self._filename}")
 
+        # Apriamo con chunking Dask invece di chunks=None.
+        use_dask = self._stream_info.get("use_dask", True) and _HAS_DASK
+
         ds = xr.open_zarr(
             self._filename,
             consolidated=True,
@@ -104,14 +131,30 @@ class DataReaderGREP(DataReaderTimestep):
             **kwargs,
         )
 
+        if use_dask:
+            _logger.info(
+                f"{self._stream_info.get('name', '?')}: dataset opened with Dask."
+            )
+        else:
+            _logger.info(
+                f"{self._stream_info.get('name', '?')}: dataset opened without Dask chunking "
+                "(eager reads, same as original behaviour)."
+            )
+
         # ---- Time axis -------------------------------------------------------
 
         for coord_name in ["time", "time_centered", "time_counter"]:
             if coord_name in ds.coords:
                 time_coord: NDArray = ds.coords[coord_name].values
+                self._time_coord_name = coord_name
                 break
         else:
             raise KeyError("Nessuna coordinata temporale trovata")
+
+        # Rechunk with 1 along time dimension cause we read one timestep at a time in _get() (or a small batch of timesteps),
+        # and we want to avoid reading multiple timesteps from the same chunk.
+        if use_dask:
+            ds=ds.chunk({self._time_coord_name: 1})  
 
         data_start_time = np.datetime64(time_coord[0])
         data_end_time = np.datetime64(time_coord[-1])
@@ -131,12 +174,10 @@ class DataReaderGREP(DataReaderTimestep):
             # The base module exposes no str_to_timedelta, so we parse manually.
             period = _str_to_timedelta(self._stream_info["frequency"])
 
-        # Patch the instance attributes that DataReaderTimestep._get_dataset_idxs reads.
-        # This is equivalent to what super().__init__(..., data_start_time, data_end_time, period)
-        # would set, but without the side-effects of calling __init__ a second time.
         self.data_start_time = data_start_time
         self.data_end_time = data_end_time
         self.period = period
+        self.time_coord = time_coord
 
         # ---- Spatial grid ----------------------------------------------------
         grid_defs = [
@@ -401,20 +442,11 @@ class DataReaderGREP(DataReaderTimestep):
 
         # Map channel indices → variable names
         selected_channels = [self.available_vars[i] for i in channels_idx]
-        # selected_channels = self.available_vars
 
-        data_arrays: list[NDArray] = []
-        datetimes_list: list[np.datetime64] = []
-
-        for coord in ["time", "time_centered", "time_counter"]:
-            if coord in self.ds.coords:
-                time_values = self.ds.coords[coord].values
-                break
-        else:
-            raise KeyError(
-                f"No time coordinate found. Available coordinates: {list(self.ds.coords)}"
-            )
-
+        # Time coord/dim were resolved once in _lazy_init; reuse them instead of
+        # re-searching ds.coords / ds.dims on every call.
+        time_values = self.time_coord
+        time_dim = self._time_coord_name
         n_time = len(time_values)
 
         if self.log_debug:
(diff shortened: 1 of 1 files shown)


simone99n:
Just to keep track, can you point out the performance issue in the DataReaderGREP?

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -89,21 +94,33 @@ def _lazy_init(self) -> None:
...
+                ds: xr.Dataset = xr.open_zarr(
+                    self._filename,
+                    consolidated=True,
+                    chunks=None,
+                )
         except Exception as e:
             name = self._stream_info["name"]
             _logger.error(f"Failed to open {name} at {self._filename}: {e}")
-            return  # leave in empty state
+            return
```
simone99n, line 133: Instead of adding another `try/except` block (**TODO**: we should remove the existing one), we can determine the `zarr_format` by inspecting the metadata files:

- If `.zattrs`, `.zarray`, or `.zgroup` are present, use `zarr_format=2`.
- Otherwise, if `zarr.json` is present, use `zarr_format=3`.

This allows us to call xr.open_zarr() only once:
```
ds: xr.Dataset = xr.open_zarr(
    self._filename,
    consolidated=True,
    chunks=None,
    zarr_format=zarr_format,
)
```

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -89,21 +94,33 @@ def _lazy_init(self) -> None:
...
-        try:
-            time_coord: NDArray = ds.coords["time"].values
-        except KeyError:
-            time_coord: NDArray = ds.coords["time_centered"].values
+        for coord_name in ["time", "time_centered", "time_counter"]:
+            if coord_name in ds.coords:
+                time_coord: NDArray = ds.coords[coord_name].values
+                break
+        else:
+            raise KeyError("Nessuna coordinata temporale trovata")
```
simone99n, line 153: Great!
Just remove the #TODO comment

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -177,14 +194,37 @@ def _lazy_init(self) -> None:
...
+            self._nav_lat_flat = nav_lat.flatten()  # (n_points,)
+            self._nav_lon_flat = nav_lon.flatten()  # (n_points,)
+
+            self.n_lat, self.n_lon = nav_lat.shape
+            self.n_points = self.n_lat * self.n_lon
+
+            # Provide sorted unique 1-D views for callers that inspect .latitudes/.longitudes
+            self.latitudes = np.unique(self._nav_lat_flat)
+            self.longitudes = np.unique(self._nav_lon_flat)
+
```
simone99n: This code is duplicated (see the nav_lat/nav_lon and `latitide`/`longitude` case)

You can use the same technique you used for the time dimension:
```
for coord_name in ["time", "time_centered", "time_counter"]:
    if coord_name in ds.coords:
        time_coord: NDArray = ds.coords[coord_name].values
        break
else:
    raise KeyError("Nessuna coordinata temporale trovata")
```

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -358,12 +398,14 @@ def _get(self, idx: TIndex, channels_idx: list[int]) -> ReaderData:
...
 
         # TODO remove try/except
-        try:
-            # EOBS uses 'time'
-            time_values = self.ds.coords["time"].values
-        except KeyError:
-            # C-GLORS uses 'time_centered'
-            time_values = self.ds.coords["time_centered"].values
+        for coord in ["time", "time_centered", "time_counter"]:
+            if coord in self.ds.coords:
```
simone99n: remove TODO comment

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -382,25 +424,22 @@ def _get(self, idx: TIndex, channels_idx: list[int]) -> ReaderData:
...
-                    [
-                        self.ds[ch]
-                        .isel(time_centered=int(t_idx))
-                        .values.astype(np.float32)
-                        .flatten()
-                        for ch in selected_channels
-                    ],
-                    axis=1,
-                )
+            for time_dim in ["time", "time_centered", "time_counter"]:
```
simone99n: Remove TODO comment

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -425,9 +464,20 @@ def _get(self, idx: TIndex, channels_idx: list[int]) -> ReaderData:
...
+            elif "nav_lat" in self.ds.coords and "nav_lon" in self.ds.coords:
+                # Maschera i punti dove nav_lat == 0 AND nav_lon == 0
+                masked = (self._nav_lat_flat == 0.0) & (self._nav_lon_flat == 0.0)
+                coords_single[masked] = np.nan
+            elif "TLAT" in self.ds.coords and "TLON" in self.ds.coords:
+                # Maschera i punti di terra
+                data_mask=self.ds[selected_channels[0]].isel(time=0)
+                masked = np.isnan(data_mask).values.flatten()
+                coords_single[masked] = np.nan
         else:
```
simone99n: remove duplicate `coords_single[masked] = np.nan`

simone99n (changes requested):
I just left few comments.
We need also to investigate the performance issue

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -93,28 +88,24 @@ def _lazy_init(self) -> None:
...
-            return
+        kwargs = {}
+        store = Path(self._filename)
+        if (store / "zarr.json").exists():
+            kwargs = {"zarr_format": 3}
+        elif (store / ".zgroup").exists():
+            kwargs = {"zarr_format": 2}
+        else:
+            raise ValueError(f"Cannot determine Zarr format for {self._filename}")
+
```
simone99n, line 123: This works great, but for safety, let's add two more `elif` branches to also check for `.zattrs` and `.zarray`, and assign `kwargs = {"zarr_format": 2}` in those cases as well.

simone99n (changes requested):

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
simone99n, line 111: Possiamo spostare il contenuto di `_lazy_init()` nell'`__init__()` (ormai non è più lazy)

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -269,6 +341,24 @@ def _select_channels(
...
+        def depth_key(name, no_depth_position='end'):
+            parts = name.rsplit('_', 1)
+            if len(parts) == 2:
+                var, maybe_depth = parts
+                try:
+                    depth = float(maybe_depth)
+                    return (0, depth, var)  # 0 = "ha profondità", ordina normalmente
+                except ValueError:
+                    pass
+            # nessuna profondità valida trovata
```
simone99n, line 355: Nell'except aggiungiamo un messaggio di warning/info

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -269,6 +341,24 @@ def _select_channels(
...
             selected_idxs.append(i)
+        
+        # --- ordina per profondità (e a parità di profondità per nome variabile) ---
+        def depth_key(name, no_depth_position='end'):
+            parts = name.rsplit('_', 1)
+            if len(parts) == 2:
+                var, maybe_depth = parts
+                try:
+                    depth = float(maybe_depth)
+                    return (0, depth, var)  # 0 = "ha profondità", ordina normalmente
```
simone99n, line 352: Rely only on the suffix `_XXX` can be dangerous (if we, for example,  have ensemble_1, ensemble_2, ...)
My suggestion is to have a kind of identifier --> eg. temperature_depth_100 --> *_depth_* is the identifier

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -375,73 +457,63 @@ def _get(self, idx: TIndex, channels_idx: list[int]) -> ReaderData:
...
+
+        for batch_start in range(0, n_valid_timesteps, batch_size):
+            batch_t_idxs = valid_t_idxs[batch_start : batch_start + batch_size]
+            ds_batch = ds_sub.isel({time_dim: batch_t_idxs})
+
+            # Forma canonica (time_dim, *dims_spaziali...) per ogni canale,
+            # preservando l'ordine relativo delle dimensioni spaziali (stesso
+            # ordine di flatten() usato per costruire coords_single). Il cast a
+            # float32 viene saltato se il dato è già in quel dtype, per evitare
+            # una copia inutile.
```
simone99n, line 491: Let's switch to English for the code comments

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -88,22 +112,49 @@ def _lazy_init(self) -> None:
...
+
+        # Apriamo con chunking Dask invece di chunks=None.
+        use_dask = self._stream_info.get("use_dask", True) and _HAS_DASK
+
+        ds = xr.open_zarr(
+            self._filename,
+            consolidated=True,
+            chunks=None,
+            **kwargs,
+        )
```
simone99n, line 132: Here we do xr.open_zarr(..., chunks=None --> no dask invocation) and then we re-chunk with ds.chunk(...)
We should save the dataset with the correct chunking and then use xr.open_zarr(..., chunks='auto') or  xr.open_zarr(..., chunks={})

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -375,73 +457,63 @@ def _get(self, idx: TIndex, channels_idx: list[int]) -> ReaderData:
...
-        data = np.vstack(data_arrays)
+        # Dentro ogni batch, tutti i canali richiesti vengono letti con
+        # un'unica selezione xarray/Dask: se il dataset è chunkato con Dask, i
+        # chunk Zarr coinvolti (canali x timestep del batch) vengono scaricati e
+        # decompressi in parallelo su più thread in una singola .compute().
+        n_channels = len(selected_channels)
+        n_valid_timesteps = len(valid_t_idxs)
+        ds_sub = self.ds[selected_channels]
+        batch_size = max(1, n_valid_timesteps)
+
```
simone99n, line 480: `batch_size = max(1, n_valid_timesteps)` is not clear to me: batch_size will be always equal to 0 or n_valid_timesteps --> the for loop below has only 0 or 1 iterations
Moreover, what is the definition of batch_size in this DataReader?

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -375,73 +457,63 @@ def _get(self, idx: TIndex, channels_idx: list[int]) -> ReaderData:
...
+                ds_batch[ch].transpose(time_dim, ...).data for ch in selected_channels
+            ]
+
+            if _HAS_DASK and isinstance(channel_arrays[0], dask_array.Array):
+                # Un'unica .compute(): Dask scarica/decomprime in parallelo tutti
+                # i chunk necessari per l'intero batch (tutti i canali e tutti i
+                # timestep del batch) invece di farlo in sequenza.
+                arr = dask_array.stack(channel_arrays, axis=-1).compute()
+            else:
+                arr = np.stack([np.asarray(a) for a in channel_arrays], axis=-1)
```
simone99n, line 503: Since batch_size can be only 0 or n_valid_timesteps, we have at most 1 batch --> the `.compute()` is not parallelized and we lose the advantages of using dask

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
```diff
@@ -375,73 +457,63 @@ def _get(self, idx: TIndex, channels_idx: list[int]) -> ReaderData:
...
 
-        # use actual count, not len(t_idxs)
-        n_valid_timesteps = len(data_arrays)
+            # (batch_t, *spatial_dims, n_channels) -> (batch_t * n_points, n_channels)
+            arr = np.asarray(arr, dtype=np.float32).reshape(len(batch_t_idxs) * self.n_points, n_channels)
```
simone99n, line 506: Here, `arr = np.asarray(...)` overwrites the `arr` variable that was computed in the if/else statement above

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
simone99n: Moreover, it fails if we have variables like `var_name_100` or `temperature_100m`.
Seems to be very brittle --> can we use regex?

simone99n (changes requested):
I left some comments.
I'm a little concerned about backward compatibility with the other dataset we use in WG --> we should test them before and after the changes

Moreover, we can split the __init__() in something like:
_open_dataset()
_init_time()
_init_grid()
_init_statistics()
_init_channels()
_init_coordinates()
This is just an idea

packages/readers_extra/src/weathergen/readers_extra/data_reader_grep.py
pierluigicosi: I agree, this is a workaround to have always a good chunking along time, but in the future we should re-chunk the dataset(s).

PR #2578, opened 2026-07-01
CRPS decoder integrated in latest develop

saibr:
## Description

This is the cleaned CRPS decoder code integrated in the latest develop on 1/7. It is based on the code from https://github.com/saibr/WeatherGenerator/tree/crps-decoder. This version should be easy to merge without conflicts. 

In the config, the following lines should be added to use the CRPS decoder:

```
decoder_ens_latent_perturbation : 
  num_members: 2    # 0 / 1 = disabled; >1 = number of ensemble members
  sigma_init: 0.2  # initial noise standard deviation
  sigma_learnable: True  # true = nn.Parameter; false = fixed buffer
```

Besides the changes on the decoder, I made a modification on the loss function, to make sure the model can run with the kernel_crps loss without including the mse loss, by using the following setup:
```
losses : {
    "physical": {
        type: LossPhysical,
        loss_fcts: {"mse": null, "kernel_crps": {}, },
        },
    }
```
In case there is an alternative for this that I missed, these modifications can be discarded.


src/weathergen/model/model.py
```diff
@@ -745,95 +762,137 @@ def predict_decoders(
...
-        batch_size = len(batch)
-        s = [batch_size, self.num_healpix_cells, self.cf.ae_local_num_queries, tokens.shape[-1]]
-        idxs = model_params.hp_nbours.unsqueeze(0).repeat((batch_size, 1, 1)).flatten(0, 1)
-        tokens_nbors = tokens.reshape(s).flatten(0, 1)[idxs.flatten()].flatten(0, 1)
-        # TODO: precompute in model_params?
-        tokens_nbors_lens = torch.full(
-            (s[0] * s[1] + 1,), fill_value=9, dtype=torch.int32, device=tokens_nbors.device
-        )
-        tokens_nbors_lens[0] = 0
+        # remove register and class tokens
```
clessig, line 827: Can we encapsulate this into a separate function

src/weathergen/model/model.py
```diff
@@ -745,95 +762,137 @@ def predict_decoders(
...
+        tokens = tokens[:, self.num_aux_tokens:]
+
+        B = len(batch)
```
clessig: The single letter vars make the code more difficult to read

src/weathergen/model/model.py
```diff
@@ -745,95 +762,137 @@ def predict_decoders(
...
+            sigma = torch.exp(self.latent_perturbation_log_sigma).to(tokens.dtype)
+
+            eps = torch.randn(M, B, H * Q, D, device=tokens.device, dtype=tokens.dtype)
+            tokens_tiled = (tokens.unsqueeze(0) + sigma * eps).reshape(M * B, H * Q, D)
+        else:
+            M = 1
+            tokens_tiled = tokens
+        assert tokens_tiled.shape == (M * B, H * Q, D)
+
+        MB = M * B
```
clessig: Is this variable really useful? Again, short var name is an issue

src/weathergen/model/model.py
```diff
@@ -745,95 +762,137 @@ def predict_decoders(
...
+        # flatten to cell level: tokens_flat[i*H + h] = token for (member i//B, batch i%B, cell h)
+        tokens_flat = tokens_tiled.reshape(MB, H, Q, D).flatten(0, 1)  # [MB*H, Q, D]
+
+        # neighbour gather, offset per (m, b) group since flat index space is now MB*H
+        cell_offsets = torch.arange(MB, device=tokens.device).view(MB, 1, 1) * H
+        idxs = (model_params.hp_nbours.unsqueeze(0) + cell_offsets).flatten(0, 1)  # [MB*H, 9]
+        tokens_nbors = tokens_flat[idxs.flatten()].flatten(0, 1)  # [MB*H*9*Q, D]
+
+        tokens_nbors_lens = tokens_flat.new_zeros(MB * H + 1, dtype=torch.int32)
+        tokens_nbors_lens[1:] = 9 * Q
```
clessig: We should not hard-code 9 but have a class-level variable with a useful name, e.g. `hp_num_nbors`

src/weathergen/model/model.py
```diff
@@ -745,95 +762,137 @@ def predict_decoders(
...
         for stream_name in self.streams.keys():
-            # extract target coords for current stream and fstep and convert to one tensor
             t_coords = [
                 batch.samples[i_b].streams_data[stream_name].target_coords[step]
-                for i_b in range(batch_size)
+                for i_b in range(B)
             ]
             t_coords_lens = [len(t) for t in t_coords]
             t_coords = torch.cat(t_coords)
+            P = t_coords.shape[0]
```
clessig: Again, single letter var name.

src/weathergen/train/loss_modules/loss_module_physical.py
```diff
@@ -118,7 +118,7 @@ def __init__(
                 name,
             ]
             for name, params in loss_fcts.items()
-            if name != "dynamic_loss"
+            if params is not None and name != "dynamic_loss"
```
clessig: Why is the additional condition necessary here?

src/weathergen/model/model.py
```diff
@@ -745,95 +762,137 @@ def predict_decoders(
...
+            M = 1
+            tokens_tiled = tokens
+        assert tokens_tiled.shape == (M * B, H * Q, D)
+
+        MB = M * B
+
+        # flatten to cell level: tokens_flat[i*H + h] = token for (member i//B, batch i%B, cell h)
+        tokens_flat = tokens_tiled.reshape(MB, H, Q, D).flatten(0, 1)  # [MB*H, Q, D]
+
+        # neighbour gather, offset per (m, b) group since flat index space is now MB*H
```
clessig: Comment not very helpful if it also uses single letter var names

src/weathergen/model/model.py
```diff
@@ -745,95 +762,137 @@ def predict_decoders(
         Predict outputs at the specific target coordinates based on the input weather state and
         pre-training task and projects the latent space representation back to physical space.
 
+        If `latent_perturbation_num_members` > 1 (and `latent_perturbation_log_sigma` is set),
+        this generates a CRPS-style ensemble by perturbing the assimilated latent tokens with
```
clessig: Why `CRPS-style` and not `CRPS`?

src/weathergen/datasets/masking.py
```diff
@@ -249,6 +249,8 @@ def parse_src_target_correspondence(self, losses, target_cfgs, source_cfgs) -> d
         corrs = []
         for _, loss_term in losses.items():
             for loss_name, loss_fct in loss_term.loss_fcts.items():
+                if loss_fct is None:
```
clessig: Why can we now have  `loss_fct = None`?

clessig: Ah, you have:

```
losses : {
    "physical": {
        type: LossPhysical,
        loss_fcts: {"mse": null, "kernel_crps": {}, },
        },
    }
```

but why not


```
losses : {
    "physical": {
        type: LossPhysical,
        loss_fcts: {"kernel_crps": {}, },
        },
    }
```

src/weathergen/model/model.py
```diff
@@ -578,6 +579,15 @@ def create(self) -> "Model":
                         use_patch_token=False,
                     )
 
+        # Latent-perturbation noise scale (learnable or fixed)
+        # torch.zeros() is used (not torch.tensor) so this respects the meta-device context used by FSDP. The actual value is filled in reset_parameters().
+        num_mem = cf.get("latent_perturbation_num_members", 0)
+        if num_mem > 1:
```
clessig: why not `if cf.get("latent_perturbation_num_members", 0) > 1 :`

clessig: With the changed config, we should use 

```
if cf.get("decoder_ens_latent_perturbation") is not None
```

or we check in the constructor and have a class variable for this.

src/weathergen/model/model.py
```diff
@@ -745,95 +762,137 @@ def predict_decoders(
...
         pre-training task and projects the latent space representation back to physical space.
 
+        If `latent_perturbation_num_members` > 1 (and `latent_perturbation_log_sigma` is set),
+        this generates a CRPS-style ensemble by perturbing the assimilated latent tokens with
+        Gaussian noise and running the decoder once per member.
+
         Args:
             model_params : Query and embedding parameters
-            fstep : Number of forecast steps
+            step : Forecast step
```
clessig, line 887: Why not keep `fstep`?

src/weathergen/model/model.py
```diff
@@ -745,95 +762,137 @@ def predict_decoders(
...
+        # fully vectorised. Layout convention: M is the OUTER axis, B is INNER.
+        #   tokens_tiled[m*B + b] belongs to member m, batch item b
+        M = self.cf.get("latent_perturbation_num_members", 0)
+        if M > 1 and self.latent_perturbation_log_sigma is not None:
+            sigma = torch.exp(self.latent_perturbation_log_sigma).to(tokens.dtype)
+
+            eps = torch.randn(M, B, H * Q, D, device=tokens.device, dtype=tokens.dtype)
+            tokens_tiled = (tokens.unsqueeze(0) + sigma * eps).reshape(M * B, H * Q, D)
+        else:
+            M = 1
```
clessig: Avoid by `M = self.cf.get("latent_perturbation_num_members", 1)` above. But I think it should be valid to have `self.cf.get("latent_perturbation_num_members", 1)` and then we also add noise for decoding.

src/weathergen/model/model.py
```diff
@@ -745,95 +762,137 @@ def predict_decoders(
...
+        assert tokens.shape == (B, H * Q, D), f"unexpected token shape {tokens.shape}"
+        # tokens_nbors_lens below assumes 9 neighbours * Q tokens/cell; only verified for Q=1.
+        assert Q == 1, "predict_decoders assumes ae_local_num_queries==1; revisit tokens_nbors_lens if Q>1"
+
+        # ---- CRPS latent perturbation: build M ensemble members ----------
+        # Each member is treated as an extra "batch" element so the decoder runs once,
+        # fully vectorised. Layout convention: M is the OUTER axis, B is INNER.
+        #   tokens_tiled[m*B + b] belongs to member m, batch item b
+        M = self.cf.get("latent_perturbation_num_members", 0)
+        if M > 1 and self.latent_perturbation_log_sigma is not None:
```
clessig: Maybe really good to have a class variable if CRPS-perturbs are used or not

src/weathergen/model/model.py
```diff
@@ -745,95 +762,137 @@ def predict_decoders(
...
-        s = [batch_size, self.num_healpix_cells, self.cf.ae_local_num_queries, tokens.shape[-1]]
-        idxs = model_params.hp_nbours.unsqueeze(0).repeat((batch_size, 1, 1)).flatten(0, 1)
-        tokens_nbors = tokens.reshape(s).flatten(0, 1)[idxs.flatten()].flatten(0, 1)
-        # TODO: precompute in model_params?
-        tokens_nbors_lens = torch.full(
-            (s[0] * s[1] + 1,), fill_value=9, dtype=torch.int32, device=tokens_nbors.device
-        )
-        tokens_nbors_lens[0] = 0
+        # remove register and class tokens
+        tokens = tokens[:, self.num_aux_tokens:]
```
clessig, line 828: If we call this tokens_tiled (or a bit later) then we don't need the else branch in 803

clessig (commented):
Thanks for the implementation. Left some comments: I think the code can be a bit tidier overall. Three questions:
1. Should we generate a warning if training is with multiple ens members and CRPS perturbations but without CRPS loss (or another ens loss)?
2. Do we know how the memory requirements change with the CRPS decoder?
3. Do we have any training results?

For the config, can we make it:

```
decoder_ens_latent_perturbation : 
  num_members: 2    # 0 / 1 = disabled; >1 = number of ensemble members
  sigma_init: 0.2  # initial noise standard deviation
  sigma_learnable: True  # true = nn.Parameter; false = fixed buffer
```

This is clearer and more composable (we still need to refactor the model config into dicts).

clessig:
@saibr : can you also lint (this might also complain about single letter variables ;)) and open a PR.

src/weathergen/model/model.py
saibr: Which lines are you referring to for the separate function? This comment refers only to the single line below

src/weathergen/model/model.py
saibr: predict_decoders takes step as input, not fstep. This is not something I changed, it is used in the latest develop. I believe the comment wasn't updated before.

src/weathergen/datasets/masking.py
saibr: If I remove mse from loss_fcts, it still uses the mse loss. I believe it has to do with how OmegaConf is using the default_config.yml. It seems to merge the new dict with this default dict. Therefore I used `mse: null` and made some additional changes to make it work such as `loss_fct = None`

src/weathergen/train/loss_modules/loss_module_physical.py
saibr: Because of the `mse` in default_config.yml

saibr:
Thank you for all the comments @clessig , I incorporated them. I am not sure how to deal with the scenario of `num_members: 1`. I added noise to it, but the current implementation of the kernel crps loss does not allow num_members to be 1. Let me know if you have a specific use case for this in mind and I can adapt it accordingly.

src/weathergen/model/model.py
clessig: Ok, thanks

src/weathergen/datasets/masking.py
clessig: As above: The standard workflow is to use `--base-config`; please use override configs only for subsequent stages. Then only the loss terms you specify in the base config will be present.

src/weathergen/model/model.py
clessig: The whole block that is green (which I think is a rewrite of some code that was there before and some new code).

src/weathergen/datasets/masking.py
saibr: I'm not sure if I understand it correctly. Is your suggestion to start a new train run with both `--config` and `--base-config`, where `--base-config` is a modification of `default_config.yml` without the mse loss?

I usually only specify --config, but when I set 
```
losses : {
    "physical": {
        type: LossPhysical,
        loss_fcts: {"kernel_crps": {}, },
        },
    }
```
in the config file, it doesn't override the loss function, but it merges it's content with those from the base-config. For every other variable it overrides it, but not for the dict `losses`

src/weathergen/model/model.py
```diff
@@ -578,6 +580,24 @@ def create(self) -> "Model":
                         use_patch_token=False,
                     )
 
+        # Latent-perturbation noise scale (learnable or fixed)
+        self.use_latent_perturbation = (
+            self.decoder_ens_latent_perturbation is not None
+            and self.decoder_ens_latent_perturbation.get("num_members", 0) >= 1
```
shmh40: Should this be > 1? Do we only want this to work when the user has specified multiple members? What happens if they set it as 1? Do you get a perturbed single member?

shmh40: Otherwise the implementation looks ok to me, thanks @saibr !

saibr: @shmh40 Christian mentioned in a comment that perturbation on 1 member would be a valid case, so I included it, but I'm not sure about the use case. The current implementation of the crps loss does not accept one member, so in my first implementation I discarded the scenario with 0 and 1 members.

src/weathergen/model/model.py
saibr: I created _build_ensemble_latent_tokens for this

src/weathergen/model/model.py
```diff
@@ -745,94 +910,134 @@ def predict_decoders(
...
+
+                    pred = self.pred_heads[stream_name](tc_tokens_out)
+                    ens_head_size, total_target_points, output_channels = pred.shape
+                    assert (
+                        total_target_points == num_members * t_coords.shape[0]
+                    ), (
+                        f"expected {num_members * t_coords.shape[0]} pts in pred, "
+                        f"got {total_target_points}"
+                    )
+                    pred = (
```
clessig: This statement cuts across 12 lines and is extremely difficult to read for me. Can we precompute the shapes to get this statement in 1 or 2 lines

src/weathergen/model/model.py
```diff
@@ -745,94 +910,134 @@ def predict_decoders(
...
+                num_groups = tcs_lens.shape[0] - 1
+                assert batch_size * self.num_healpix_cells == num_groups, (
+                    f"expected {batch_size * self.num_healpix_cells} query groups, "
+                    f"got {num_groups}"
+                )
 
                 if self.cf.decoder_type == "Linear":
+                    # repeat target-coord tokens once per ensemble member,
+                    # matching tokens_tiled's ordering
+                    tc_tokens_in = tc_tokens.repeat(num_members, 1)
```
clessig, line 942: Can't we repeat the output of line 983 instead of repeating the input? There is no ensembling happening here. Otherwise, the ` if self.cf.decoder_type == "Linear":` branch should be in line 1002, I think.

src/weathergen/model/model.py
```diff
@@ -745,94 +910,134 @@ def predict_decoders(
...
-                        output_lens=tcs_lens,
-                        coordinates=t_coords,
+                    # Run members sequentially to stay under CUDA's max grid size
+                    # for large ensembles or large hpl. Gather HEALPix neighbor
+                    # tokens inside the loop so memory scales with one member
+                    # instead of num_members.
+                    tc_tokens_outs = []
+                    for m in range(num_members):
+                        member_tokens = tokens_tiled[m * batch_size: (m + 1) * batch_size]
+                        tokens_nbors, tokens_nbors_lens = self._gather_neighbor_tokens(
```
clessig, line 964: thank you

src/weathergen/model/model.py
```diff
@@ -745,94 +910,134 @@ def predict_decoders(
...
-        batch_size = len(batch)
-        s = [batch_size, self.num_healpix_cells, self.cf.ae_local_num_queries, tokens.shape[-1]]
-        idxs = model_params.hp_nbours.unsqueeze(0).repeat((batch_size, 1, 1)).flatten(0, 1)
-        tokens_nbors = tokens.reshape(s).flatten(0, 1)[idxs.flatten()].flatten(0, 1)
-        # TODO: precompute in model_params?
-        tokens_nbors_lens = torch.full(
-            (s[0] * s[1] + 1,), fill_value=9, dtype=torch.int32, device=tokens_nbors.device
-        )
-        tokens_nbors_lens[0] = 0
+        ens = self._build_ensemble_latent_tokens(model_params, tokens, batch)
```
clessig: It looks like `EnsembleLatentTokens` is only used as return type for the function. Then we should not introduce it and just return a tuple/list with the values.

src/weathergen/model/model.py
```diff
@@ -578,6 +592,46 @@ def create(self) -> "Model":
...
+
+                if not (configured_loss_fcts & ensemble_aware_loss_fcts):
+                    logger.warning(
+                        f"decoder_ens_latent_perturbation.num_members={num_members} (>1) is "
+                        f"configured, but none of the enabled loss_fcts {sorted(configured_loss_fcts)} "
+                        "are ensemble-aware"
+                    )
+
+        if self.use_latent_perturbation:
+            if self.decoder_ens_latent_perturbation.get("sigma_learnable", True):
```
clessig: The branches differ just by `requires_grad`. We can save a few lines and make the code more readable by having one statement and just having the flag conditional (either using a temp variable or by directly having `self.decoder_ens_latent_perturbation.get("sigma_learnable", True)` there. (`self.decoder_ens_latent_perturbation` is very long and leads to a lot of line breaks that make the code difficult to read; maybe use a shorter name, e.g. `self.ens_latent_perturb`

src/weathergen/model/model.py
```diff
@@ -731,6 +791,111 @@ def predict_latent(
...
+        )
+
+        num_members = (
+            self.decoder_ens_latent_perturbation.get("num_members", 1)
+            if self.use_latent_perturbation
+            else 1
+        )
+
+        tokens_tiled = tokens
+        if self.use_latent_perturbation:
```
clessig: Could one return from this function immediately at the beginning if `self.use_latent_perturbation == False`

clessig (commented):
Thanks for the changes. Some more comments.

src/weathergen/model/model.py
saibr: I believe we cannot remove the repeat output instead if the input, the ensemble component is in `tokens_tiled` which is create by `_build_ensemble_latent_tokens`. If we repeat the output `num_members `times, we get the same prediction for each member

src/weathergen/model/model.py
saibr: Changed in latest commit

src/weathergen/model/model.py
saibr: I agree, changed

src/weathergen/model/model.py
saibr: Yes thanks, changed it

clessig:
Closes via #2623

PR #2639, opened 2026-07-14
Muon optimizer

### LINKED ISSUE #2638: Support Muon Optimizer

Currently, we only use AdamW optimizer but recently, there are several works which prefer Muon over AdamW including DLWP models (https://arxiv.org/pdf/2604.16429 and https://arxiv.org/abs/2604.09041).

The issue deals with supporting Muon Optimizer in the current codebase

https://gitlab.jsc.fz-juelich.de/hedgedoc/x0JCAfWdQX-1brM9tY7sXw?edit

### Area

- [ ] model
- [x] science
- [ ] documentation
- [x] performance

ankitpatnala:
## Description

Implemented Muon optimizer 
*  Currently, it is only implemented for forecasting configs. `config/config_forecasting.yml`
*  Just adapted the code to log multiple lr, to keep track of lr of AdamW and Muon separately (needs to be tested)

## Issue Number
Closes #2638


ankitpatnala:
@sophie-xhonneux 
@florianscheidl

clessig:
@ankitpatnala : could you outline the hyperparams that are involved. How should the lr change between muon and AdamW?

ankitpatnala:
The `name` field inside `optimizer` selects which optimizer is used: `adamw` or `muon`. If it is set to `adamw`, the `muon` config block is ignored.
 
 ```yaml
   optimizer:
    # which optimizer to use: "adamw" or "muon"
    # muon only optimizes 2D weight matrices of hidden layers; all other
    # parameters (biases, norms, ...) are still optimized with adamw
    name: muon
    grad_clip: 1.0
    weight_decay: 0.1
    log_grad_norms: False
    adamw :
      # parameters are scaled by number of DDP workers
      beta1 : 0.98125 # == 0.85 on 2 nodes x 4 gpus
      beta2 : 0.9875  # == 0.90 on 2 nodes x 4 gpus
      eps : 2e-08
    muon:
      # peak lr override for muon; null shares adamw's lr schedule (relies on
      # adjust_lr_fn=match_rms_adamw's per-matrix scaling instead of a fixed multiplier)
      lr_max: null
      momentum: 0.95
      nesterov: True
      ns_steps: 5
      eps: 1e-7
      # "original" | "match_rms_adamw"
      adjust_lr_fn: match_rms_adamw
```
 
When it is set to `muon`, the general hyperparameters stay the same:
 
```yaml
grad_clip: 1.0
weight_decay: 0.1
log_grad_norms: False
```
 
The optimizer then explicitly checks whether each parameter is 2D ([optimizer.py#L97](https://github.com/ankitpatnala/WeatherGenerator/blob/d46e95e0bce5f3d8373f546fc871cc83698be88a/src/weathergen/train/optimizer/optimizer.py#L97)):
 
- **Non-2D parameters** (biases, norms, etc.) use the base AdamW config:
```yaml
adamw:
  # parameters are scaled by number of DDP workers
  beta1: 0.98125  # == 0.85 on 2 nodes x 4 gpus
  beta2: 0.9875   # == 0.90 on 2 nodes x 4 gpus
  eps: 2e-08
```
 
- **2D parameters** are updated by Muon, using this config:
```yaml
muon:
  # peak lr override for muon; null shares adamw's lr schedule (relies on
  # adjust_lr_fn=match_rms_adamw's per-matrix scaling instead of a fixed multiplier)
  lr_max: null
  momentum: 0.95
  nesterov: True
  ns_steps: 5
  eps: 1e-7
  # "original" | "match_rms_adamw"
  adjust_lr_fn: match_rms_adamw
```
 
If `lr_max` is defined, Muon's learning rate is scaled between this value and the `lr_max` set in `learning_rate_scheduling` [optimizer.py#37](https://github.com/ankitpatnala/WeatherGenerator/blob/d46e95e0bce5f3d8373f546fc871cc83698be88a/src/weathergen/train/optimizer/optimizer.py#L37). It is typically set to `1e-2`, but that did not work for me:
 
```yaml
learning_rate_scheduling:
  lr_start: 1e-6
  lr_max: 5e-5
  lr_final_decay: 2e-6
  lr_final: 0.0
  num_steps_warmup: 256
  num_steps_cooldown: 512
  policy_warmup: "cosine"
  policy_decay: "constant"
  policy_cooldown: "linear"
  parallel_scaling_policy: "sqrt"
```
 
All other Muon parameters take their defaults from [`torch.optim.Muon`](https://docs.pytorch.org/docs/2.13/generated/torch.optim.Muon.html).
 
I tested roughly `lr_max=1e-2`, and training failed — I'm not yet sure why. Leaving `lr_max: null` together with `adjust_lr_fn: match_rms_adamw` ([optimizer.py#L56](https://github.com/ankitpatnala/WeatherGenerator/blob/d46e95e0bce5f3d8373f546fc871cc83698be88a/src/weathergen/train/optimizer/optimizer.py#L56)) instead gives stable training. Because Muon optimizer now handles the LR scaling internally, it logs the same learning rate for both optimizers.

src/weathergen/train/optimizer/optimizer.py
```diff
@@ -0,0 +1,135 @@
...
+import numpy as np
+import torch
+from omegaconf import OmegaConf
+
+from weathergen.utils.distributed import is_root
+
+logger = logging.getLogger(__name__)
+
+
+def _adamw_betas_eps(optimizer_cfg, kappa: float) -> dict:
```
sophie-xhonneux, line 20: Only tangentially related to this PR, but do we want to remove this scaling of beta parameters with batch size because I think it makes setting those variables so much more complicated

ankitpatnala: Should we then hardcode it for the 2-node, 4-GPU setup?

florianscheidl:
Hey Ankit,
I ran some memory and pytorch profiler traces with and without the Muon optimizer on config_forecasting and a fixed seed. 

- **Little impact on memory**: qualitatively, the memory traces look similar, the peak memory is in the same ballpark. 

**Adam:**
Screenshot 2026-07-15 at 19 27 22

**Muon**:
Screenshot 2026-07-15 at 19 27 28

- **Computational overhead**: The Muon optimizer launches many small tensor ops. We could investigate if this can be optimized. We see a sum in the gradient scaler, which launches a sum and a synchronization. We can probably remove gradient scaling (since we're training with fp32 and bf16) and thereby the sync.

**Muon**
Screenshot 2026-07-16 at 10 06 15

**Adam**: The optimizer step is nearly negligible, so not showing it here.

Note: these are only qualitative, not quantitative results. For the latter, we could compare the peak memory consumption and throughput for a longer run.

clessig:
> Hey Ankit, I ran some memory and pytorch profiler traces with and without the Muon optimizer on config_forecasting and a fixed seed.
> 
> * **Little impact on memory**: qualitatively, the memory traces look similar, the peak memory is in the same ballpark.
> 
> **Adam:** Screenshot 2026-07-15 at 19 27 22
> 
> **Muon**: Screenshot 2026-07-15 at 19 27 28
> 
> * **Computational overhead**: The Muon optimizer launches many small tensor ops. We could investigate if this can be optimized. We see a sum in the gradient scaler, which launches a sum and a synchronization. We can probably remove gradient scaling (since we're training with fp32 and bf16) and thereby the sync.
> 
> **Muon** Screenshot 2026-07-16 at 10 06 15
> 
> **Adam**: The optimizer step is nearly negligible, so not showing it here.
> 
> Note: these are only qualitative, not quantitative results. For the latter, we could compare the peak memory consumption and throughput for a longer run.

Thanks! Could we also try with the config with the GEOs that is more demanding in terms of memory to make sure we don't break configs that are currently working.

src/weathergen/train/optimizer/optimizer.py
clessig: We would need at least some evidence that it's neutral and doesn't affect skill.

ankitpatnala:
@clessig @sophie-xhonneux  this PR is now ready for review. Training plots for the current forecasting config are in this HedgeDoc:https://gitlab.jsc.fz-juelich.de/hedgedoc/x0JCAfWdQX-1brM9tY7sXw?both#New-run-after-cleanup

src/weathergen/train/optimizer/optimizer.py
sophie-xhonneux: I definitively would not hard code it @ankitpatnala

src/weathergen/train/optimizer/optimizer.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def _muon_adjust_lr_factor(shape, adjust_lr_fn: str) -> float:
+    """
+    Mirrors torch.optim.Muon's internal per-parameter lr-adjustment factor
+    (torch/optim/_muon.py::_adjust_lr), so a representative effective lr can be logged.
+    """
+    a, b = shape[0], shape[1]
+    if adjust_lr_fn == "match_rms_adamw":
+        return 0.2 * max(a, b) ** 0.5
```
sophie-xhonneux, line 43: a) the `0.2` factor seems arbitrary, can we at least explain in a comment as to why it is this factor?
b) I would either give this function a more general name if it is useful for more optimizers or make it a function of the Muon class

ankitpatnala: I'm using this only to log Muon's initial lr (called in __init__).
The factor is applied internally by PyTorch (see https://docs.pytorch.org/docs/2.13/generated/torch.optim.Muon.html). In PyTorch 2.9, it supports two options: `original` (Keller's implementation) and `match_rms_adamw` (Moonshot's implementation, where the 0.2 factor was defined through empirical observation). Because the adjustment happens internally, we don't have direct access to the value to log it, so we reimplemented the same computation ourselves. In principle, it isn't required.

ankitpatnala: @sophie-xhonneux I think it is better if I remove the logging part.

sophie-xhonneux: @ankitpatnala

ankitpatnala: > @sophie-xhonneux I think it is better if I remove the logging part.

Should I remove it?

PR #2662, opened 2026-07-20
Restore SEEPS

### LINKED ISSUE #2305: Compute SEEPS weights

The seeps weights we used for AtmoRep got lost so we need to recompute them. I will first ask around if there is a way to get the same ones used at ECMWF or by some member states.   

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation
- [ ] performance

### LINKED ISSUE #2657: Restore SEEPS score

As far as I can tell, the currently implemented SEEPS has no way of accessing the climatological weights needed for its computation (see Issue #2305). We could treat the weights like the climatology used for acc etc.
In addition, the existing implementation is hard to read and appears to not be vectorized over time, I would recommend just re-implementing it using xarray. We should also maybe add the option to remove gridpoints that are too dry or too wet from the computation, as recommended in the original paper https://onlinelibrary.wiley.com/doi/abs/10.1002/qj.656. 

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation
- [ ] performance

s6sebusc:
## Description

This PR contains
1. all the plumbing to process the newly computed seeps climatology computation see [here](https://gitlab.jsc.fz-juelich.de/esde/WeatherGenerator-private/-/blob/buschow1/seeps_climatology/data/seeps_climatology/compute_seeps_climatology.py) through the score and plot pipeline
2. a new implementation of the score itself using xarray, which is shorter and hopefully easier to understand than the original (possibly also faster)

Passing a second auxiliary dataset (besides the original climatology) through all these function calls looks pretty bad to me and is clearly not scalable if we later decide to add more scores with different requirements. A better solution might be to include the SEEPS things into the existing climatology file. I think this would kind of work for SEEPS, but maybe not for future auxiliary datasets we might need. Another alternative would be to pass around a single climatology dict containing all auxiliary data we need, collected from various files under the hood. I did not attempt to implement that because of the impending refactor.

Tested with this config on jupiter (`climatology_path` is probably necessary, the other paths were just because I'm not allowed to write to the directory of that run):
```yaml
evaluation:
  metrics:
    - seeps
  regions: ["global"]
  summary_plots : true
  summary_dir: "./plots/"
  plot_ensemble: "members" #supported: false, "std", "minmax", "members"
  plot_score_maps: true #plot scores on a 2D maps. it slows down score computation 
  print_summary: true #print out score values on screen. it can be verbose
  log_scale: false
  add_grid: false
  score_cards: true
  bar_plots: false


run_ids :
    n46jvepr:
      type: "zarr"
      label: "Threshold20"
      mini_epoch: 0
      rank: 0
      metrics_dir: "../imerg_metrics/"
      runplot_base_dir: "../score_map_plots/"
      streams:
        IMERG_ANEMOI:
          climatology_path: "/e/scratch/weatherai/shared_work/assets/climatology/nasa-imerg-grib-n320-1998-2024-6h-v1_climatology.zarr"
          channels: ["tp"] 
          evaluation:
            forecast_step: "all"
            sample: "all"
            ensemble: "all" #supported: "all", "mean", [0,1,2]
          plotting:
            sample: [0,1]
            forecast_step: [0,1]
            ensemble: [0] #supported: "all", "mean", [0,1,2]
            plot_maps: false
            plot_histograms: false
            plot_animations: false
```

## Issue Number

Closes #2657
Closes #2305

### DIFF
--- a/packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
+++ b/packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
@@ -103,9 +103,7 @@ class WeatherGenReader(Reader):
 
         return config
 
-    def get_climatology_filename(
-        self, stream: str, climatology_type: str = "default_climatology"
-    ) -> str | None:
+    def get_climatology_filename(self, stream: str) -> str | None:
         """
         Get the climatology filename for a given stream from the inference
         configuration.
@@ -153,18 +151,12 @@ class WeatherGenReader(Reader):
                 " explicitly via 'climatology_path' in the evaluation config."
             )
             return None
-        if climatology_type == "default_climatology":
-            clim_data_path = (
-                Path(clim_base_dir)
-                / "climatology"
-                / climatology_partial_filename.replace(".zarr", "_climatology.zarr")
-            )
-        elif climatology_type == "seeps":
-            clim_data_path = (
-                Path(clim_base_dir)
-                / "climatology_seeps"
-                / climatology_partial_filename.replace(".zarr", "_SEEPS.nc")
-            )
+
+        clim_data_path = (
+            Path(clim_base_dir)
+            / "climatology"
+            / climatology_partial_filename.replace(".zarr", "_climatology.zarr")
+        )
 
         if not clim_data_path.exists():
             _logger.warning(
--- a/packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
+++ b/packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
@@ -103,7 +103,9 @@ class WeatherGenReader(Reader):
 
         return config
 
-    def get_climatology_filename(self, stream: str, climatology_type: str = "default_climatology") -> str | None:
+    def get_climatology_filename(
+        self, stream: str, climatology_type: str = "default_climatology"
+    ) -> str | None:
         """
         Get the climatology filename for a given stream from the inference
         configuration.
--- a/packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
+++ b/packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
@@ -103,7 +103,7 @@ class WeatherGenReader(Reader):
 
         return config
 
-    def get_climatology_filename(self, stream: str) -> str | None:
+    def get_climatology_filename(self, stream: str, climatology_type: str = "default_climatology") -> str | None:
         """
         Get the climatology filename for a given stream from the inference
         configuration.
@@ -151,12 +151,18 @@ class WeatherGenReader(Reader):
                 " explicitly via 'climatology_path' in the evaluation config."
             )
             return None
-
-        clim_data_path = (
-            Path(clim_base_dir)
-            / "climatology"
-            / climatology_partial_filename.replace(".zarr", "_climatology.zarr")
-        )
+        if climatology_type == "default_climatology":
+            clim_data_path = (
+                Path(clim_base_dir)
+                / "climatology"
+                / climatology_partial_filename.replace(".zarr", "_climatology.zarr")
+            )
+        elif climatology_type == "seeps":
+            clim_data_path = (
+                Path(clim_base_dir)
+                / "climatology_seeps"
+                / climatology_partial_filename.replace(".zarr", "_SEEPS.nc")
+            )
 
         if not clim_data_path.exists():
             _logger.warning(
--- a/packages/evaluate/src/weathergen/evaluate/scores/score.py
+++ b/packages/evaluate/src/weathergen/evaluate/scores/score.py
@@ -15,6 +15,7 @@ import numpy as np
 import pandas as pd
 import xarray as xr
 from scipy.spatial import cKDTree
+import scores
 
 from weathergen.evaluate.scores.psd import compute_psd_score, detect_grid_type
 from weathergen.evaluate.scores.score_utils import calc_latitude_weights, to_list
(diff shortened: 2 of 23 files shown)


clessig:
@iluise @jpolz : See above. You know this better but to me it also appears that we should think about a more robust and scale approach to handling the climatology, both for the pre-computation and during scoring. Happy to brain storm together.

jpolz:
> @iluise @jpolz : See above. You know this better but to me it also appears that we should think about a more robust and scale approach to handling the climatology, both for the pre-computation and during scoring. Happy to brain storm together.

I think currently this seems to be the correct way to do it. It is only for precipitation so it should perform well enough (likely high res but not many channels).

Happy to discuss ideas on how to improve. Not sure though if we should wait until @iluise is done with the restructuring of the repo.

Regarding pre-computation: Is this currently critical? To me there is no way around precomputed climatology. One could split the current statistics dimension into independent data vars, then the seeps statistics could be integrated easily without inflating that dimension. We already have different scores that require different statistics so that option would definitely avoid the changes made in this PR

clessig:
> > @iluise @jpolz : See above. You know this better but to me it also appears that we should think about a more robust and scale approach to handling the climatology, both for the pre-computation and during scoring. Happy to brain storm together.
> 
> I think currently this seems to be the correct way to do it. It is only for precipitation so it should perform well enough (likely high res but not many channels).
> 
> Happy to discuss ideas on how to improve. Not sure though if we should wait until @iluise is done with the restructuring of the repo.
> 
> Regarding pre-computation: Is this currently critical? To me there is no way around precomputed climatology. One could split the current statistics dimension into independent data vars, then the seeps statistics could be integrated easily without inflating that dimension. We already have different scores that require different statistics so that option would definitely avoid the changes made in this PR

@jpolz : Precomputation is not critical but it's something we should think about--especially a good solution is not obvious :)

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
```diff
@@ -151,12 +153,18 @@ def get_climatology_filename(self, stream: str) -> str | None:
...
-            / "climatology"
-            / climatology_partial_filename.replace(".zarr", "_climatology.zarr")
-        )
+        if climatology_type == "default_climatology":
+            clim_data_path = (
+                Path(clim_base_dir)
+                / "climatology"
+                / climatology_partial_filename.replace(".zarr", "_climatology.zarr")
+            )
+        elif climatology_type == "seeps":
```
iluise: I was thinking: Would it make sense to have one single file with SEEPS and Climatology infos together? In the way we have all the machinery in place already. 
We should just add the SEEPS code to the existing climatology code. What do you think?

packages/evaluate/src/weathergen/evaluate/scores/score.py
```diff
@@ -1332,112 +1335,31 @@ def calc_spatial_variability(
...
-            prediction, ground_truth = p, gt
-            lstack = False
-        else:
-            raise ValueError("Data must be a two-or-three-dimensional array.")
-
-        # check dimensioning of data
-        assert prediction.ndim <= 2, (
-            f"Data must be one- or two-dimensional, but has {prediction.ndim} dimensions. "
-            "Check if stacking with spatial_dims may help."
+    def calc_seeps(self, p: xr.DataArray, gt: xr.DataArray, c_seeps: xr.Dataset) -> xr.DataArray:
```
iluise: Can't we use the SEEPS score in `scores` directly? 
https://scores.readthedocs.io/en/stable/tutorials/SEEPS.html


Also, in that page I just discovered that there are ERA5 climatologies pre-computed in WeatherBench as well:

clim = xr.open_zarr(
    "gs://weatherbench2/datasets/era5-hourly-climatology/1990-2017_6h_240x121_equiangular_with_poles_conservative.zarr"
)


maybe we should check if they could be of interest for us cc @jpolz

iluise (commented):
I think the cleanest way would be to have a single file with climatologies and seeps thresholds together, a bit like what WeatherBench seems to provide (see in the comments below)

 I am a bit reluctant in adding the difference between default_climatology and seeps_climatology as we have full control of the climatology file structure.

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
s6sebusc: Yes as I wrote above that's probably better, at least for SEEPS. My only concern would be that the SEEPS values are defined per month, not per calendar day so I'm not sure how that would fit in (I also have no idea how or where the existing climatology was computed). Also they are only defined for one channel, so computing and storing them for all channels of ERA5 would be a pretty big waste and also meaningless.

jpolz: As said, we would need to turn the current statistics dimension of the climatology into separate variables. Then upon computation you can just define SEEPS channels and it will do it only for them.
Are you using a rolling month or just calendar months? Fastest would be to still repeat the climatology along the datasets native temporal resolution. Just a bit of a waste on the storage side

s6sebusc: Ahh I see. As far as I can tell the definition of seeps calls for one statistical value per calendar month, not rolling, so thats what I did (see [here](https://gitlab.jsc.fz-juelich.de/esde/WeatherGenerator-private/-/blob/buschow1/seeps_climatology/data/seeps_climatology/compute_seeps_climatology.py?ref_type=heads), it's a very small script).

packages/evaluate/src/weathergen/evaluate/scores/score.py
s6sebusc: done.

packages/evaluate/src/weathergen/evaluate/io/wegen_reader.py
s6sebusc: changed it so that the seeps statistics are expected to be in the regular climatology file. For IMERG, I created a climatology file that contains only those statistics. If we later want to compute also acc for that dataset we have to consolidate the climatology files.

packages/evaluate/src/weathergen/evaluate/scores/score.py
```diff
@@ -1332,113 +1334,19 @@ def calc_spatial_variability(
...
-        assert prediction.ndim <= 2, (
-            f"Data must be one- or two-dimensional, but has {prediction.ndim} dimensions. "
-            "Check if stacking with spatial_dims may help."
+    def calc_seeps(self, p: xr.DataArray, gt: xr.DataArray, c: xr.Dataset) -> xr.DataArray:
+        return scores.categorical.seeps(
+            fcst=p * 1000,  # converted to mm
+            obs=gt * 1000,
+            prob_dry=c.sel(statistic="prob_dry"),
+            light_heavy_threshold=c.sel(statistic="light_heavy_threshold"),
+            dry_light_threshold=0.2,
```
iluise, line 1350: Why these scores are hardcoded? Shouldn't we add them as customizable options to the SEEPS score like for e.g. the FBI or ETS scores?

iluise (commented):
Great! I tested it and it works. I am just a bit puzzled about the hardcoded thresholds. Can we make it more flexible?
Does it depend on the thresholds you used to compute the weights ?


p.s. Can you lint and solve the checks in red?

packages/evaluate/src/weathergen/evaluate/scores/score.py
s6sebusc: dry_light_threshold is kind of universally set to 0.2 for seeps as far as I can tell. It is used in the computation of the weights so users can't just select another value on the fly. We can make lower_masked_value and upper_masked value user parameters if we want. They are kind of arbitrary but the values I set here (defaults from scores) seem to be what people typically use.

iluise: Allright, let's keep it like this then. Can you add these details to the documentation file here? https://github.com/ecmwf/WeatherGenerator/blob/develop/docs/evaluate_config_reference.md

I will merge the PR right after

s6sebusc: Done.

iluise (approved):

PR #2670, opened 2026-07-22
Fix export bug, add IMERG N320

### LINKED ISSUE #2669: export works only if ensemble coordinate is present
### What happened?

Just a small bug, `data_dims` are currently only assigned if the ensemble coordinate is present. Datasets where it is completely absent (as opposed to present with size 1) will result in a crash.
For me, export also complained about a wrong indent in the zarr2cf config, that will be fixed too. 

_No response_

_No response_

s6sebusc:
## Description

Just a small PR to add to enable exporting high-res IMERG (where precip is called `tp` and not `tp_imerg_0` for some reason) and fix two small bugs. 

## Issue Number

Closes #2669

### DIFF
--- a/config/evaluate/config_zarr2cf.yaml
+++ b/config/evaluate/config_zarr2cf.yaml
@@ -96,7 +96,7 @@ variables:
     wg_unit: m
     std_unit: m
     level_type: sfc
-   hcc:
+  hcc:
     var: hcc
     long: high_cloud_cover
     std: cloud_area_fraction_high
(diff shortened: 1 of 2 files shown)


iluise (approved):

config/evaluate/config_zarr2cf.yaml
```diff
@@ -138,6 +138,13 @@ variables:
     wg_unit: J m**-2
     std_unit: J m-2
     level_type: sfc
+  tp:
+    var: tp
+    long: imerg_total_precipitation
```
clessig, line 143: Can this only be IMERG? ERA5 and many other datasets have tp. The other standard precipitation variable is cp (https://codes.ecmwf.int/grib/param-db/143). Could we also add it to be future proof.

iluise: Ah that's a good catch. I missed this in the review yesterday. Indeed it should be just `total_precipitation`

PR #2683, opened 2026-07-25
Mh/diffusion full pipeline latent rmse curve

moritzhauschulz:
## Description

This PR enables plotting of latent RMSE during inference of the diffusion model. Simply enable via `test_config.latent_rollout_rmse=True`.

**NB**: This involved adding a new batch to the sampler (see changes). I dropped the idea of reusing the `num_steps_input` as this would mess with the indexing of the entire pipeline, and there are many ways that it could fail silently. This implementation is a bit more intrusive, but all changes are gated with a flag, so that the standard path is minimally disrupted. If this is to become a permanent feature on develop, I think a broader discussion should be had about what is the best way.

## Issue Number

### DIFF
--- a/src/weathergen/datasets/batch.py
+++ b/src/weathergen/datasets/batch.py
@@ -309,6 +309,23 @@ class ModelBatch:
         self.source2target_matching_idxs = np.full(num_source_samples, -1, dtype=np.int32)
         self.target2source_matching_idxs = [[] for _ in range(num_target_samples)]
 
+        # Optional isolated samples carrying source-channel network input at the forecast
+        # times, used only by the latent rollout RMSE diagnostic to encode truth latents.
+        # None on every standard batch; the source/target samples above are never affected.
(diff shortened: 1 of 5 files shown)


clessig:
Why do we need this? How do you distinguish between latent (R)MSE and physical (R)MSE in the training config?

moritzhauschulz:
> Why do we need this? How do you distinguish between latent (R)MSE and physical (R)MSE in the training config?

@clessig We noticed that the problem with analyzing the diffusion training using physical RMSE is heavily dependent on the decoder. Looking at the latent RMSE will give us direct insights on when roll-out diverges in latent space. Afaik there is currently no way of saving the latents for eval, so it needed to be implemented. Note that this is at inference time, not at training time. Training is already via latent loss (as before) unless you put a weight on the physical space – so this is purely a diagnostics tool, and it is triggered via a separate arg in the config (`test_config.latent_rollout_rmse=True`).

PR #2686, opened 2026-07-27
fix fsdp error from unused register token module

Jubeku:
Only create the auxiliary-token parameter when aux tokens are actually used.
With `num_register_tokens == num_class_tokens == 0`, a zero-size parameter still participates in autograd and yields an empty gradient, which makes FSDP2's reduce-scatter fail with `"RuntimeError: _chunk_cat expects non-empty tensor"`.

## Issue Number


clessig:
@shmh40 : can you have a look? `q_aux` comes from develop-ssl.

clessig (commented):
Can we also add this:

```
diff --git a/src/weathergen/model/model_interface.py b/src/weathergen/model/model_interface.py
index d67ceaed..9a52aede 100644
--- a/src/weathergen/model/model_interface.py
+++ b/src/weathergen/model/model_interface.py
@@ -94,7 +94,7 @@ def init_model_and_shard(
             find_unused_parameters=cf.get("ddp_find_unused_parameters", True),
             gradient_as_bucket_view=True,
             bucket_cap_mb=512,
-            static_graph=cf.get("ddp_static_graph", True),
+            static_graph=cf.get("ddp_static_graph", False),
         )
 
     elif with_ddp and with_fsdp:
```

It's related to the PR and causes silent degradation of model convergence. Potentially we should remove it at all as a config parameter.

PR #2704, opened 2026-07-31
Add O96 grid specific lat weighting

### LINKED ISSUE #2635: Add grid specific density based latitude weighting options in loss and evaluation

Current cosine based version estimates line convergence towards the pole. This is not accurate for the number of points in latitude rings for e.g. O96

ToDo: calculate look-up for specific grids based on density or research if there is a known function that describes different grid densities as a function of latitude

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation
- [ ] performance

jpolz:
## Description

Implements O96 specific latitude weighting in loss computation to compensate for the varying point density shown below.

Usage with
```yaml
location_weight: o96_latitude
```

image

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/2635


clessig (commented):
How is this weighting to be used? I am wondering if we need something so specific--O96 is very ECMWF specific. Would it at least work for all O-grids? All reduced Gaussian grids?

jpolz:
> How is this weighting to be used? I am wondering if we need something so specific--O96 is very ECMWF specific. Would it at least work for all O-grids? All reduced Gaussian grids?

I can check how they are built. For usage: you pass the function name in the stream config. instead of `cosine_latitude` use `o96_latitude`.

Update: Good description here: https://confluence.ecmwf.int/plugins/viewsource/viewpagesrc.action?pageId=192923347

I could generalize the function (variable slope of the linear component), but the current way of specifying the function in the stream config would not allow for a parameter (i.e. which O-grid)

clessig:
> > How is this weighting to be used? I am wondering if we need something so specific--O96 is very ECMWF specific. Would it at least work for all O-grids? All reduced Gaussian grids?
> 
> I can check how they are built. For usage: you pass the function name in the stream config. instead of `cosine_latitude` use `o96_latitude`.
> 
> Update: Good description here: https://confluence.ecmwf.int/plugins/viewsource/viewpagesrc.action?pageId=192923347
> 
> I could generalize the function (variable slope of the linear component), but the current way of specifying the function in the stream config would not allow for a parameter (i.e. which O-grid)

In my opinion, `lat_weighting` should automatically try to detect which grid we have and if it can, use the apprropriate weighting, e.g. `o96_latitude`.

jpolz:
> > > How is this weighting to be used? I am wondering if we need something so specific--O96 is very ECMWF specific. Would it at least work for all O-grids? All reduced Gaussian grids?
> > 
> > 
> > I can check how they are built. For usage: you pass the function name in the stream config. instead of `cosine_latitude` use `o96_latitude`.
> > Update: Good description here: https://confluence.ecmwf.int/plugins/viewsource/viewpagesrc.action?pageId=192923347
> > I could generalize the function (variable slope of the linear component), but the current way of specifying the function in the stream config would not allow for a parameter (i.e. which O-grid)
> 
> In my opinion, `lat_weighting` should automatically try to detect which grid we have and if it can, use the apprropriate weighting, e.g. `o96_latitude`.

should be doable. I will revise the PR

clessig:
> > > > How is this weighting to be used? I am wondering if we need something so specific--O96 is very ECMWF specific. Would it at least work for all O-grids? All reduced Gaussian grids?
> > > 
> > > 
> > > I can check how they are built. For usage: you pass the function name in the stream config. instead of `cosine_latitude` use `o96_latitude`.
> > > Update: Good description here: https://confluence.ecmwf.int/plugins/viewsource/viewpagesrc.action?pageId=192923347
> > > I could generalize the function (variable slope of the linear component), but the current way of specifying the function in the stream config would not allow for a parameter (i.e. which O-grid)
> > 
> > 
> > In my opinion, `lat_weighting` should automatically try to detect which grid we have and if it can, use the apprropriate weighting, e.g. `o96_latitude`.
> 
> should be doable. I will revise the PR

Thanks! I would probably use the number of points to detect the respective grid.

jpolz:
> > > > > How is this weighting to be used? I am wondering if we need something so specific--O96 is very ECMWF specific. Would it at least work for all O-grids? All reduced Gaussian grids?
> > > > 
> > > > 
> > > > I can check how they are built. For usage: you pass the function name in the stream config. instead of `cosine_latitude` use `o96_latitude`.
> > > > Update: Good description here: https://confluence.ecmwf.int/plugins/viewsource/viewpagesrc.action?pageId=192923347
> > > > I could generalize the function (variable slope of the linear component), but the current way of specifying the function in the stream config would not allow for a parameter (i.e. which O-grid)
> > > 
> > > 
> > > In my opinion, `lat_weighting` should automatically try to detect which grid we have and if it can, use the apprropriate weighting, e.g. `o96_latitude`.
> > 
> > 
> > should be doable. I will revise the PR
> 
> Thanks! I would probably use the number of points to detect the respective grid.

Done. The current version does not change the behavior from before, but you can set `location_weight : auto` which will use the number of points to detect the O<N>-grid via `#points=4*N*(N+9)`. It does fall back to uniform weighting if the number of points does not fit to any `N in 96, 160, 256, 320, 640, 1280`. Will test now, but happy to change anything about the behavior

clessig:
> > > > > > How is this weighting to be used? I am wondering if we need something so specific--O96 is very ECMWF specific. Would it at least work for all O-grids? All reduced Gaussian grids?
> > > > > 
> > > > > 
> > > > > I can check how they are built. For usage: you pass the function name in the stream config. instead of `cosine_latitude` use `o96_latitude`.
> > > > > Update: Good description here: https://confluence.ecmwf.int/plugins/viewsource/viewpagesrc.action?pageId=192923347
> > > > > I could generalize the function (variable slope of the linear component), but the current way of specifying the function in the stream config would not allow for a parameter (i.e. which O-grid)
> > > > 
> > > > 
> > > > In my opinion, `lat_weighting` should automatically try to detect which grid we have and if it can, use the apprropriate weighting, e.g. `o96_latitude`.
> > > 
> > > 
> > > should be doable. I will revise the PR
> > 
> > 
> > Thanks! I would probably use the number of points to detect the respective grid.
> 
> Done. The current version does not change the behavior from before, but you can set `location_weight : auto` which will use the number of points to detect the O-grid via `#points=4*N*(N+9)`. It does fall back to uniform weighting if the number of points does not fit to any `N in 96, 160, 256, 320, 640, 1280`. Will test now, but happy to change anything about the behavior

Super! What happens for N320? And what were your scientific results for this?

jpolz:
> > > > > > > How is this weighting to be used? I am wondering if we need something so specific--O96 is very ECMWF specific. Would it at least work for all O-grids? All reduced Gaussian grids?
> > > > > > 
> > > > > > 
> > > > > > I can check how they are built. For usage: you pass the function name in the stream config. instead of `cosine_latitude` use `o96_latitude`.
> > > > > > Update: Good description here: https://confluence.ecmwf.int/plugins/viewsource/viewpagesrc.action?pageId=192923347
> > > > > > I could generalize the function (variable slope of the linear component), but the current way of specifying the function in the stream config would not allow for a parameter (i.e. which O-grid)
> > > > > 
> > > > > 
> > > > > In my opinion, `lat_weighting` should automatically try to detect which grid we have and if it can, use the apprropriate weighting, e.g. `o96_latitude`.
> > > > 
> > > > 
> > > > should be doable. I will revise the PR
> > > 
> > > 
> > > Thanks! I would probably use the number of points to detect the respective grid.
> > 
> > 
> > Done. The current version does not change the behavior from before, but you can set `location_weight : auto` which will use the number of points to detect the O-grid via `#points=4*N*(N+9)`. It does fall back to uniform weighting if the number of points does not fit to any `N in 96, 160, 256, 320, 640, 1280`. Will test now, but happy to change anything about the behavior
> 
> Super! What happens for N320? And what were your scientific results for this?

N320 with `location_weight : auto` will be uniform (no weighting). If you keep the configs as they are using `location_weight : cosine_latitude`, then the previous cosine will be used. No weighting is also the default when no `location_weight` is specified. 

Results show minor differences globally with consistent but small benefits of the O-grid specific weighting, but over antarctica one has lower bias, lower RMSE and higher ACC (20 samples Oct 2023). Longer rollout stability to be tested.
image
image
image

clessig:
> > > > > > > > How is this weighting to be used? I am wondering if we need something so specific--O96 is very ECMWF specific. Would it at least work for all O-grids? All reduced Gaussian grids?
> > > > > > > 
> > > > > > > 
> > > > > > > I can check how they are built. For usage: you pass the function name in the stream config. instead of `cosine_latitude` use `o96_latitude`.
> > > > > > > Update: Good description here: https://confluence.ecmwf.int/plugins/viewsource/viewpagesrc.action?pageId=192923347
> > > > > > > I could generalize the function (variable slope of the linear component), but the current way of specifying the function in the stream config would not allow for a parameter (i.e. which O-grid)
> > > > > > 
> > > > > > 
> > > > > > In my opinion, `lat_weighting` should automatically try to detect which grid we have and if it can, use the apprropriate weighting, e.g. `o96_latitude`.
> > > > > 
> > > > > 
> > > > > should be doable. I will revise the PR
> > > > 
> > > > 
> > > > Thanks! I would probably use the number of points to detect the respective grid.
> > > 
> > > 
> > > Done. The current version does not change the behavior from before, but you can set `location_weight : auto` which will use the number of points to detect the O-grid via `#points=4*N*(N+9)`. It does fall back to uniform weighting if the number of points does not fit to any `N in 96, 160, 256, 320, 640, 1280`. Will test now, but happy to change anything about the behavior
> > 
> > 
> > Super! What happens for N320? And what were your scientific results for this?
> 
> N320 with `location_weight : auto` will be uniform (no weighting). If you keep the configs as they are using `location_weight : cosine_latitude`, then the previous cosine will be used. No weighting is also the default when no `location_weight` is specified.
> 
> Results show minor differences globally with consistent but small benefits of the O-grid specific weighting, but over antarctica one has lower bias, lower RMSE and higher ACC (20 samples Oct 2023). Longer rollout stability to be tested. image image image

Why is there no weighting for N320 ? The grid is less uniform than the O-grids so the weighting should be stronger, not?

jpolz:
> Why is there no weighting for N320 ? The grid is less uniform than the O-grids so the weighting should be stronger, not?

N320 should not be detected as an O-grid. it is an original reduced gaussian, not an octahedral one. So in my implementation here it is not detected and, as an unknown grid, the fallback weighting is used. IMO it should use uniform weighting anyways because the points already decrease cosine-like towards the pole:

image
image


[Comparison of the zonal variation in resolution for the N1280 original reduced Gaussian grid (left) with the corresponding O1280 octahedral grid (right) on the surface of the sphere.](https://confluence.ecmwf.int/plugins/viewsource/viewpagesrc.action?pageId=192923347)

src/weathergen/datasets/data_reader_anemoi.py
```diff
@@ -119,6 +119,7 @@ def __init__(
         # caches lats and lons
         self.latitudes = _clip_lat(ds.latitudes)
         self.longitudes = _clip_lon(ds.longitudes)
+        self.n_grid_points = len(self.latitudes)
```
clessig: It should be called `n_latitudes`

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -276,6 +276,7 @@ def _init_stream_datasets(self, cf) -> dict[StreamName, _Stream]:
                 if ds.target_channel_weights is not None
                 else [1.0 for _ in ds.target_channels]
             )
+            stream_info["n_grid_points"] = ds.n_grid_points
```
clessig: This will break for non-anemoi data-readers, which do not set ds.n_grid_points. But I think we should avoid this altogether. The total number of points is available in the loss calculator, through the coords, which we anyway need for computing the weights. So we can just use this, not?

src/weathergen/train/loss_modules/loss_module_physical.py
```diff
@@ -177,7 +177,16 @@ def _get_location_weights(self, stream_info, target_coords, substep_masks):
             return [None for _ in substep_masks]
 
         target_coords = target_coords.to(self.device, non_blocking=True)
-        weights_locations_fct = getattr(loss_fns, location_weight_type)
+
+        if location_weight_type == "auto":
+            n_points = stream_info.get("n_grid_points")
```
clessig: Use target_coords from l. 180 to get the latitudes.

clessig (commented):
Thanks. Some minor changes still needed before we merge to make the code more generic and work for all data readers. Can you please also test with a config with obs.

src/weathergen/datasets/multi_stream_data_sampler.py
jpolz: I think this is in conflict with setting `max_num_targets` to below full grid size. It would be good to get the actual full grid at reader init.. otherwise we can still set an explicit parameter to set in the config (no autodetect)

jpolz:
@clessig the new version still works for anemoi. Any future reader that wants to support O-grid auto-detection just needs to expose n_latitudes. I can make the fallback non-silent. 

Another issue is that with autodetection and even in case of a warning the config would say auto which does not tell you what was actually used

jpolz:
> @clessig the new version still works for anemoi. Any future reader that wants to support O-grid auto-detection just needs to expose n_latitudes. I can make the fallback non-silent.
> 
> Another issue is that with autodetection and even in case of a warning the config would say auto which does not tell you what was actually used

New version resolves actual function used

src/weathergen/datasets/multi_stream_data_sampler.py
clessig: > ct with setting `max_num_targets` to below full grid size. It would be good to get the actual full grid at rea

That's a good point, true.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -276,6 +276,8 @@ def _init_stream_datasets(self, cf) -> dict[StreamName, _Stream]:
                 if ds.target_channel_weights is not None
                 else [1.0 for _ in ds.target_channels]
             )
+            if (n := getattr(ds, "n_latitudes", None)) is not None:
+                stream_info["n_latitudes"] = n
```
clessig, line 280: We should be able to directly amend stream_info in the data_reader.

clessig:
> in case of a warning the config would say auto which does not tell you what was actually used

But we could provide a logger message? (That's semi-useful since we have at the moment too many of them but better than nothing)

PR #2705, opened 2026-07-31
Feature/encoder and decoder healpix cell parallelism

pkuyyj:
## Description

Based on https://github.com/ecmwf/WeatherGenerator/pull/2700, we further implemented the spatial parallelism of the decoder. This is also done during the Hackathon in Zurich.

### Memory allocation profiling results:

* Before:
image

* After encoder parallelization (16 ranks):
image

* After encoder and decoder parallelization (16 ranks):
image

## Issue Number

### DIFF
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -66,9 +66,16 @@ num_register_tokens: 0
 
 healpix_level: 5
 
-# Split the model's HEALPix/location dimension across this many consecutive
-# distributed ranks.
-spatial_parallel_size: 4
+distributed:
+  data_parallel:
+    # Set at runtime from the launched world size.
+    with_ddp: False
+    with_fsdp: True
+    find_unused_parameters: True
+  spatial_parallel:
+    # Split the model's HEALPix/location dimension across this many consecutive
+    # distributed ranks.
+    size: 4
 
 # Use 2D RoPE instead of traditional global positional encoding
 # When True: uses 2D RoPE based on healpix cell coordinates (lat/lon)
@@ -78,7 +85,6 @@ rope_2D: False
 with_mixed_precision: True
 with_flash_attention: True
 compile_model: False
-with_fsdp: True
 attention_dtype: bf16
 mixed_precision_dtype: bf16
 mlp_norm_eps: 1e-5
@@ -112,7 +118,6 @@ general:
   world_size: ???
 
   # local_rank, 
-  # with_ddp,
   # data_path_*, 
   # model_path, 
   # run_path, 
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -66,6 +66,10 @@ num_register_tokens: 0
 
 healpix_level: 5
 
+# Split the encoder's HEALPix/location dimension across this many consecutive
+# distributed ranks through local assimilation.
+encoder_spatial_parallel_size: 4
+
 # Use 2D RoPE instead of traditional global positional encoding
 # When True: uses 2D RoPE based on healpix cell coordinates (lat/lon)
 # When False: uses traditional pe_global positional encoding
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -66,9 +66,9 @@ num_register_tokens: 0
 
 healpix_level: 5
 
-# Split the encoder's HEALPix/location dimension across this many consecutive
-# distributed ranks through local assimilation.
-encoder_spatial_parallel_size: 4
+# Split the model's HEALPix/location dimension across this many consecutive
+# distributed ranks.
+spatial_parallel_size: 4
 
 # Use 2D RoPE instead of traditional global positional encoding
 # When True: uses 2D RoPE based on healpix cell coordinates (lat/lon)
--- a/src/weathergen/datasets/healpix_domain.py
+++ b/src/weathergen/datasets/healpix_domain.py
@@ -29,6 +29,9 @@ def build_local_healpix_cell_splits(
     local_domain_mask = (cell_ids >= cell_start) & (cell_ids < cell_end)
     local_point_idxs = np.flatnonzero(local_domain_mask)
     local_cell_ids = cell_ids[local_point_idxs]
+    cell_splits = [np.array([], dtype=np.int64) for _ in range(cell_end - cell_start)]
+    if local_point_idxs.size == 0:
+        return cell_splits
 
     stable_args = {"stable": True} if int(np.__version__.split(".")[0]) >= 2 else {}
     local_order = np.argsort(local_cell_ids, **stable_args)
@@ -37,7 +40,6 @@ def build_local_healpix_cell_splits(
     split_offsets = np.flatnonzero(np.diff(sorted_cell_ids))
     point_idxs_by_occupied_cell = np.split(sorted_point_idxs, split_offsets + 1)
 
-    cell_splits = [np.array([], dtype=np.int64) for _ in range(cell_end - cell_start)]
     for cell_id, point_idxs in zip(
         np.unique(sorted_cell_ids),
         point_idxs_by_occupied_cell,
--- a/src/weathergen/datasets/healpix_domain.py
+++ b/src/weathergen/datasets/healpix_domain.py
@@ -1,0 +1,48 @@
+# (C) Copyright 2025 WeatherGenerator contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+import numpy as np
+
+
+def build_local_healpix_cell_splits(
+    cell_ids: np.typing.NDArray[np.integer],
+    num_cells: int,
+    cell_start: int,
+    cell_end: int,
+) -> list[np.typing.NDArray[np.int64]]:
+    """Group original point indices for one consecutive HEALPix-cell domain."""
+
+    if not 0 <= cell_start < cell_end <= num_cells:
+        raise ValueError(
+            f"invalid HEALPix cell range [{cell_start}, {cell_end}) for {num_cells} cells"
+        )
+
+    # Domain-parallel filtering mask: this is applied independently to every
+    # stream immediately after its coordinates have been mapped to nested
+    # HEALPix cell IDs.
+    local_domain_mask = (cell_ids >= cell_start) & (cell_ids < cell_end)
+    local_point_idxs = np.flatnonzero(local_domain_mask)
+    local_cell_ids = cell_ids[local_point_idxs]
+
+    stable_args = {"stable": True} if int(np.__version__.split(".")[0]) >= 2 else {}
+    local_order = np.argsort(local_cell_ids, **stable_args)
+    sorted_point_idxs = local_point_idxs[local_order]
+    sorted_cell_ids = local_cell_ids[local_order]
+    split_offsets = np.flatnonzero(np.diff(sorted_cell_ids))
+    point_idxs_by_occupied_cell = np.split(sorted_point_idxs, split_offsets + 1)
+
+    cell_splits = [np.array([], dtype=np.int64) for _ in range(cell_end - cell_start)]
+    for cell_id, point_idxs in zip(
+        np.unique(sorted_cell_ids),
+        point_idxs_by_occupied_cell,
+        strict=True,
+    ):
+        cell_splits[cell_id - cell_start] = point_idxs
+
+    return cell_splits
--- a/src/weathergen/datasets/multi_stream_data_sampler.py
+++ b/src/weathergen/datasets/multi_stream_data_sampler.py
@@ -34,7 +34,7 @@ from weathergen.datasets.utils import (
 )
 from weathergen.readers_extra.registry import get_extra_reader
 from weathergen.train.utils import Stage, get_batch_size_from_config
-from weathergen.utils.distributed import get_encoder_spatial_parallel_size, is_root
+from weathergen.utils.distributed import get_spatial_parallel_size, is_root
 
 type AnyDataReader = DataReaderBase | DataReaderAnemoi | DataReaderObs
 type StreamName = str
@@ -100,13 +100,13 @@ class MultiStreamDataSampler(torch.utils.data.IterableDataset):
 
         self.mini_epoch = 0
         self.mask_value = 0.0
-        # Ranks in one encoder-spatial group must consume the same batch. Data
+        # Ranks in one spatial group must consume the same batch. Data
         # parallelism therefore operates across groups, not across individual ranks.
-        spatial_parallel_size = get_encoder_spatial_parallel_size(cf)
+        spatial_parallel_size = get_spatial_parallel_size(cf)
         self.spatial_parallel_size = spatial_parallel_size
         self.spatial_parallel_rank = cf.rank % spatial_parallel_size
-        self.rank = cf.rank // spatial_parallel_size
-        self.world_size = cf.world_size // spatial_parallel_size
+        self.ddp_rank = cf.rank // spatial_parallel_size
+        self.ddp_world_size = cf.world_size // spatial_parallel_size
         self.repeat_data = cf.data_loading.get("repeat_data_in_mini_epoch", False)
 
         # initialise healpic
@@ -115,7 +115,7 @@ class MultiStreamDataSampler(torch.utils.data.IterableDataset):
         if self.num_healpix_cells % spatial_parallel_size:
             raise ValueError(
                 f"number of HEALPix cells ({self.num_healpix_cells}) must be divisible by "
-                f"encoder_spatial_parallel_size ({spatial_parallel_size})"
+                f"spatial_parallel_size ({spatial_parallel_size})"
             )
         self.local_num_healpix_cells = self.num_healpix_cells // spatial_parallel_size
         self.local_cell_start = self.spatial_parallel_rank * self.local_num_healpix_cells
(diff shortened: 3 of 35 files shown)


config/default_config.yml
```diff
@@ -66,6 +66,10 @@ num_register_tokens: 0
 
 healpix_level: 5
 
+# Split the encoder's HEALPix/location dimension across this many consecutive
+# distributed ranks through local assimilation.
+encoder_spatial_parallel_size: 4
```
clessig, line 71: We should have a section `distributed` where we collect everything related to parallelization, including `with_ddp`, `with_fsdp` etc. Please try to structure it well so that we can pass sub-dicts of it to the relevant parts

clessig: Remove `encoder_` since we will have homogeneous domain decomposition througout the model

src/weathergen/utils/distributed.py
```diff
@@ -12,6 +12,7 @@
 import torch.distributed as dist
 
 SYNC_TIMEOUT_SEC = 60 * 60  # 1 hour
+_ENCODER_SPATIAL_GROUPS: dict[int, tuple[dist.ProcessGroup, int]] = {}
```
clessig, line 15: Remove _ENCODER_

src/weathergen/utils/distributed.py
```diff
@@ -59,6 +60,57 @@ def get_rank() -> int:
     return dist.get_rank()
 
 
+def get_encoder_spatial_parallel_size(cf) -> int:
```
clessig, line 63: Remove _encoder

src/weathergen/datasets/healpix_domain.py
```diff
@@ -0,0 +1,50 @@
...
+
+import numpy as np
+
+
+def build_local_healpix_cell_splits(
+    cell_ids: np.typing.NDArray[np.integer],
+    num_cells: int,
+    cell_start: int,
+    cell_end: int,
+) -> list[np.typing.NDArray[np.int64]]:
```
clessig, line 18: What are the assumptions on `cell_start` and `cell_end`? The goal of the function is to split the data in all healpix cells according to regions?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -100,15 +100,41 @@ def __init__(self, cf: Config, mode_cfg: dict, stage: Stage):
 
         self.mini_epoch = 0
         self.mask_value = 0.0
-        self.rank = cf.rank
-        self.world_size = cf.world_size
+        # Ranks in one encoder-spatial group must consume the same batch. Data
+        # parallelism therefore operates across groups, not across individual ranks.
+        spatial_parallel_size = get_encoder_spatial_parallel_size(cf)
```
clessig, line 105: This information should be organized in a small class that is defined in utils.distributed. This will also be need in model.

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -100,15 +100,41 @@ def __init__(self, cf: Config, mode_cfg: dict, stage: Stage):
...
+        self.spatial_parallel_size = spatial_parallel_size
+        self.spatial_parallel_rank = cf.rank % spatial_parallel_size
+        self.rank = cf.rank // spatial_parallel_size
+        self.world_size = cf.world_size // spatial_parallel_size
```
clessig, line 109: It would probably be best to call this ddp_world_size

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -100,15 +100,41 @@ def __init__(self, cf: Config, mode_cfg: dict, stage: Stage):
...
         # initialise healpic
         self.healpix_level = cf.healpix_level
         self.num_healpix_cells = 12 * 4**self.healpix_level
+        if self.num_healpix_cells % spatial_parallel_size:
+            raise ValueError(
+                f"number of HEALPix cells ({self.num_healpix_cells}) must be divisible by "
+                f"encoder_spatial_parallel_size ({spatial_parallel_size})"
+            )
+        self.local_num_healpix_cells = self.num_healpix_cells // spatial_parallel_size
+        self.local_cell_start = self.spatial_parallel_rank * self.local_num_healpix_cells
```
clessig, line 121: Why does the linear splitting give a contiguous spatial subdivision?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -100,15 +100,41 @@ def __init__(self, cf: Config, mode_cfg: dict, stage: Stage):
...
+        self.local_cell_end = self.local_cell_start + self.local_num_healpix_cells
         self.masker = Masker(cf.healpix_level, stage, cf.streams, self.mode_cfg)
-        self.tokenizer = TokenizerMasking(cf.healpix_level, self.masker)
+        self.tokenizer = TokenizerMasking(
+            cf.healpix_level,
+            self.masker,
+            self.local_cell_start,
```
clessig, line 127: Does the Tokenizer need to know about `self.local_cell_start` and `self.local_cell_end`. It cannot work if we just pass the subset of tokens?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -532,6 +558,7 @@ def _build_stream_data(
             num_steps_input,
             num_output_steps,
             self.num_healpix_cells,
+            source_healpix_cells=self.local_num_healpix_cells,
```
clessig, line 561: Can we rename this to `source_num_healpix_cells`. I find it a bit amgiguous otherwise if the variable contain the data or the number

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -791,7 +820,13 @@ def __iter__(self) -> ModelBatch:
                 # ensure the batch is valid, i.e. not completely empty and no NaN values
                 # student teacher has no classical targets
                 mode = self.mode_cfg.get("training_mode")
-                not_valid = batch.sources_empty() or batch.is_nan()
+                # A valid global sample may have no observations in one rank's
```
clessig, line 823: This is a very good point but I don't see how the code below resolves the problem

src/weathergen/datasets/stream_data.py
```diff
@@ -84,6 +88,9 @@ def __init__(
         self.input_steps = input_steps
         self.output_steps = output_steps
         self.healpix_cells = healpix_cells
+        self.source_healpix_cells = (
```
clessig, line 91: See comments above: `self.source_healpix_cells` -> `self.source_num_healpix_cells`

src/weathergen/datasets/tokenizer_masking.py
```diff
@@ -53,7 +70,7 @@ def reset_rng(self, rng) -> None:
         self.masker.reset_rng(rng)
         self.rng = rng
 
-    def get_tokens_windows(self, stream_info, data, pad_tokens):
+    def get_tokens_windows(self, stream_info, data, pad_tokens, local_source=False):
```
clessig, line 73: Why is this additional flag needed. For `local_source` one could just specify `self.source_cell_start=0` and `self.source_cell_end = total number of healpix cells`

src/weathergen/datasets/tokenizer_masking.py
```diff
@@ -72,7 +91,12 @@ def get_tokens_windows(self, stream_info, data, pad_tokens):
                 continue
             # tokenize data
             idxs_cells, idxs_cells_lens = tok(
-                readerdata_to_torch(rdata), token_size, hl, pad_tokens
+                readerdata_to_torch(rdata),
+                token_size,
+                hl,
+                pad_tokens,
+                cell_start=cell_start,
```
clessig, line 98: Can't we just pass `readerdata_to_torch(rdata)[cell_start : cell_end ]` so that the function doesn't need to know if it operates globally or on a region?

src/weathergen/datasets/tokenizer_utils.py
```diff
@@ -135,7 +144,9 @@ def hpy_splits(
     """
 
     # list of data points per healpix cell
-    (hpy_idxs_ord_split, thetas, phis) = hpy_cell_splits(coords, hl)
+    (hpy_idxs_ord_split, thetas, phis) = hpy_cell_splits(
+        coords, hl, cell_start=cell_start, cell_end=cell_end
```
clessig, line 148: Can't we apply `cell_start` and `cell_end` to coords before calling?

src/weathergen/model/encoder.py
```diff
@@ -43,6 +49,18 @@ def __init__(self, cf: Config, sources_size, targets_num_channels, targets_coord
 
         self.healpix_level = cf.healpix_level
         self.num_healpix_cells = 12 * 4**self.healpix_level
+        self.spatial_parallel_size = get_encoder_spatial_parallel_size(cf)
```
clessig, line 52: See my comment in MultiStreamDataSampler: this should be encapsulated in a small class

src/weathergen/model/encoder.py
```diff
@@ -125,9 +143,35 @@ def forward(self, model_params, batch):
         stream_cell_tokens = checkpoint(
             self.embed_engine, batch, model_params.pe_embed, use_reentrant=False
         )
+        cell_lens = torch.sum(batch.tokens_lens, 2).flatten()
```
clessig, line 146: Can we encapsulate this into a separate function?

src/weathergen/model/encoder.py
```diff
@@ -181,6 +239,20 @@ def assimilate_local_project_chunked(self, tokens, tokens_global, cell_lens, q_c
             # skip processing of the empty chunk in this case
             # Check if this chunk is empty
             if l0 == l1 or toks.shape[0] == 0:
+                # Spatial ranks must enter FSDP-wrapped local modules in the
```
clessig, line 242: Do we need this with torch 2.13? We could merge this first then if it saves us the hassle of introducing fudge code like this

src/weathergen/model/encoder.py
```diff
@@ -287,7 +366,23 @@ def assimilate_local(
             Tokens for global assimilation
         """
 
-        cell_lens = torch.sum(batch.tokens_lens, 2).flatten()
+        tokens_lens_global = batch.tokens_lens
```
clessig, line 369: Can we encapsulate this into a separate function

src/weathergen/model/engines.py
```diff
@@ -115,7 +115,12 @@ def forward(self, batch, pe_embed):
         )
         " Increase ae_local_max_tokens_per_cell in config."
 
-        if batch.tokens_lens.shape[2] == 1:
+        if not x_embeds:
+            # A spatial rank can legitimately own a domain with no observations
```
clessig, line 119: Tighten comment

src/weathergen/model/model.py
```diff
@@ -329,6 +340,12 @@ def __init__(self, cf: Config, sources_size, targets_num_channels, targets_coord
         self.q_cells: torch.Tensor | None = None
         self.streams: dict[str, typing.Any] = cf.streams
         self.target_token_engines = None
+        self.decoder_spatial_parallel_group = None
```
clessig, line 343: Same as above in MultiStreamDataSampler: we want a class for this

src/weathergen/model/model.py
```diff
@@ -376,6 +393,15 @@ def create(self) -> "Model":
         self.encoder = EncoderModule(
             cf, self.sources_size, self.targets_num_channels, self.targets_coords_size
         )
+        # Decoder cells follow the encoder's contiguous spatial-rank ownership.
+        # Keeping the same group also means a cell's nine-token HEALPix
+        # neighbourhood is sent to the rank that already owns that cell.
+        self.decoder_spatial_parallel_group = self.encoder.spatial_parallel_group
```
clessig, line 399: Same as above

src/weathergen/model/model.py
```diff
@@ -782,58 +819,121 @@ def predict_decoders(
...
+                    pred,
+                    pred.new_zeros(
+                        pred.shape[0],
+                        max_len - pred.shape[1],
+                        pred.shape[2],
+                    ),
+                ],
+                dim=1,
+            )
+        gathered_pred = all_gather(pred, group=self.decoder_spatial_parallel_group)
```
clessig, line 933: Anemoi has this function wrapped. Worth to have a discussion of what the pros and cons are.

clessig (commented):
Didn't go through the code in detail but here are some high level comments, in addition to what we already discussed yesterday. For some pieces I probably also need an explanation at some point.

src/weathergen/datasets/multi_stream_data_sampler.py
pkuyyj: Because the nested healpix cell numbering guarantees that low level cells that belongs to one high level cell have contiguous ids. As long as the spatial parallel size divides the number of healpix cells, the linear splitting will take several complete healpix cells at the highest possible level. This figure might be helpful: https://healpix-geo.readthedocs.io/en/stable/healpix/index.html

src/weathergen/datasets/multi_stream_data_sampler.py
pkuyyj: Agreed

src/weathergen/datasets/multi_stream_data_sampler.py
pkuyyj: Yes, the current implementation still needs them because filtering happens inside the tokenizer. I agree that the cleaner design is to filter/remap the observations before tokenization and pass only the rank-local data, allowing the tokenizer to remain unaware of the distributed topology.

src/weathergen/datasets/multi_stream_data_sampler.py
pkuyyj: Agreed

src/weathergen/datasets/multi_stream_data_sampler.py
pkuyyj: The code prevents ranks with an empty local source shard from independently skipping the batch, keeping all spatial ranks synchronized, but it does not distinguish this from a globally empty batch, which is meaningless anyway.

src/weathergen/datasets/stream_data.py
pkuyyj: Agreed.

src/weathergen/datasets/tokenizer_masking.py
pkuyyj: Not directly: cell_start and cell_end are HEALPix cell IDs, whereas ReaderData rows follow the source dataset’s observation order and may be shuffled or combined from multiple readers. We first need to map each row’s coordinates to a HEALPix cell and filter using those IDs. However, I agree that this filtering could be moved before tokenization with the dataloading so the tokenizer only receives rank-local data.

src/weathergen/datasets/tokenizer_masking.py
pkuyyj: The flag distinguishes rank-local source tokenization from global target tokenization because both currently use the same method. For sources, it uses `source_cell_start` and `source_cell_end`; for targets, it uses `[0, num_healpix_cells)`. Separate source and target tokenization methods would make the intended ranges explicit.

src/weathergen/datasets/tokenizer_utils.py
pkuyyj: Not directly, because we need to map coords to healpix cell ids first. The mappings happen inside `hpy_cell_splits`. Otherwise we will do the mapping twice.

src/weathergen/model/encoder.py
pkuyyj: Agreed

src/weathergen/model/encoder.py
pkuyyj: Yes, I'll put them in a function.

src/weathergen/model/engines.py
pkuyyj: Agreed

src/weathergen/model/model.py
pkuyyj: Agreed

src/weathergen/utils/distributed.py
pkuyyj: Agreed

src/weathergen/utils/distributed.py
pkuyyj: Agreed

config/default_config.yml
pkuyyj: Agreed. The corresponding configs are also updated.

src/weathergen/datasets/healpix_domain.py
pkuyyj: The assumptions are based on the nested Healpix cell numbering. Detailed explanation can be found below.

src/weathergen/model/encoder.py
pkuyyj: Agreed. This is a workaround for FSDP2 handling different parameter usage across ranks. Let’s merge the PyTorch 2.13/FSDP2 update first and test the empty-rank case with set_reduce_scatter_unused_params(True). If forward and backward complete correctly, I’ll remove the dummy execution and simply skip empty chunks. We should keep the workaround only if the FSDP collectives still require identical forward ordering across ranks.

src/weathergen/model/encoder.py
pkuyyj: Agreed

src/weathergen/model/model.py
pkuyyj: Agreed

src/weathergen/model/model.py
pkuyyj: Agreed. A wrapper could centralize the process-group handling and hide the tensor-collective details. For now, I would keep the explicit call because this is the only use site and the required operation is clear, but we can revisit this if the pattern appears elsewhere.

PR #2711, opened 2026-08-03
switch for consistently enabling gradient tracking

### LINKED ISSUE #2661: add config arg for inplace scatter

for xai work:

We are using WeatherGenerator for Layer-wise Relevance Propagation (LRP) interpretability work. This requires computing gradients with respect to the ERA5 input tokens, rather than the model weights, so the autograd computation graph needs to remain intact through the forward pass.
While looking at this workflow, we noticed a small autograd-related point in EmbeddingEngine.forward in engines.py that is relevant for interpretability use cases. The current code is:
tokens_all.scatter_(0, scatter_idxs, torch.cat(x_embeds) + pe_embed[pe_idxs])
 
Our understanding is that, because tokens_all is a requires_grad=False tensor, the in-place scatter_ operation does not preserve the autograd graph back to the ERA5 input tokens. The forward-pass values appear to be unchanged; the difference is in whether gradients can be tracked for interpretability workflows.
We have tested a few possible ways of handling this locally, and would welcome your guidance on which approach would best fit the WeatherGenerator codebase.

Agreed upon config flag change

_No response_

### Area

- [x] model
- [x] science
- [ ] documentation
- [ ] performance

enssow:
## Description

In the current WeatherGenerator setup when the model does a forward pass, it runs through `_scatter` in some instances this does not preserve the autograd graph back to the ERA5 input tokens. I've implemented a switch in the config. Will add some sample runs to show that performance of model doesn't change here.

## Issue Number

Closes #2661 

### DIFF
--- a/config/default_config.yml
+++ b/config/default_config.yml
@@ -92,6 +92,8 @@ load_chkpt: {}
 norm_type: "LayerNorm"
 qk_norm_type: null  # if null, defaults to norm_type
 
+lrp_enabled: False
+
 #####################################
 
 streams_directory: "./config/streams/era5_1deg/"
--- a/src/weathergen/model/engines.py
+++ b/src/weathergen/model/engines.py
@@ -123,8 +123,12 @@ class EmbeddingEngine(torch.nn.Module):
             scatter_idxs = self.get_scatter_idxs_vectorized(batch)
             scatter_idxs = scatter_idxs.unsqueeze(1).repeat((1, self.cf.ae_local_dim_embed))
 
-            # actual scatter operation and apply per cell positional encoding
-            tokens_all.scatter_(0, scatter_idxs, torch.cat(x_embeds))
+            # for xai lrp technique 
+            if self.cf.lrp_enabled:
+                tokens_all = tokens_all.scatter(0, scatter_idxs, torch.cat(x_embeds))
+            else:
+                # actual scatter operation and apply per cell positional encoding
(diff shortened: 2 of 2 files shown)


config/default_config.yml
```diff
@@ -92,6 +92,8 @@ load_chkpt: {}
 norm_type: "LayerNorm"
 qk_norm_type: null  # if null, defaults to norm_type
 
+lrp_enabled: False
```
clessig, line 95: The meaning of config options should be self-explanatory, at least for someone familiar with the model. This is not the case here.

src/weathergen/model/engines.py
```diff
@@ -123,8 +123,12 @@ def forward(self, batch, pe_embed):
             scatter_idxs = self.get_scatter_idxs_vectorized(batch)
             scatter_idxs = scatter_idxs.unsqueeze(1).repeat((1, self.cf.ae_local_dim_embed))
 
-            # actual scatter operation and apply per cell positional encoding
-            tokens_all.scatter_(0, scatter_idxs, torch.cat(x_embeds))
+            # for xai lrp technique 
+            if self.cf.lrp_enabled:
+                tokens_all = tokens_all.scatter(0, scatter_idxs, torch.cat(x_embeds))
+            else:
+                # actual scatter operation and apply per cell positional encoding
```
clessig, line 130: The comment should be above the if-statement, the comment in l126 above l128

clessig (commented):
Thanks for the contribution. Minor changes required.

PR #2717, opened 2026-08-03
Fix NaN poisoning from mask-excluded tokens in SSL L1 loss

sophie-xhonneux:
In `_masked_l1_loss_per_sample`, masked-out tokens were zeroed by multiplying with the mask, but NaN * 0 = NaN, so a non-finite value at an excluded position poisoned the whole loss (and its gradients). Fixed by selecting with `torch.where(mask.bool(), per_token_loss, 0.0)` so excluded positions contribute exactly zero. Sanity-tested: NaN outside the mask now yields the expected finite L1 mean, a NaN inside the mask still propagates, and gradients are unchanged for finite inputs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sophie-xhonneux:
Tracking issue: #2723. Also opened against `develop-ssl-diffusion-v1` as #2725 (same commit, cherry-picked).

PR #2718, opened 2026-08-03
Make loss-spike batch skip a global all-reduce decision across ranks

sophie-xhonneux:
Under DDP the loss-spike/NaN skip decision was computed only on rank 0 and broadcast, but losses are rank-local, so a NaN or spike on ranks 1..N-1 with a finite rank-0 loss was never skipped and its gradients poisoned the all-reduce (only the GradScaler silently rejected the step). Every rank now evaluates the spike/NaN check on its local loss (each maintaining its own history) and the flags are combined with an all-reduce MAX (logical OR), so an anomaly on any rank skips the batch on all ranks. Collective-call symmetry is preserved in all code paths and the non-distributed path is unchanged; jsonl logging stays rank-0-only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/weathergen/train/trainer.py
+++ b/src/weathergen/train/trainer.py
@@ -1230,7 +1230,7 @@ class Trainer(TrainerBase):
             skip_flag = torch.tensor(
                 [int(should_skip)], dtype=torch.int32, device=self.device or torch.device("cpu")
             )
-            torch.distributed.broadcast(skip_flag, src=0)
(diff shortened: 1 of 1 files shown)


sophie-xhonneux:
Tracking issue: #2724. Also opened against `develop-ssl-diffusion-v1` as #2726 (same commit, cherry-picked).

Issue #2721, opened 2026-08-04
NaNs in SSL


sophie-xhonneux:
# Description

When running space JEPA, we get NaNs. For example:
```
/e/home/jusers/xhonneux2/jupiter/projects/dev-ssl/WeatherGenerator/src/weathergen/train/trainer.py:1304: RuntimeWarning: Mean of empty slice
  + f"{self.cf.general.istep:06d} : loss = {np.nanmean(avg_loss):.4E} "
000 : 00010/02046 : 000010 : loss = NAN (lr=1.23E-06, s/sec=0.058)
	
LossPhysical.ERA5.mse.avg : 8.4205E-01 	
LossPhysical.loss_avg : 8.4205E-01 	
LossLatentSSLStudentTeacher.loss_avg : nan 	

000 : 00020/02046 : 000020 : loss = NAN (lr=1.83E-06, s/sec=0.088)
	
LossPhysical.ERA5.mse.avg : 7.9669E-01 	
LossPhysical.loss_avg : 7.9669E-01 	
LossLatentSSLStudentTeacher.loss_avg : nan
```

Using Fable to analyse it was found that they come from uninitialized GPU memory: EmbeddingEngine.forward (src/weathergen/model/engines.py:89) allocates tokens_all = torch.empty(num_tokens, dim) and fills it with scatter_, but num_tokens counts tokens from all streams while the embed loop skips streams with an nn.Identity embed — i.e. your target-only, diagnostic ERA5 stream (diagnostic: true / sources_size == 0, engines.py:55-57). In the teacher's target view, that ERA5 stream contributes ~52,000 token rows that are never written and stay as garbage memory. Random bit patterns make ~0.2% of those elements NaN; one NaN is enough — local assimilation spreads it to ~65% of cells and the aggregation engine's global attention to 100% of teacher tokens. The student's model_input view has no ERA5 source tokens (its rows are fully covered), so LossPhysical stays finite while LossLatentSSLStudentTeacher goes NaN. I reproduced and instrumented this end-to-end on a free GH200 on this node — on every teacher forward: num_tokens=108632, nonfinite=113187, bad_rows=52174, skipped=[('ERA5','identity')], while every student forward is clean.

Why the symptoms looked the way they did

- Intermittent ~57% on the cluster: whether the garbage memory happens to contain NaN bit patterns depends on allocator reuse — nondeterministic per batch. On my repro node it was 100%.
[...]

PR #2722, opened 2026-08-04
Fix bugs related to uninitialised GPU memory and istep

### LINKED ISSUE #2720: istep not passed to loss modules
# Description

istep is being passed from trainer.py to loss_calculator.py, but from here it doesn't get passed to the loss modules.
This is a bug, e.g. when using context loss warm_up in the JEPA case as the istep will forever be zero. Even worse it
is silent!

sophie-xhonneux:
## Description

Change torch.empty to torch.zeros in StreamEmbedder to ensure there is no uninitialised GPU memory, also skip streams
with stream embed Identity when considering memory to allocate (#2721)

Fix istep not being passed along in loss_calculator.py (#2720)

Config changes for space JEPA (no need to review)

Remark: Linting is broken, but this was true before.

## Issue Number

Closes #2720 #2721


src/weathergen/model/engines.py
```diff
@@ -1171,14 +1181,19 @@ class DeepSSLFusion(nn.Module):
 
     def __init__(self, num_levels: int, dim_embed: int, hidden_factor: int = 2):
         super().__init__()
+        # Intermediate encoder levels are tapped before the encoder's final LayerNorm, so they
+        # arrive on very different (and much larger) scales than the final level. Normalize each
+        # level before concatenating so the fusion and the downstream SSL head see a common scale.
+        self.level_norms = nn.ModuleList([nn.LayerNorm(dim_embed) for _ in range(num_levels)])
```
clessig, line 1187: This change seems unrelated to the problem that is addressed in the PR?

src/weathergen/model/engines.py
```diff
@@ -1171,14 +1181,19 @@ class DeepSSLFusion(nn.Module):
...
         self.proj = nn.Sequential(
             nn.Linear(num_levels * dim_embed, hidden_factor * dim_embed, bias=False),
             nn.GELU(),
             nn.Linear(hidden_factor * dim_embed, dim_embed, bias=False),
         )
 
     def forward(self, levels: list[torch.Tensor]) -> torch.Tensor:
-        return self.proj(torch.cat(levels, dim=-1))
+        normed = [norm(level) for norm, level in zip(self.level_norms, levels, strict=True)]
```
clessig, line 1195: Same as above

src/weathergen/model/model.py
```diff
@@ -690,10 +690,10 @@ def print_num_parameters(self) -> None:
             np.prod(self.encoder.q_cells.shape) if self.encoder.q_cells.requires_grad else 0
         )
 
-        if self.encoder.q_aux:
-            num_params_q_aux = (
-                np.prod(self.encoder.q_aux.shape) if self.encoder.q_aux.requires_grad else 0
-            )
+        # if self.encoder.q_aux:
```
clessig: This change was introduced to make FSDP work

src/weathergen/model/model.py
```diff
@@ -736,8 +736,8 @@ def print_num_parameters(self) -> None:
         print(f" Local assimilation engine: {num_params_ae_local:,}")
         print(f" Local-global adapter: {num_params_ae_adapter:,}")
         print(f" Learnable spatial queries: {num_params_q_cells:,}")
-        if self.encoder.q_aux:
-            print(f" Learnable auxiliary queries: {num_params_q_aux:,}")
+        # if self.encoder.q_aux:
```
clessig: Same as above

clessig (commented):
Minor comments. This PR seems to have some unrelated changes and would break FSDP again, I think.

config/streams/era5_georing_avhrr_synop_lowres_sjepa/avhrr.yml
```diff
@@ -0,0 +1,86 @@
+# (C) Copyright 2024 WeatherGenerator contributors.
```
shmh40, line 1: Configs

src/weathergen/model/model.py
shmh40: Need to revert this I think

src/weathergen/train/loss_modules/loss_module_ssl.py
```diff
@@ -321,7 +321,10 @@ def _masked_l1_loss_per_sample(
     mask_f = mask.to(per_token_loss.dtype)
     token_counts = mask_f.sum(dim=-1)
     valid_samples = token_counts > 0
-    per_sample_loss = (per_token_loss * mask_f).sum(dim=-1) / token_counts.clamp(min=1.0)
+    # torch.where (not multiply): NaN at mask-excluded positions must not poison the sum.
```
shmh40, line 324: Can merge this PR in last so it is after the PR that explicitly does this

src/weathergen/model/model.py
sophie-xhonneux: this broke in single GPU mode for me :/ can you explain the error to me?

sophie-xhonneux: see my comment below, the statement breaks in single GPU mode if q_aux exists because it is a nn.Parameter tensor so not bool or can be reduced to bool

shmh40: Here is the PR:

https://github.com/ecmwf/WeatherGenerator/pull/2686

sophie-xhonneux: we are in the function that prints parameters, does not seem like it will break FSDP

src/weathergen/model/model.py
sophie-xhonneux: we are printing parameters, this should not break FSDP, it doesn't create the parameters

src/weathergen/model/model.py
shmh40: Yes I think you're right it is not fsdp but whether it breaks I think depends if you have register tokens or not right? 

You do have register tokens (I guess), so now it works. But if we want to run with 0 register tokens with this code, we are now doing shape of None and we get an error?

shmh40: So I think we should actually have: 

`
if self.encoder.q_aux is not None:
`

sophie-xhonneux: yes! you are right, I couldn't think of the right check at midnight yesterday so I commented it out, but what you right is correct I think

packages/common/src/weathergen/common/config.py
```diff
@@ -140,7 +140,7 @@ def _strip_interpolation(conf: Config) -> Config:
     stripped = {}
     if OmegaConf.is_dict(conf):
         for key in list(conf.keys()):
-            key = str(key)
+            # key = str(key) # this lines breaks target_source_correspondence please stop re introducing it
```
shmh40: Do we need this for anything else in the config?

I get around it with e.g. `target_source_correspondence: {"0": {"0": independent}}` in target source I think, in case it is necessary for something else in the config

sophie-xhonneux: I would ask the question the other way around, is there any reason this line is needed, it breaks the type assumptions of YAML

clessig (commented):
Can we clean up the PR so that we don't change 20 files unrelated to the problems. In engines.py are also changes that are unrelated?

config/streams/era5_georing_avhrr_synop_lowres_sjepa/avhrr.yml
clessig: They should go to a separate repository

clessig (approved):

PR #2727, opened 2026-08-04
[2712] Time-aggregation loss

### LINKED ISSUE #2712: Task: Add time-aggregated loss for physical time evolution

Producing outputs at a higher temporal frequency than the model’s time-window length now appears to work reasonably well. However, the resulting trajectories sometimes contain undesirable and physically unrealistic jumps within each time window.

This behaviour was observed when using a 6-hour time window with hourly outputs:

Image

A likely explanation is that the output states close to the 6-hour boundaries are more strongly constrained than those near the centre of the window. Consequently, the intermediate states may be less predictable and may evolve inconsistently.

A related problem is discussed in the recent HourGlass preprint, [arXiv:2607.11457](https://arxiv.org/pdf/2607.11457). The paper encourages temporal consistency by applying the training loss not only to the individual output states, but also to temporal statistics and consecutive differences.

https://gitlab.jsc.fz-juelich.de/hedgedoc/s/tzqiMi8vQ

### Area

- [x] model
- [x] science
- [ ] documentation
- [ ] performance

evenmn:
## Description

Time-aggregation loss to avoid jumps within a time window when frequency < window. The ideas are taken from [arxiv:2607.11457](https://arxiv.org/pdf/2607.11457), and the implementation is highly inspired by the Anemoi implementation:

https://github.com/ecmwf/anemoi-core/blob/main/training/src/anemoi/training/losses/aggregate.py

### Invoking the loss

``` bash
train --config ... --options training_config.losses.physical.loss_fcts.mse.time_aggregation_types=[mean,max,min,diff]
```

Screenshot from 2026-08-06 15-14-20

## Issue Number

Closes #2712


clessig (commented):
Before looking at details, I have some high-level questions: why is `_record_time_agg_candidate()` called in the inner loop of the loss computation. I would have expected that this happens in the contructor of `LossPhysical`. Also, could you give an example for how the config looks like. This doens't seem to be commited?

evenmn:
> Before looking at details, I have some high-level questions: why is `_record_time_agg_candidate()` called in the inner loop of the loss computation. I would have expected that this happens in the contructor of `LossPhysical`. Also, could you give an example for how the config looks like. This doens't seem to be commited?

`_record_time_agg_candidate()` checks if each batch is a candidate for time-aggregation loss. Masking and other things may disable the time-aggregation loss. The function name is a little confusing though, should I replace `record` with something more intuitive? 

This is how to enable the time-aggregation loss. I can push `config_forecasting.yml` adn the stream config as well, it's based on the `atmofs05_6x1h` branch

```
train --config ... --options training_config.losses.physical.loss_fcts.mse.time_aggregation_types=[mean,max,min,diff]
```

PR #2765, opened 2026-08-14
Sophiex/dev ssl diffusion/make write target optional

sophie-xhonneux:
## Description

Don't write the target to the zarrs and drop the target channels as soon as possible. From my and Claude's understanding of anemoi zarr reading, we still need the bulk dataread to get the geoinfos.

Note this is only useful once we have the evaluation package able to read the target zarrs from predictions alone.

## Issue Number

- [x] tested: train+ validation space JEPA
- [x] tested: train+ validation temporal JEPA
- [ ] tested: train+ validation diffusion
- [ ] tested: train+ validation forecasting

### DIFF
--- a/packages/common/src/weathergen/common/io.py
+++ b/packages/common/src/weathergen/common/io.py
@@ -597,6 +597,10 @@ class OutputBatchData:
     forecast_offset: int
     forecast_steps: list[int]
 
+    # if False, no target datasets are built/written (skip_target_values inference);
+    # self.targets then only provides row counts via targets_lens
+    write_targets: bool = True
+
     @functools.cached_property
     def samples(self):
         """Continous indices of all samples accross all batches."""
@@ -686,20 +690,26 @@ class OutputBatchData:
                 datapoints
             ]
 
-        assert len(data_coords.channels) == target_data.shape[1], (
-            "Number of channel names does not align with target data."
-        )
         assert len(data_coords.channels) == preds_data.shape[1], (
             "Number of channel names does not align with prediction data."
         )
 
-        target_dataset = OutputDataset(
-            "target",
-            key,
-            source_interval,
-            target_data,
-            **dataclasses.asdict(data_coords),
-        )
+        if self.write_targets:
+            assert len(data_coords.channels) == target_data.shape[1], (
+                "Number of channel names does not align with target data."
+            )
+            target_dataset = OutputDataset(
+                "target",
+                key,
+                source_interval,
+                target_data,
+                **dataclasses.asdict(data_coords),
+            )
+        else:
+            # target values were never read (skip_target_values); coords/times are
+            # still written as part of the prediction dataset
+            target_dataset = None
+
         prediction_dataset = OutputDataset(
             "prediction",
             key,
--- a/src/weathergen/datasets/data_reader_base.py
+++ b/src/weathergen/datasets/data_reader_base.py
@@ -389,7 +389,7 @@ class DataReaderBase(metaclass=ABCMeta):
 
         return rdata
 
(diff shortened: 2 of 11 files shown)


packages/common/src/weathergen/common/io.py
```diff
@@ -597,6 +597,10 @@ class OutputBatchData:
     forecast_offset: int
     forecast_steps: list[int]
 
+    # if False, no target datasets are built/written (skip_target_values inference);
+    # self.targets then only provides row counts via targets_lens
+    write_targets: bool = True
```
clessig, line 602: Can we put this under validation_config.output.write_targets

src/weathergen/datasets/data_reader_base.py
```diff
@@ -389,24 +389,44 @@ def get_source(self, idx: TIndex) -> ReaderData:
 
         return rdata
 
-    def get_target(self, idx: TIndex) -> ReaderData:
+    def get_target(self, idx: TIndex, coords_geoinfos_only: bool = False) -> ReaderData:
```
clessig, line 392: The flag is a bit misleading since coords, geoinfos, and time are provided. Maybe call is `no_data` or similar?

src/weathergen/datasets/data_reader_obs.py
```diff
@@ -299,3 +305,12 @@ def _get(self, idx: int, channels_idx: list[int]) -> ReaderData:
         check_reader_data(rdata, dtr)
 
         return rdata
+
+    @override
+    def _get_coords_geoinfos_only(self, idx: int, channels_idx: list[int]) -> ReaderData:
```
clessig, line 310: Why do we need the override here? Wouldn't the implementation in the base class work?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -50,11 +55,15 @@
 }
 
 
-def collect_datasources(stream_datasets: list, idx: int, type: str, rng) -> IOReaderData:
+def collect_datasources(
+    stream_datasets: list, idx: int, type: str, rng, coords_geoinfos_only: bool = False
```
clessig, line 59: Same about arg name as above

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -137,6 +150,17 @@ def __init__(self, cf: Config, mode_cfg: dict, stage: Stage):
                 )
             self.teacher_time_offset = 0
 
+        # skip_target_values: inference-only; do not read/store target values for physically
+        # reconstructed streams, only coords/geoinfos/datetimes (data gets zero width)
+        self.skip_target_values = bool(mode_cfg.get("skip_target_values", False))
```
clessig, line 155: Why do you have a flag in `io.py` and a config argument? As above, the config argument should be `validation_config.output.write_targets`

clessig (commented):
I wouldn't merge right now. There are quite a few changes in places that also touch training. It would have to be tested properly with various configs, which is too big a distraction right now.

src/weathergen/datasets/data_reader_obs.py
sophie-xhonneux: I think it has to do with the fact that different readers have different capability to only get the geoinfos, e.g. in anemoi it isn't possible, you can only drop them at the end of the stack frame

src/weathergen/datasets/multi_stream_data_sampler.py
sophie-xhonneux: sure, idm

PR #2768, opened 2026-08-18
add additional column case in plot_training

### LINKED ISSUE #2767: Update plot_training.py

The EERIE dataset has channel names with "." such as `avg_theta_2.5`. This case is not covered in the `plot_training.py`.

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation
- [ ] performance

SavvasMel:
## Description

Introduces a small "if statement" to cover channels with "." in the last parts of the string. This allow for:

image

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/2767

#### FastEvaluation

### DIFF
--- a/src/weathergen/utils/plot_training.py
+++ b/src/weathergen/utils/plot_training.py
@@ -469,6 +469,15 @@ def plot_loss_per_stream(
                                 ):
                                     data_cols += [col]
                                     title_col = col if title_col is None else title_col
+                            elif len(col_split) == 6:
+                                if (
+                                    col_split[1].lower() == stream_name.lower()
+                                    and col_split[2].lower() == err.lower()
+                                    and col_split[3] + "." + col_split[4] == channel
+                                    and int(col_split[5]) in forecast_steps
+                                ):
+                                    data_cols += [col]
+                                    title_col = col if title_col is None else title_col
 
                         for col in data_cols:
                             x_vals = np.array(run_data_mode[x_col])
--- a/src/weathergen/utils/plot_training.py
+++ b/src/weathergen/utils/plot_training.py
@@ -21,7 +21,7 @@ import yaml
 
 import weathergen.common.config as config
 from weathergen.train.utils import TRAIN
-from weathergen.utils.train_logger import Metrics, TrainLogger
+from weathergen.utils.train_logger import Metrics, TrainLogger, sanitize_channel_for_matching
 
 _logger = logging.getLogger(__name__)
 
@@ -420,8 +420,17 @@ def plot_loss_per_stream(
     prop_cycle = plt.rcParams["axes.prop_cycle"]
     colors = prop_cycle.by_key()["color"] + ["r", "g", "b", "k", "m", "y"]
 
+    if channels is not None:
+        channels_sanitized = [sanitize_channel_for_matching(c) for c in channels]
+    else:
+        channels_sanitized = channels
+
     for err in errs:
-        for channel in channels:
+        channels_to_use = channels if channels else [None]
+        channels_sanitized_to_use = channels_sanitized if channels_sanitized else [None]
+        for channel, channel_sanitized in zip(
+            channels_to_use, channels_sanitized_to_use, strict=True
+        ):
             for stream_name in stream_names:
                 _fig = plt.figure(figsize=(10, 7), dpi=PLOT_DPI_VALUE)
 
@@ -456,7 +465,7 @@ def plot_loss_per_stream(
                                 if (
                                     col_split[1].lower() == stream_name.lower()
                                     and col_split[2].lower() == err.lower()
-                                    and col_split[3] == channel
+                                    and col_split[3] == channel_sanitized
                                 ):
                                     data_cols += [col]
                                     title_col = col if title_col is None else title_col
@@ -464,17 +473,8 @@ def plot_loss_per_stream(
                                 if (
                                     col_split[1].lower() == stream_name.lower()
                                     and col_split[2].lower() == err.lower()
-                                    and col_split[3] == channel
+                                    and col_split[3] == channel_sanitized
                                     and int(col_split[4]) in forecast_steps
-                                ):
-                                    data_cols += [col]
-                                    title_col = col if title_col is None else title_col
-                            elif len(col_split) == 6:
-                                if (
-                                    col_split[1].lower() == stream_name.lower()
-                                    and col_split[2].lower() == err.lower()
-                                    and col_split[3] + "." + col_split[4] == channel
-                                    and int(col_split[5]) in forecast_steps
                                 ):
                                     data_cols += [col]
                                     title_col = col if title_col is None else title_col
@@ -612,6 +612,11 @@ def plot_loss_per_run(
     if errs is None:
         errs = ["mse"]
 
+    if channels is not None:
+        channels_sanitized = [sanitize_channel_for_matching(c) for c in channels]
+    else:
+        channels_sanitized = channels
+
     plot_dir = Path(plot_dir)
 
     modes = [modes] if type(modes) is not list else modes
@@ -640,7 +645,7 @@ def plot_loss_per_run(
                 if (
                     len(col_split) >= 4
                     and col_split[2].lower() == err.lower()
-                    and col_split[3] in channels
+                    and col_split[3] in channels_sanitized
                 ):
                     data_cols += [col]
 
--- a/src/weathergen/utils/train_logger.py
+++ b/src/weathergen/utils/train_logger.py
@@ -19,7 +19,6 @@ from pathlib import Path
 import numpy as np
 import polars as pl
 import torch
-
 import weathergen.common.config as config
 
 # from weathergen.train.trainer import cfg_keys_to_filter
@@ -62,7 +61,9 @@ class TrainLogger:
         self.cf = cf
         self.path_run = path_run
 
-    def log_metrics(self, stage: Stage, metrics: dict[str, float], step: int | None = None) -> None:
+    def log_metrics(
+        self, stage: Stage, metrics: dict[str, float], step: int | None = None
+    ) -> None:
         """
         Log metrics to a file.
         For now, just scalar values are expected. There is no check.
@@ -145,7 +146,9 @@ class TrainLogger:
 
         # Load config from given model_path if provided, otherwise use path from private config
         if model_path:
-            cf = config.load_run_config(run_id=run_id, mini_epoch=mini_epoch, model_path=model_path)
+            cf = config.load_run_config(
+                run_id=run_id, mini_epoch=mini_epoch, model_path=model_path
+            )
         else:
             cf = config.load_merge_configs(
                 private_home=None, from_run_id=run_id, mini_epoch=mini_epoch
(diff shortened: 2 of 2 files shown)


clessig (commented):
I don't have a strong opinion but wouldn't it be best to sanitize the channel names when the data is loaded and, e.g., replace "." with "_"?

jpolz (approved):
Looks good!

SavvasMel:
> I don't have a strong opinion but wouldn't it be best to sanitize the channel names when the data is loaded and, e.g., replace "." with "_"?

The data has been transferred to `ml_training` folder so someone with write/read access rights needs to this, and also the sanitation needs to take place also for all the v3 zarr files. It is not a challenge or anything per se to have a script to do this automatically, or something that an agent cannot do fast and clean, but I find it time consuming, even only the effort on pinging people just for a ".".

Also to be fair I do not really find self explained having `2.5` as `2_5`. What do you think?

clessig:
> > I don't have a strong opinion but wouldn't it be best to sanitize the channel names when the data is loaded and, e.g., replace "." with "_"?
> 
> The data has been transferred to `ml_training` folder so someone with write/read access rights needs to this, and also the sanitation needs to take place also for all the v3 zarr files. It is not a challenge or anything per se to have a script to do this automatically, or something that an agent cannot do fast and clean, but I find it time consuming, even only the effort on pinging people just for a ".".
> 
> Also to be fair I do not really find self explained having `2.5` as `2_5`. What do you think?

No, I mean when we read in the training log files we should santize the column name to a valid format.

SavvasMel:
> > > I don't have a strong opinion but wouldn't it be best to sanitize the channel names when the data is loaded and, e.g., replace "." with "_"?
> > 
> > 
> > The data has been transferred to `ml_training` folder so someone with write/read access rights needs to this, and also the sanitation needs to take place also for all the v3 zarr files. It is not a challenge or anything per se to have a script to do this automatically, or something that an agent cannot do fast and clean, but I find it time consuming, even only the effort on pinging people just for a ".".
> > Also to be fair I do not really find self explained having `2.5` as `2_5`. What do you think?
> 
> No, I mean when we read in the training log files we should santize the column name to a valid format.

I tried, if you sanitize the column names then the legends are affected necessarily:

image

Also you cannot avoid but to have something of the sort to correct the titles. I believe sanitizing is much more prune to bugs/errors in the future.

```
                            elif len(col_split) == 4:
                                if (
                                    col_split[1].lower() == stream_name.lower()
                                    and col_split[2].lower() == err.lower()
                                    and col_split[3] == channel_sanitized
                                ):
                                    data_cols += [col]
                                    title_col = (
                                        f"LossPhysical.{stream_name}.{err}.{channel_display}"                 <---------
                                        if title_col is None
                                        else title_col
                                    )
                            elif len(col_split) == 5:
                                if (
                                    col_split[1].lower() == stream_name.lower()
                                    and col_split[2].lower() == err.lower()
                                    and col_split[3] == channel_sanitized
                                    and int(col_split[4]) in forecast_steps
                                ):
                                    data_cols += [col]
                                    title_col = (
                                        f"LossPhysical.{stream_name}.{err}.{channel_display}"               <---------
                                        if title_col is None
                                        else title_col
                                    )
```

clessig:
> > > > I don't have a strong opinion but wouldn't it be best to sanitize the channel names when the data is loaded and, e.g., replace "." with "_"?
> > > 
> > > 
> > > The data has been transferred to `ml_training` folder so someone with write/read access rights needs to this, and also the sanitation needs to take place also for all the v3 zarr files. It is not a challenge or anything per se to have a script to do this automatically, or something that an agent cannot do fast and clean, but I find it time consuming, even only the effort on pinging people just for a ".".
> > > Also to be fair I do not really find self explained having `2.5` as `2_5`. What do you think?
> > 
> > 
> > No, I mean when we read in the training log files we should santize the column name to a valid format.
> 
> I tried, if you sanitize the column names then the legends are affected necessarily:
> 
> image
> Also you cannot avoid but to have something of the sort to correct the titles. I believe sanitizing is much more prune to bugs/errors in the future.
> 
> ```
>                             elif len(col_split) == 4:
>                                 if (
>                                     col_split[1].lower() == stream_name.lower()
>                                     and col_split[2].lower() == err.lower()
>                                     and col_split[3] == channel_sanitized
>                                 ):
>                                     data_cols += [col]
>                                     title_col = (
>                                         f"LossPhysical.{stream_name}.{err}.{channel_display}"                 <---------
>                                         if title_col is None
>                                         else title_col
>                                     )
>                             elif len(col_split) == 5:
>                                 if (
>                                     col_split[1].lower() == stream_name.lower()
>                                     and col_split[2].lower() == err.lower()
>                                     and col_split[3] == channel_sanitized
>                                     and int(col_split[4]) in forecast_steps
>                                 ):
>                                     data_cols += [col]
>                                     title_col = (
>                                         f"LossPhysical.{stream_name}.{err}.{channel_display}"               <---------
>                                         if title_col is None
>                                         else title_col
>                                     )
> ```

But is it an issue to have `...thetao_2_5` instead of `...thetao_2.5` ? plot_train is an absolute expert system and the 3 people that would be affected would know, not?

SavvasMel:
> > > > > I don't have a strong opinion but wouldn't it be best to sanitize the channel names when the data is loaded and, e.g., replace "." with "_"?
> > > > 
> > > > 
> > > > The data has been transferred to `ml_training` folder so someone with write/read access rights needs to this, and also the sanitation needs to take place also for all the v3 zarr files. It is not a challenge or anything per se to have a script to do this automatically, or something that an agent cannot do fast and clean, but I find it time consuming, even only the effort on pinging people just for a ".".
> > > > Also to be fair I do not really find self explained having `2.5` as `2_5`. What do you think?
> > > 
> > > 
> > > No, I mean when we read in the training log files we should santize the column name to a valid format.
> > 
> > 
> > I tried, if you sanitize the column names then the legends are affected necessarily:
> > image
> > Also you cannot avoid but to have something of the sort to correct the titles. I believe sanitizing is much more prune to bugs/errors in the future.
> > ```
> >                             elif len(col_split) == 4:
> >                                 if (
> >                                     col_split[1].lower() == stream_name.lower()
> >                                     and col_split[2].lower() == err.lower()
> >                                     and col_split[3] == channel_sanitized
> >                                 ):
> >                                     data_cols += [col]
> >                                     title_col = (
> >                                         f"LossPhysical.{stream_name}.{err}.{channel_display}"                 <---------
> >                                         if title_col is None
> >                                         else title_col
> >                                     )
> >                             elif len(col_split) == 5:
> >                                 if (
> >                                     col_split[1].lower() == stream_name.lower()
> >                                     and col_split[2].lower() == err.lower()
> >                                     and col_split[3] == channel_sanitized
> >                                     and int(col_split[4]) in forecast_steps
> >                                 ):
> >                                     data_cols += [col]
> >                                     title_col = (
> >                                         f"LossPhysical.{stream_name}.{err}.{channel_display}"               <---------
> >                                         if title_col is None
> >                                         else title_col
> >                                     )
> > ```
> 
> But is it an issue to have `...thetao_2_5` instead of `...thetao_2.5` ? plot_train is an absolute expert system and the 3 people that would be affected would know, not?

Hello, I understand, I committed the second solution, that sanitizes the logging data. I used the term "sanitize" in the functions but it might be that another term would be better. Let me know if this meets your request better :)

SavvasMel:
Are we still interested for this to be merged or should I close the issue?

TillHae:
@SavvasMel @clessig I stumbled across this, because the issue was still open. Maybe the best idea would be to just fix how we select `data_cols` and handle `title_col` completely. 

If you look at [how it is currently implemented](https://github.com/ecmwf/WeatherGenerator/blob/develop/src/weathergen/utils/plot_training.py#L449-L471), we try to reverse-engineer column names by splitting on `.` and checking `len(col_split) < 4`, `== 4`, or `== 5`. That inevitably fails whenever a channel name contains a dot (like `avg_thetao_2.5`).

Instead of reverse-engineering with string splits, we already know the exact `stream_name`, `err`, and `channel` we are looking for from the loops. We can just construct the expected column string directly (e.g. `...{stream_name}.{err}.{channel}`) and check whether the column ends with that suffix directly or with an explicit forecast step appended.

In general, if you take a deeper look at how weird the entire handling of `title_col` is, we don't even need it. It only exists to chop off the forecast step on line 525 with `.split(".")[:-1]` to set the chart title. This is something we could already set directly from our loop variables without any of that.

We should probably think about cleaning up this entire 5-nested-for-loop *"architecture"*, but doing a direct string match instead of splitting on dots should immediately fix the bug without touching `train_logger.py` or renaming columns in the DataFrame.

What do you think?

clessig:
> @SavvasMel @clessig I stumbled across this, because the issue was still open. Maybe the best idea would be to just fix how we select `data_cols` and handle `title_col` completely.
> 
> If you look at [how it is currently implemented](https://github.com/ecmwf/WeatherGenerator/blob/develop/src/weathergen/utils/plot_training.py#L449-L471), we try to reverse-engineer column names by splitting on `.` and checking `len(col_split) < 4`, `== 4`, or `== 5`. That inevitably fails whenever a channel name contains a dot (like `avg_thetao_2.5`).
> 
> Instead of reverse-engineering with string splits, we already know the exact `stream_name`, `err`, and `channel` we are looking for from the loops. We can just construct the expected column string directly (e.g. `...{stream_name}.{err}.{channel}`) and check whether the column ends with that suffix directly or with an explicit forecast step appended.
> 
> In general, if you take a deeper look at how weird the entire handling of `title_col` is, we don't even need it. It only exists to chop off the forecast step on line 525 with `.split(".")[:-1]` to set the chart title. This is something we could already set directly from our loop variables without any of that.
> 
> We should probably think about cleaning up this entire 5-nested-for-loop _"architecture"_, but doing a direct string match instead of splitting on dots should immediately fix the bug without touching `train_logger.py` or renaming columns in the DataFrame.
> 
> What do you think?

@TillHae  I agree that the problem is encoding too many things in strings, without having properly thought about a convention that avoids collisions. But we shouldn't try to solve this now. 

It sounds like you have a clear idea on how to set up a PR that constructs the col names from the available info and that avoids the problem. Could you implement it?

TillHae:
> > @SavvasMel @clessig I stumbled across this, because the issue was still open. Maybe the best idea would be to just fix how we select `data_cols` and handle `title_col` completely.
> 
> > 
> 
> > If you look at [how it is currently implemented](https://github.com/ecmwf/WeatherGenerator/blob/develop/src/weathergen/utils/plot_training.py#L449-L471), we try to reverse-engineer column names by splitting on `.` and checking `len(col_split) < 4`, `== 4`, or `== 5`. That inevitably fails whenever a channel name contains a dot (like `avg_thetao_2.5`).
> 
> > 
> 
> > Instead of reverse-engineering with string splits, we already know the exact `stream_name`, `err`, and `channel` we are looking for from the loops. We can just construct the expected column string directly (e.g. `...{stream_name}.{err}.{channel}`) and check whether the column ends with that suffix directly or with an explicit forecast step appended.
> 
> > 
> 
> > In general, if you take a deeper look at how weird the entire handling of `title_col` is, we don't even need it. It only exists to chop off the forecast step on line 525 with `.split(".")[:-1]` to set the chart title. This is something we could already set directly from our loop variables without any of that.
> 
> > 
> 
> > We should probably think about cleaning up this entire 5-nested-for-loop _"architecture"_, but doing a direct string match instead of splitting on dots should immediately fix the bug without touching `train_logger.py` or renaming columns in the DataFrame.
> 
> > 
> 
> > What do you think?
> 
> 
> 
> @TillHae  I agree that the problem is encoding too many things in strings, without having properly thought about a convention that avoids collisions. But we shouldn't try to solve this now. 
> 
> 
> 
> It sounds like you have a clear idea on how to set up a PR that constructs the col names from the available info and that avoids the problem. Could you implement it?

Yes, I will implement something in the upcoming days.

PR #2773, opened 2026-08-23
Enable 10-day inference without targets

clessig:
## Description

Enable 10-day inference without targets using a new data reader (anemoi-rt) and chunked inference (based on work by @evenmn @sophie-xhonneux @shmh40).

Inference command (run_id on JUPITER): 

```
uv run --offline inference --from-run-id mz7oyrr5 --options streams.ERA5.max_num_targets=-1 streams.ERA5.type="anemoi_rt" streams.ERA5.frequency=01:00:00 test_config.samples_per_mini_epoch=1 test_config.forecast.num_steps=40 test_config.forecast.chunk_size=1 test_config.compute_loss=False data_loading.num_workers=0 data_loading.num_workers_validation=0
```

## Issue Number

-

#### FastEvaluation


packages/readers_extra/src/weathergen/readers_extra/data_reader_anemoi_rt.py
```diff
@@ -0,0 +1,352 @@
...
+        (
+            self.geoinfo_idx_static_lin,
+            self.geoinfo_idx_dynamic_lin,
+        ) = [], []
+        idx = 0
+        for _, (k, v) in enumerate(ds.typed_variables.items()):
+            if k not in self.geoinfo_channels:
+                continue
+
+            if v.is_constant_in_time:
```
evenmn: In many of our anemoi datasets the static fields are not labelled as constant in time as they are repeated for every time frame. If a field is not constant in time, can we still take the first frame and raise a warning? We changed this to make your code work with `lsm` and `z` (should probably raise a proper warning instead of a print statement):

```python
if k in _anemoi_dynamic_forcings():
    self.geoinfo_channels_dynamic += [k]
    self.geoinfo_idx_dynamic += [ds.variables.index(k)]
    self.geoinfo_idx_dynamic_lin += [idx]
else:
    if not v.is_constant_in_time:
        print(f"Warning: variable '{k}' is not constant in time")
    self.geoinfo_channels_static += [k]
    self.geoinfo_idx_static += [ds.variables.index(k)]
    self.geoinfo_idx_static_lin += [idx]
```

clessig: We should use `_logger.warning`. Yes, we should still use the first frame to get all the data. 

We need to have a way to find out which geoinfo channels should come from file and which are generated on the fly. Maybe we should turn around the current logic: match what can be generated on the fly from `_anemoi_dynamic_forcings()` and then try to match the rest from the file, with a warning if it's not labelled as static and failing if the channel is not present at all in the file

src/weathergen/train/trainer.py
```diff
@@ -323,7 +454,9 @@ def run(self, cf, devices, run_id_contd=None, mini_epoch_contd=None):
         # https://www.cs.princeton.edu/~smalladi/blog/2024/01/22/SDEs-ScalingRules/
         # aiming for beta1=0.9 and beta2=0.95 following the MAE paper
         # https://arxiv.org/pdf/2111.06377
-        kappa = self.get_batch_size_total(self.batch_size_per_gpu)
+        # TODO
```
shmh40, line 457: Should this be here? Are we happy with 4?

src/weathergen/utils/validation_io.py
```diff
@@ -95,7 +140,7 @@ def write_output(
                     t_times_s += [t_times.astype("datetime64[ns]")]
 
             targets_lens[-1] += [[]]
-            targets_lens[-1][-1] += [t.shape[0] for t in targets_s]
+            targets_lens[-1][-1] += [t.shape[0] for t in preds_s]
```
shmh40, line 143: This needs to be preds now, makes sense. But preds has ensemble dimension as dim 0 right? So needs t.shape[1] I think?

packages/readers_extra/src/weathergen/readers_extra/data_reader_anemoi_rt.py
```diff
@@ -0,0 +1,352 @@
...
+
+        # use time_window and frequency to compute required time information
+        datetimes = []
+        t_cur = dtr.start
+        while t_cur < dtr.end:
+            datetimes += [t_cur]
+            t_cur += self.frequency
+
+        # extract geoinfo channels (can be time-varying, so read from dataset)
+        geoinfos_static = np.repeat(
```
shmh40, line 201: Is this the right repeat? Or we should tile?

       geoinfos_static = np.tile(
           self.ds[0, list(self.geoinfo_idx_static)][0], (1, len(t_idxs))
       )

and then we can have strict=True below?

src/weathergen/train/trainer.py
```diff
@@ -617,31 +744,20 @@ def validate(self, mini_epoch, mode_cfg, batch_size):
                                 self.model,
                             )
 
-                    _ = self.loss_calculator_val.compute_loss(
-                        preds=preds,
-                        targets_and_aux=targets_and_auxs,
-                        metadata=extract_batch_metadata(batch),
-                    )
+                            preds, targets_and_auxs = self._process_validation_chunks(
```
shmh40, line 747: Should this be in the loop? I guess usually only 1 loss, so not too important

packages/readers_extra/src/weathergen/readers_extra/data_reader_anemoi_rt.py
```diff
@@ -0,0 +1,350 @@
...
+            for path in data_paths:
+                _anemoi_datasets.add_dataset_path(path)
+            # provide some visibility since we ignore filename
+            if is_root():
+                _logger.info("Ignoring filename and using _anemoi_config option.")
+
+        assert stream_info.get("frequency") is not None, "stream_info must contain 'frequency' key"
+        # self.frequency = np.timedelta64( timedelta_to_str(stream_info["frequency"]))
+        # TODO, TODO, TODO
+        self.frequency = np.timedelta64(
```
clessig: TODO: needs to be fixed

PR #2787, opened 2026-08-27
Sgrasse/develop/2786 remove time step

### LINKED ISSUE #2786: Remove `forecast.time_step`

Remove `forecast.time_step` and replace it with the value from `time_window_step`. It simplifies code of MultiStreamDataSampler and reduces the chance for misconfiguration (both values should be the same).

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation
- [ ] performance

grassesi:
## Description

Removes `forecast.time_step` from mode config and make TimeWindowHandler solely responsible for computations involving `time_window_step` and `time_window_len`. (if the latter is not desired only the last commit can be dropped)

## Issue Number

closes #2786 

#### FastEvaluation
 - [x] I have updated the public documentation if necessary

### DIFF
--- a/src/weathergen/datasets/data_reader_base.py
+++ b/src/weathergen/datasets/data_reader_base.py
@@ -47,6 +47,9 @@ class TimeIndexRange:
     start: TIndex
     end: TIndex
 
+    def __len__(self) -> int:
+        return int(self.end - self.start)
+
 
 @dataclass
 class DTRange:
@@ -94,8 +97,13 @@ class TimeWindowHandler:
         self.t_window_len: NPTDel64 = t_window_len_hours
         self.t_window_step: NPTDel64 = t_window_step_hours
 
+        idx_start: TIndex = np.int64(0)
+        idx_end = np.int64((self.t_end - self.t_start) // self.t_window_step)
+        self.index_range = TimeIndexRange(idx_start, idx_end)
+
         assert self.t_start < self.t_end, "end datetime has to be in the past of start datetime"
         assert self.t_start > _DT_ZERO, "start datetime has to be >= 1850-01-01T00:00."
+        assert idx_start <= idx_end, f"time window idxs invalid: {idx_start} <= {idx_end}"
 
     def __str__(self) -> str:
         # Helper to ensure readable timedelta formatting
@@ -104,26 +112,6 @@ class TimeWindowHandler:
         return (
             f"TimeWindowHandler: start={self.t_start}, end={self.t_end}, len={l_str}, step={s_str}"
         )
-
-    def get_index_range(self) -> TimeIndexRange:
-        """
-        Temporal window corresponding to index
-
-        Parameters
-        ----------
-        idx :
-            index of temporal window
-
-        Returns
-        -------
-            start and end of temporal window
-        """
-
-        idx_start: TIndex = np.int64(0)
-        idx_end = np.int64((self.t_end - self.t_start) // self.t_window_step)
-        assert idx_start <= idx_end, f"time window idxs invalid: {idx_start} <= {idx_end}"
-
-        return TimeIndexRange(idx_start, idx_end)
 
     def window(self, idx: TIndex) -> DTRange:
         """
@@ -143,6 +131,14 @@ class TimeWindowHandler:
         t_end_win = t_start_win + self.t_window_len
 
         return DTRange(t_start_win, t_end_win)
+
+    def max_index(self, n_steps) -> int:
+        """Last viable index to make a n_steps long forecast (including offset)."""
+        return self.index_range.end - (
+            # time that windows for n_steps cover
+            (self.t_window_step * n_steps + self.t_window_len)
+            // self.t_window_step  # as number of indexes
+        )
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


src/weathergen/datasets/data_reader_base.py
```diff
@@ -94,8 +97,13 @@ def __init__(
         self.t_window_len: NPTDel64 = t_window_len_hours
         self.t_window_step: NPTDel64 = t_window_step_hours
 
+        idx_start: TIndex = np.int64(0)
```
clessig, line 100: How is this related to the problem that the PR is about?

src/weathergen/datasets/data_reader_base.py
```diff
@@ -94,8 +97,13 @@ def __init__(
...
+        idx_end = np.int64((self.t_end - self.t_start) // self.t_window_step)
+        self.index_range = TimeIndexRange(idx_start, idx_end)
+
         assert self.t_start < self.t_end, "end datetime has to be in the past of start datetime"
         assert self.t_start > _DT_ZERO, "start datetime has to be >= 1850-01-01T00:00."
+        assert idx_start <= idx_end, f"time window idxs invalid: {idx_start} <= {idx_end}"
```
clessig, line 106: Shouldn't the check be on t_start and t_end. Then the correct ordering would be implied by l.101

src/weathergen/datasets/data_reader_base.py
```diff
@@ -144,6 +132,14 @@ def window(self, idx: TIndex) -> DTRange:
 
         return DTRange(t_start_win, t_end_win)
 
+    def max_index(self, n_steps) -> int:
```
clessig, line 135: I think it's confusing to have both `idx_end` and max_index(). Do we use the unadjusted idx_end somewhere?

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -457,7 +441,7 @@ def _build_stream_data_output(
         # collect for all forecast steps
         num_output_steps = self._get_output_length(num_forecast_steps)
         for step, timestep_idx in enumerate(range(self.output_offset, num_output_steps)):
-            step_forecast_dt = idx + (self.time_step * timestep_idx) // self.step_timedelta
+            step_forecast_dt = idx + timestep_idx
```
clessig, line 444: Why is this change correct

clessig (commented):
I left some higher level comments that we should decide on before going into the details.

src/weathergen/datasets/multi_stream_data_sampler.py
grassesi: `self.step_timedelta = mode_cfg.time_window_step = self.time_step`
=>
`(self.time_step * timestep_idx) // self.step_timedelta)` == `(self.step_timedelta * time_step_idx) // self.step_timedelta` == `(self.step_timedelta*time_step_idx)/self.step_timedelta == time_step_idx`

src/weathergen/datasets/data_reader_base.py
grassesi: No the unadjusted end index is not used anywhere, Instead two different versions of adjusted indexes are used:
1. in `check_samples` to check for the suitability of samples_per_mini_epoch (and eventual adjustment):
`(forecast_len(h) + window_len(h)) // step_len(h)`
2. in `_calc_baseperms` to size the permutation array of samples drawn:
`forecast_len(h) // step_len(h)`

In the usual case where step_len(h) = window_len(h) the difference in indexes is exactly 1 which is arguably correct: The last valid index into the permutations as sized by (2) is exactly (1) (`len(perms)-1`).

For step_len(h) =/= window_len(h) this does not hold true since the difference is exactly window_len(h) // step_len(h).
=> For window_len(h) < step_len(h) the last valid index as calculated by (1) is out of range when indexing (2)
=> For window_len(h) > step_len(h) the permutation array contains "extra" indices where forecasting from one of these would lead to targets outside the available data.

I think (1) should be the "correct" way to adjust the sampling indexes. `_calc_baseperms` should size the permuatation according to (1)-1 and not its own formular. Please tell me if this logic is sound to you.

src/weathergen/datasets/data_reader_base.py
grassesi: Yes this assertion could be removed (esp. since `idx_start` is hardcoded to be 0) :+1:

src/weathergen/datasets/data_reader_base.py
grassesi: This is just an artifact of the `get_index_range` method beeing wrapped into the constructor. I did this since get_index_range is the first thing that is called on the TimeWindowHandler, right after construction. It also the only time this method is ever called. Since I moved other index calculations into the TimeWindowHandler it made sense that `index_range` is a attribute of the TimeWindowHandler instead of the MultiStreamDataSampler. There is also no benefit over the lazy construction of the range (previous) instead of the eager construction (now).

grassesi:
@clessig There is a bit of nuance in how different indexes are handled (see my comment). I am quite sure that the adjusted length of the index range is the only value the MultiStreamDataHandler needs. If you agree with my logic (and there are no future requirements to account for), the PR can be simplified even further: Remove the TimeIndexRange dataclass since the only quantity of interest is the adjusted end idx. Calculate the adjusted end idx as a method of the TimeWindowHandler class (needs information about max number of forecast steps only known by the MultiStreamDataSampler).

Issue #2778, opened 2026-08-25
Performance issues in the inference pipeline


javak87:
The profiling results show that inference is slow and needs to be optimized.

Running nsys profiling on the branch `develop-ssl-diffusion-v1 `can give you the following trace:

<img width="1354" height="458" alt="Image" src="https://github.com/user-attachments/assets/c177817f-e679-47ff-bb5b-4c29f1c1166a" />

This trace corresponds to running the following command:

`../WeatherGenerator-private/hpc/launch-slurm.py --stage inference --from-run-id cw6a4szu   --mini-epoch 8 --nodes 1  --time 15:00   --options test_config.output.num_samples=0 test_config.forecast.num_steps=8     test_config.samples_per_mini_epoch=4 test_config.validation_noise_levels=[]     data_loading.num_workers=1 data_loading.num_workers_validation=1     test_config.compute_full_validation_loss=False`

PR #2791, opened 2026-08-31
Inference Pipeline Optimization - vectorize hpy_splits

javak87:
## Description
# Inference Pipeline Optimization

Tokenizer CPU time was dominated by `hpy_splits`: a Python loop over every HEALPix cell (`12 * 4**hl`) with a tiny `torch.argsort` / `cat` / `split` per cell. This replaces that with a vectorized NumPy path while preserving the same token index lists.

- **`hpy_cell_splits`**: Uses one `np.lexsort((theta, cell_id))` to group points by cell and sort them by latitude within each cell. Occupancy is computed using `np.bincount(..., minlength=12 * 4**hl)`. It returns packed `(idxs_ord, counts)` instead of a Python list of per-cell arrays.
- **`hpy_splits`**: Scatters the packed indices into one padded buffer, splits once into tokens, and then slices the tokens back per cell. The same padding (`index 0`) and `offset_step` behavior are preserved.
- **`tokenize_space` / `tokenize_spacetime`**: Unchanged; they continue to consume the same nested token lists.
- **Tests**: `tests/test_hpy_splits.py` freezes the original implementations and checks `torch.equal` on tokens for padded and unpadded cases, `offset_step`, tied coordinates, remainder tokens, and `hl=5`.

`hpy_cell_splits` is only used by `hpy_splits` in this tree.

## Inference Timing

```bash
```

| 1 | `n7sqya5y` | `1517694` | 32.39 | 3.2944E-02 | yes |
| 2 | `zl4t0n1w` | `1517696` | 32.38 | 3.2943E-02 | yes |
| 3 | `cio16jn2` | `1517699` | 32.58 | 3.2942E-02 | yes |
| **Mean (valid)** | — | — | **32.45** | cluster | **yes** |

Vs **develop-ssl-diffusion-v1** control mean **59.15 s/it**: **1.82×**.

**Note:** These timings only show up if the job tree is running `javad/optimize-hpy_splits` branch on private repo. `launch-slurm.py --from-run-id cw6a4szu` copies Python from that training run, not from your checkout, so launching from the optimized branch alone does not change the tokenizer unless those files are overlaid (or otherwise copied) into the new job directory.

## Issue Number
#2778 

#### FastEvaluation

### DIFF
--- a/src/weathergen/datasets/tokenizer_utils.py
+++ b/src/weathergen/datasets/tokenizer_utils.py
@@ -121,26 +121,23 @@ def hpy_cell_splits(coords: torch.tensor, hl: int):
     """Compute healpix cell id for each coordinate on given level hl
 
     Returns
-      hpy_idxs_ord_split : list of per cell indices into thetas,phis,posr3
-      thetas : thetas in rad
-      phis : phis in rad
+      idxs_ord : indices into thetas,phis,posr3, grouped by ascending healpix cell and
+        ordered by theta within each cell
+      counts : number of points in each of the 12 * 4**hl cells, so that
+        np.split(idxs_ord, np.cumsum(counts)) recovers the per cell indices
     """
     thetas, phis = theta_phi_to_standard_coords(coords)
     # healpix cells for all points
     hpy_idxs = ang2pix(2**hl, thetas, phis, nest=True)
 
-    # extract information to split according to cells by first sorting and then finding split idxs
-    hpy_idxs_ord = np.argsort(hpy_idxs, **numpy_argsort_args)
-    splits = np.flatnonzero(np.diff(hpy_idxs[hpy_idxs_ord]))
+    # One lexicographic sort yields the cell grouping and the by-theta order within every cell
+    # at once. Sorting stably by cell and then stably by theta per cell gives the same
+    # permutation, since lexsort is stable and applies the last key as the primary one.
+    thetas_np = thetas.numpy() if isinstance(thetas, torch.Tensor) else np.asarray(thetas)
+    idxs_ord = np.lexsort((thetas_np, hpy_idxs))
+    counts = np.bincount(hpy_idxs, minlength=12 * 4**hl)
 
-    # extract per cell data
-    hpy_idxs_ord_temp = np.split(hpy_idxs_ord, splits + 1)
-    hpy_idxs_ord_split = [np.array([], dtype=np.int64) for _ in range(12 * 4**hl)]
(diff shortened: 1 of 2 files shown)


src/weathergen/datasets/tokenizer_utils.py
```diff
@@ -156,37 +153,52 @@ def hpy_splits(
...
-            for idxs in hpy_idxs_ord_split
-        ]
-    else:
-        rem = np.zeros(len(hpy_idxs_ord_split), dtype=np.int32)
+    idxs_ord_flat, counts = hpy_cell_splits(coords, hl)
+
+    num_cells = len(counts)
+    if len(idxs_ord_flat) == 0:
+        empty = [[] for _ in range(num_cells)]
+        return empty, [[] for _ in range(num_cells)]
```
florianscheidl, line 164:
```suggestion
        return [[] for _ in range(num_cells)], [[] for _ in range(num_cells)]
```

tests/test_hpy_splits.py
florianscheidl, line 1: Good sanity check to see that the new function produces the same outputs as the original for these test cases. However, this is only an equivalence test; it doesn't guarantee that the function is correct.

For merging, I'd suggest we instead add a simple test (e.g., hl=1) where we understand the behavior and verify it.

florianscheidl (commented):
Looks good to me.
Small improvement in my opinion would be to move some of the comments on what happens in hpy_splits to the docstring.
RE the test, I'd suggest replacing it with a stand-alone test we trust (rather than verifying equal behaviour to the previous version, though that is a good check to have run :)).

javak87:
> Looks good to me. Small improvement in my opinion would be to move some of the comments on what happens in hpy_splits to the docstring. RE the test, I'd suggest replacing it with a stand-alone test we trust (rather than verifying equal behaviour to the previous version, though that is a good check to have run :)).

Done!

src/weathergen/datasets/tokenizer_utils.py
javak87: replaced!

tests/test_hpy_splits.py
florianscheidl: I see that the current test still includes the equivalence comparison, I think we can remove that.
Thanks for adding the simple test case.

florianscheidl (commented):
Small comment RE test. Otherwise looks good to me.

javak87:
> Small comment RE test. Otherwise looks good to me.

EQ tests have been removed.

florianscheidl (approved):
Thanks!

PR #2792, opened 2026-08-31
Inference Pipeline Optimization - vectorize get target coords

javak87:
## Description
Target-coordinate encoding spent most of its time **splitting points into one tiny tensor per HEALPix cell** (`12 * 4**hl`), concatenating that list again inside `locs_to_cell_coords_ctrs` / `locs_to_ctr_coords`, and repeating that for 5 vertex frames and 8 neighbor-center frames. This keeps the same output layout and replaces the split/cat round-trips with packed `bmm` / `matmul`.
- **`_rotate_points_per_cell`:** `repeat_interleave` cell ids from `masked_points_per_cell`, then one `torch.bmm` with that cell’s 3×3. Same math as `locs_to_cell_coords_ctrs` after `torch.cat` of the per-cell list, without building the list.
- **`verts_local`:** one `torch.repeat_interleave` on occupied cells instead of a Python `repeat` per cell and `torch.cat`.
- **Eight neighbor frames:** one `vecs_to_rots` on all neighbors×cells and one batched `matmul`, instead of eight `locs_to_ctr_coords` calls (each cat → bmm → split → cat).
- **`tokenize_apply_mask_target`:** unchanged; it still calls `get_target_coords_local` with packed coords and counts.

## Inference timing
```bash
```
| 1 | `hauhyu9g` | `1517706` | 36.77 | 3.2941E-02 | yes |
| 2 | `at9hwlw5` | `1517710` | 36.47 | 3.2944E-02 | yes |
| 3 | `qdn8mf8r` | `1517712` | 36.27 | 3.2944E-02 | yes |
| **Mean (valid)** | — | — | **36.50** | cluster | **yes** |

Vs **develop-ssl-diffusion-v1** control mean **59.15 s/it**: **1.62×**.

**Note:** These timings only show up if the job tree is running `javad/optimize-hpy_splits`. `launch-slurm.py --from-run-id cw6a4szu` copies Python from that training run, not from your checkout, so launching from the optimized branch alone does not change `get_target_coords_local` unless those files are overlaid (or otherwise copied) into the new job directory.

## Issue Number
#2778 

#### FastEvaluation

### DIFF
--- a/src/weathergen/datasets/tokenizer_utils.py
+++ b/src/weathergen/datasets/tokenizer_utils.py
@@ -6,10 +6,9 @@ from torch import Tensor
 
 from weathergen.common.io import IOReaderData
 from weathergen.datasets.utils import (
-    locs_to_cell_coords_ctrs,
-    locs_to_ctr_coords,
     r3tos2,
     s2tor3,
+    vecs_to_rots,
 )
 
 # on some clusters our numpy version is pinned to be 1.x.x where the np.argsort does not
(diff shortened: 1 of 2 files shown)


src/weathergen/datasets/tokenizer_utils.py
```diff
@@ -6,10 +6,9 @@
 
 from weathergen.common.io import IOReaderData
 from weathergen.datasets.utils import (
-    locs_to_cell_coords_ctrs,
```
florianscheidl, line 9: These two functions are not used anywhere else, except for tests. Before merging the PR, we should remove them (and the associated tests).

tests/test_get_target_coords_local.py
florianscheidl, line 1: Similar comment as for another PR: I think it's a good sanity check that the old and new implementations match. In my opinion, a standalone test for a simple case makes more sense.

florianscheidl (commented):
Looks good to me! Suggest a standalone test and removing the functions in the dataset utils.

src/weathergen/datasets/tokenizer_utils.py
javak87: removed!

florianscheidl (approved):
Lgtm

PR #2793, opened 2026-08-31
Inference Pipeline Optimization - vectorize normalize target channels

javak87:
## Description

`DataReaderBase._normalize` (used by `normalize_target_channels` and `normalize_source_channels`) used a Python loop over channels:

```python

for i, ch in enumerate(idx):

    data[..., i] = (data[..., i] - mean[ch]) / stdev[ch]

```

That is slow on large last-axis arrays (obs/anemoi windows). This replaces the loop with a vectorized pass while keeping bit-identical results vs the original loop.

## Inference timing

```bash
```

| 1 | `wupykk1o` | `1517714` | 34.98 | 3.2944E-02 | yes |
| 2 | `h4p86p2x` | `1517717` | 34.98 | 3.2942E-02 | yes |
| 3 | `v9z3r282` | `1517720` | 34.99 | 3.2945E-02 | yes |
| **Mean (valid)** | — | — | **34.98** | cluster | **yes** |

Vs **develop-ssl-diffusion-v1** control mean **59.15 s/it**: **1.69×**.

**Note:** These timings only show up if the job tree is running `javad/optimize-hpy_splits`. `launch-slurm.py --from-run-id cw6a4szu` copies Python from that training run, not from your checkout, so launching from the optimized branch alone does not change `_normalize` unless those files are overlaid (or otherwise copied) into the new job directory.

## Issue Number
#2778 

#### FastEvaluation

### DIFF
--- a/src/weathergen/datasets/data_reader_base.py
+++ b/src/weathergen/datasets/data_reader_base.py
@@ -556,6 +556,42 @@ class DataReaderBase(metaclass=ABCMeta):
             raise ValueError(
                 f"incorrect number of {name} channels: expected {len(idx)}, got {data.shape[-1]}"
             )
+        # Equivalent to (data[..., i] - mean[ch]) / stdev[ch] per channel.
+        # Dict stats are Python floats (weak scalars → same dtype as data). Array stats keep
+        # their dtype so float32 data minus float64 mean still promotes, matching the loop.
+        if isinstance(mean, dict):
+            mean_vec = np.array([mean[ch] for ch in idx], dtype=data.dtype)
+            std_vec = np.array([stdev[ch] for ch in idx], dtype=data.dtype)
+        else:
+            idx_arr = np.asarray(idx, dtype=np.intp)
+            mean_vec = np.asarray(mean)[idx_arr]
+            std_vec = np.asarray(stdev)[idx_arr]
+
+        if mean_vec.dtype == data.dtype and std_vec.dtype == data.dtype:
+            data -= mean_vec
+            data /= std_vec
+            return data
+
+        # Mixed dtypes (typically float32 data, float64 anemoi/obs stats): promote the same
+        # way the original channel loop did. Row-chunked ufuncs avoid a full-size float64
+        # temporary; few-channel or non-contiguous last axes stay on the scalar loop.
+        n_ch = data.shape[-1]
+        last_contiguous = data.strides[-1] == data.dtype.itemsize
+        if n_ch >= 8 and last_contiguous:
+            n_rows = int(np.prod(data.shape[:-1], dtype=np.intp))
+            flat = data.reshape(n_rows, n_ch)
+            if np.shares_memory(flat, data):
+                chunk = 8192
+                tmp_dtype = np.result_type(data.dtype, mean_vec.dtype, std_vec.dtype)
+                tmp = np.empty((min(chunk, n_rows), n_ch), dtype=tmp_dtype)
+                for start in range(0, n_rows, chunk):
+                    end = min(start + chunk, n_rows)
+                    sl = flat[start:end]
+                    view = tmp[: end - start]
+                    np.subtract(sl, mean_vec, out=view)
+                    np.divide(view, std_vec, out=view)
+                    np.copyto(sl, view)
+                return data
         for i, ch in enumerate(idx):
             data[..., i] = (data[..., i] - mean[ch]) / stdev[ch]
 
--- a/tests/test_normalize.py
+++ b/tests/test_normalize.py
@@ -1,0 +1,136 @@
+# (C) Copyright 2025 WeatherGenerator contributors.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""Equivalence of vectorized DataReaderBase._normalize vs the original channel loop."""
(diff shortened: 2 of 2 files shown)


src/weathergen/datasets/data_reader_base.py
```diff
@@ -556,6 +556,42 @@ def _normalize(
             raise ValueError(
                 f"incorrect number of {name} channels: expected {len(idx)}, got {data.shape[-1]}"
             )
+        # Equivalent to (data[..., i] - mean[ch]) / stdev[ch] per channel.
+        # Dict stats are Python floats (weak scalars → same dtype as data). Array stats keep
+        # their dtype so float32 data minus float64 mean still promotes, matching the loop.
+        if isinstance(mean, dict):
+            mean_vec = np.array([mean[ch] for ch in idx], dtype=data.dtype)
+            std_vec = np.array([stdev[ch] for ch in idx], dtype=data.dtype)
+        else:
```
florianscheidl, line 565: By type spec, mean should be a dict. Why do we need this case? We should either change the type spec or remove this case.

src/weathergen/datasets/data_reader_base.py
```diff
@@ -556,6 +556,42 @@ def _normalize(
             raise ValueError(
                 f"incorrect number of {name} channels: expected {len(idx)}, got {data.shape[-1]}"
             )
+        # Equivalent to (data[..., i] - mean[ch]) / stdev[ch] per channel.
+        # Dict stats are Python floats (weak scalars → same dtype as data). Array stats keep
+        # their dtype so float32 data minus float64 mean still promotes, matching the loop.
+        if isinstance(mean, dict):
```
florianscheidl, line 562: Why only checking mean and not stdev?

src/weathergen/datasets/data_reader_base.py
```diff
@@ -556,6 +556,42 @@ def _normalize(
...
+        # Mixed dtypes (typically float32 data, float64 anemoi/obs stats): promote the same
+        # way the original channel loop did. Row-chunked ufuncs avoid a full-size float64
+        # temporary; few-channel or non-contiguous last axes stay on the scalar loop.
+        n_ch = data.shape[-1]
+        last_contiguous = data.strides[-1] == data.dtype.itemsize
+        if n_ch >= 8 and last_contiguous:
+            n_rows = int(np.prod(data.shape[:-1], dtype=np.intp))
+            flat = data.reshape(n_rows, n_ch)
+            if np.shares_memory(flat, data):
+                chunk = 8192
```
florianscheidl, line 584: Where do these constants come from? Are they chosen based on our hardware?

florianscheidl (commented):
Overall looks good to me, just have some questions regarding chunking and type cases.

src/weathergen/datasets/data_reader_base.py
javak87: Thanks. The dict annotation is outdated — production always passes numpy self.mean / self.stdev (anemoi/obs).
 It was written for the old lookup style, not for how readers store stats.
The old loop was:
`data[..., i] = (data[..., i] - mean[ch]) / stdev[ch]`

src/weathergen/datasets/data_reader_base.py
javak87: We only inspect mean because every reader sets `self.mean` and `self.stdev` as the same kind of object (always numpy arrays).

src/weathergen/datasets/data_reader_base.py
javak87: No It's not related to hardware. 8192 is 2^13 rows, which at 8 bytes is 64 KiB of temp per channel. 4096 or 16384 would do the same job with a smaller or larger tmp. 8192 is a common power-of-two in the middle. Same order as other “tile ~8K” defaults in NumPy/BLAS code.

src/weathergen/datasets/data_reader_base.py
florianscheidl: Why do we still have the dict case if production always passes numpy arrays?

florianscheidl (commented):
Small comment on the dict vs numpy array cases. Still a bit unclear to me.

src/weathergen/datasets/data_reader_base.py
javak87: Yes. you are right. 
Removed the dict branch and typed mean / stdev as NDArray on `_normalize` and `_denormalize`. Production always passes self.mean / self.stdev as numpy arrays (anemoi/obs)

florianscheidl (approved):

src/weathergen/datasets/data_reader_base.py
```diff
@@ -556,6 +556,38 @@ def _normalize(
             raise ValueError(
                 f"incorrect number of {name} channels: expected {len(idx)}, got {data.shape[-1]}"
             )
+        # Equivalent to (data[..., i] - mean[ch]) / stdev[ch] per channel.
```
clessig, line 559: Why has an analogous optimization not been applied to `denormalize()`?

src/weathergen/datasets/data_reader_base.py
```diff
@@ -556,6 +556,38 @@ def _normalize(
...
+        idx_arr = np.asarray(idx, dtype=np.intp)
+        mean_vec = np.asarray(mean)[idx_arr]
+        std_vec = np.asarray(stdev)[idx_arr]
+
+        if mean_vec.dtype == data.dtype and std_vec.dtype == data.dtype:
+            data -= mean_vec
+            data /= std_vec
+            return data
+
+        # Mixed dtypes (typically float32 data, float64 anemoi/obs stats): promote the same
```
clessig, line 571: I don't think the float64 are needed. How much of the complexity is to perform the computations in float64?

PR #2795, opened 2026-09-01
HEALPix Curriculum

### LINKED ISSUE #2794: HEALPix curriculum

Add HEALPix curriculum to the WeatherGenerator for more efficient training.

In my bachelor thesis, I discovered that starting at a lower HEALPix Level (HL) and sequentially increasing it during training results in efficiency gains. This coarse-to-fine curriculum reduces compute time by almost 95% in the early stages of training up to HL 6. Furthermore, it continues to save compute to a smaller extent as the curriculum progresses, all without losing final model performance. When training to even finer resolutions like HL 8 it should even result in a greater efficiency gain. We should integrate a configuration mechanism to automate these HL transitions and dataset swaps natively during training.

_No response_

### Area

- [x] model
- [ ] science
- [ ] documentation
- [ ] performance

TillHae:
## Description

This adds a HEALPix curriculum to our training pipeline. During my bachelor thesis, I found that starting at a lower HEALPix Level (HL) and stepping it up during training cuts early compute time by almost 95%, all without hurting the final model performance. 

### How it works
- **Easy config**: Uses the `healpix_curriculum` parameter to set how many `isteps` each HL should run for.
- **Dynamic streams**: Uses OmegaConf interpolation (e.g., `streams_directory: "${curriculum_streams.${healpix_level}}"`) so the datasets swap automatically as the HL goes up.
- **Auto-restarts**: `run_train.py` now handles the transitions. When a stage ends, it stops the mini-epoch, saves a checkpoint and immediately starts the next stage within the same Slurm job.

### FYI
- **Configs**: I threw in `config_curriculum.yml` just to make reviewing the OmegaConf logic easier. Before merging it can be deleted so we don't clutter the repo with new configs.

## Issue Number

Closes #2794 

@kctezcan 

#### FastEvaluation

### DIFF
--- a/config/config_curriculum.yml
+++ b/config/config_curriculum.yml
@@ -1,8 +1,8 @@
 healpix_curriculum:
-  3: 100
-  4: 100
-  5: 100
-  6: 100
+  3: 250
+  4: 250
+  5: 250
+  6: 250
 
 curriculum_streams:
   3: "./config/streams/era5_1deg/"
--- a/config/config_curriculum.yml
+++ b/config/config_curriculum.yml
@@ -9,5 +9,3 @@ curriculum_streams:
   4: "./config/streams/era5_1deg/"
   5: "./config/streams/era5_1deg/"
   6: "./config/streams/era5_1deg/"
-
-streams_directory: "${curriculum_streams.${healpix_level}}"
--- a/config/config_curriculum.yml
+++ b/config/config_curriculum.yml
@@ -1,8 +1,8 @@
 healpix_curriculum:
-  3: 250
-  4: 250
-  5: 250
-  6: 250
+  3: 100
+  4: 100
+  5: 100
+  6: 100
 
 curriculum_streams:
   3: "./config/streams/era5_1deg/"
--- a/config/config_curriculum.yml
+++ b/config/config_curriculum.yml
@@ -1,13 +1,13 @@
 healpix_curriculum:
-  3: 1000
-  4: 1000
-  5: 1500
-  6: 3000
+  3: 250
+  4: 250
+  5: 250
+  6: 250
 
 curriculum_streams:
-  3: "./config/streams/era5_8deg/"
-  4: "./config/streams/era5_4deg/"
-  5: "./config/streams/era5_2deg/"
+  3: "./config/streams/era5_1deg/"
+  4: "./config/streams/era5_1deg/"
+  5: "./config/streams/era5_1deg/"
   6: "./config/streams/era5_1deg/"
 
 streams_directory: "${curriculum_streams.${healpix_level}}"
--- a/config/config_curriculum.yml
+++ b/config/config_curriculum.yml
@@ -1,0 +1,13 @@
+healpix_curriculum:
+  3: 1000
+  4: 1000
+  5: 1500
+  6: 3000
+
+curriculum_streams:
+  3: "./config/streams/era5_8deg/"
+  4: "./config/streams/era5_4deg/"
+  5: "./config/streams/era5_2deg/"
+  6: "./config/streams/era5_1deg/"
+
+streams_directory: "${curriculum_streams.${healpix_level}}"
--- a/packages/common/src/weathergen/common/config.py
+++ b/packages/common/src/weathergen/common/config.py
@@ -480,7 +480,12 @@ def load_merge_configs(
 def _load_streams_in_config(config: Config) -> Config:
     """If the config contains a streams_directory, loads the streams and returns the config with
     the streams set."""
-    streams_directory = config.get("streams_directory", None)
+    import omegaconf
+    try:
+        streams_directory = config.get("streams_directory", None)
+    except (omegaconf.errors.InterpolationKeyError, omegaconf.errors.InterpolationResolutionError):
+        streams_directory = None
+        
     config = config.copy()
     if streams_directory is not None:
         streams_directory = Path(streams_directory)
--- a/packages/common/src/weathergen/common/config.py
+++ b/packages/common/src/weathergen/common/config.py
@@ -24,6 +24,7 @@ import yaml.constructor
 import yaml.scanner
 from omegaconf import DictConfig, ListConfig, OmegaConf
 from omegaconf.omegaconf import open_dict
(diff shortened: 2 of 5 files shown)


TillHae:
n8cvp0hb_train_avg

Here are the results I got from my test run. It works pretty good! I had some problems with the configs and auto-restarting after a transition, but was able to make everything work with some help from AI.

I will do a self-review of my code on Friday.

TillHae:
> I will do a self-review of my code on Friday.

Took me some time to get back to it, but it looks good.

@kctezcan do you have time to review it?

src/weathergen/train/trainer_base.py
```diff
@@ -136,6 +136,11 @@ def init_ddp(cf):
                     dist.all_reduce(l_seed, op=torch.distributed.ReduceOp.SUM)
                     cf.data_loader_rng_seed = l_seed.item()
 
+        if dist.is_initialized():
```
clessig, line 139: Why is this change in the PR? It seems unrelated to the functionality that is targeted.

packages/common/src/weathergen/common/config.py
```diff
@@ -456,20 +457,42 @@ def load_merge_configs(
...
         # streams from an overwrite's streams_directory replace inherited streams
-        if any(o.get("streams_directory") is not None for o in overwrite_configs):
+        if any("streams_directory" in o for o in overwrite_configs):
             base_config.streams = None
     # use OmegaConf.unsafe_merge if too slow
     c = OmegaConf.merge(base_config, private_config, *overwrite_configs)
     assert isinstance(c, Config)
     c = _sanitize_time_keys(c)
 
+    if c.get("healpix_curriculum"):
```
clessig, line 467: Can we avoid this in config.py? And very specific functionality there turned out to be problematic.

config/config_curriculum.yml
```diff
@@ -0,0 +1,11 @@
+healpix_curriculum:
```
clessig, line 1: Could you explain what this config means / describes

src/weathergen/run_train.py
```diff
@@ -188,7 +217,35 @@ def run_train(args):
     trainer = Trainer(cf.train_logging)
 
     try:
-        trainer.run(cf, devices)
+        from_run_id_iter = None
```
clessig, line 220: This block does not belong to run_train.py. What is done here?

src/weathergen/train/trainer.py
```diff
@@ -358,6 +358,24 @@ def run(self, cf, devices, run_id_contd=None, mini_epoch_contd=None):
         if self.cf.general.istep > 0 and is_root():
             logger.info(f"Continuing run with learning rate: {self.lr_scheduler.get_lr()}")
 
+        if hasattr(self.cf, "healpix_curriculum") and self.cf.healpix_curriculum and is_root():
```
clessig, line 361: This should be performed in a separate function

src/weathergen/train/trainer.py
```diff
@@ -568,6 +594,22 @@ def train(self, mini_epoch):
 
             self.cf.general.istep += 1
 
+            if hasattr(self.cf, "healpix_curriculum") and self.cf.healpix_curriculum:
```
clessig, line 597: Let's try to avoid this complexity in the main training loop. Can you make a suggestion?

clessig (commented):
Thanks for the implementation. Some high-level comments for the moment.

src/weathergen/train/trainer_base.py
TillHae: The problem was that in my current implementation whenever I do a HL transition `dist` is already initialised, but `rank`, `world_size` and `local_rank` need to be reset.

config/config_curriculum.yml
TillHae: For the `healpix_curriculum` the key is always the HL and the value is the number of isteps trained at each level. At the end if the `number_of_mini_epochs` isn't met yet the training run is just continued on the last HL until the normal parameters are met.

Below you can choose which stream to use for each level.

Do you want me to add an explanation in the file?

src/weathergen/run_train.py
TillHae: This is a loop that continuously checks if the next HL stage has to be started because the `istep` limit given by the yml file was met and then it starts the next training run on the incremented HL. It also makes sure that the latest checkpoint of the model is saved before ending the last run.

src/weathergen/train/trainer.py
TillHae: I could calculate something like an exit istep in the beginning and then just do a simple comparison, which triggers the change of a config boolean parameter. Afterwards I check on the boolean outside the main training loop and then execute the transition when it comes to that case.

sophie-xhonneux:
Hey @TillHae, what is the issue that needs code changes? Can the curriculum not be implemented purely as a sequence of slurm jobs?

TillHae:
> Hey @TillHae, what is the issue that needs code changes? Can the curriculum not be implemented purely as a sequence of slurm jobs?

@sophie-xhonneux yes, it also works as a sequence of slurm jobs. You are probably right that it is not worth it to add complexity to the training process for small convenience reasons (not having to create a pipeline). If nobody intervenes I will close the PR and issue Wednesday morning.

I guess there are only two real advantages:
- Only one job is needed and therefore every transition doesn't reenter the queue
- Transitions are a bit faster due to less overhead (probably ~1-2 minutes per transition)

clessig:
> > Hey @TillHae, what is the issue that needs code changes? Can the curriculum not be implemented purely as a sequence of slurm jobs?
> 
> @sophie-xhonneux yes, it also works as a sequence of slurm jobs. You are probably right that it is not worth it to add complexity to the training process for small convenience reasons (not having to create a pipeline). If nobody intervenes I will close the PR and issue Wednesday morning.
> 
> I guess there are only two real advantages:
> 
> * Only one job is needed and therefore every transition doesn't reenter the queue
> * Transitions are a bit faster due to less overhead (probably ~1-2 minutes per transition)

It would be good to have a pipeline file then (in a config repo with longer lifetime) and a demonstration that it produces equivalent results to your implementation

PR #2811, opened 2026-09-10
[2810] Allow private config to override output directories and filename

### LINKED ISSUE #2810: Custom output file handler

The current WeatherGenerator version is quite rigid when it comes to the structure of the outputs: Checkpoints, inference outputs and logs are all written to parallel folders, and the name of the inference output is hardcoded. This makes sense when doing research, but in operation one often wants more control over the outputs. For instance, it is very useful to be able to specify the exact location of the checkpoint regardless of the output structure, and the inference output filename would also be nice to control.

The desired functionality is to extend the functionality of the private_config to be able to specify the model folder, log folder and results folder individually, as well as the output name. 

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation
- [ ] performance

evenmn:
## Description

Adds optional `private_config` overrides for model, log, and results directories, plus the output filename. Explicit directories are used directly, without appending the run ID. Existing defaults remain when overrides are omitted.

```yaml
path_shared_working_dir: /work/weathergen

path_model: /models/deployed
path_logs: /var/log/weathergen
path_results: /forecasts
output_name: forecast.zip
```
When multiple samples are generated, `rank` can be used to separate the different samples:
```yaml
output_name: "output_rank{rank:04d}.zip" 
```
Other variables that can be used in the filename are `epoch` and `step`. 

Checkpoint loading and logging use the selected private configuration.

## Issue Number

Closes #2810 

#### FastEvaluation

### DIFF
--- a/packages/common/src/weathergen/common/config.py
+++ b/packages/common/src/weathergen/common/config.py
@@ -719,14 +719,14 @@ def get_path_logs(config: Config) -> Path:
     return _get_path_output(config, "path_logs", "logs", get_run_id_from_config(config))
 
 
-def get_path_model(path_config: Config | None = None, run_id: str | None = None) -> Path:
+def get_path_model(model_config: Config | None = None, run_id: str | None = None) -> Path:
     """Get full_model_path if set, otherwise the shared per-run checkpoint directory."""
-    if path_config or run_id:
-        run_id = run_id if run_id else get_run_id_from_config(path_config)
+    if model_config or run_id:
+        run_id = run_id if run_id else get_run_id_from_config(model_config)
     else:
-        msg = f"Missing run_id and cannot infer it from config: {path_config}"
+        msg = f"Missing run_id and cannot infer it from config: {model_config}"
         raise ValueError(msg)
-    private_config = path_config if path_config is not None else _load_private_conf()
+    private_config = model_config if model_config is not None else _load_private_conf()
     return _get_path_output(private_config, "full_model_path", "models", run_id)
 
 
--- a/packages/common/src/weathergen/common/config.py
+++ b/packages/common/src/weathergen/common/config.py
@@ -731,18 +731,20 @@ def get_path_model(path_config: Config | None = None, run_id: str | None = None)
 
 
 def get_path_results(
-    config: Config,
+    run_config: Config,
     mini_epoch: int | None = None,
     step: int | None = None,
 ) -> Path:
     """Get the path for run results. Returns the results directory when mini_epoch is None."""
-    base_path = _get_path_output(config, "path_results", "results", get_run_id_from_config(config))
+    base_path = _get_path_output(
+        run_config, "path_results", "results", get_run_id_from_config(run_config)
+    )
     if mini_epoch is None:
         return base_path
 
-    ext = StoreType(config.zarr_store).value  # validate extension
-    default_name = f"validation_chkpt{mini_epoch:05d}_rank{config.rank:04d}.{ext}"
-    fname_template = config.get("output_name")
+    ext = StoreType(run_config.zarr_store).value  # validate extension
+    default_name = f"validation_chkpt{mini_epoch:05d}_rank{run_config.rank:04d}.{ext}"
+    fname_template = run_config.get("output_name")
     if fname_template is None:
         fname = default_name
     else:
@@ -750,7 +752,7 @@ def get_path_results(
             fname = fname_template.format(
                 epoch=mini_epoch,
                 step=step,
-                rank=config.rank,
+                rank=run_config.rank,
             )
         except (IndexError, KeyError, ValueError):
             fname = default_name
--- a/packages/common/src/weathergen/common/config.py
+++ b/packages/common/src/weathergen/common/config.py
@@ -219,7 +219,8 @@ def load_run_config(
         run_id: Run ID of the pretrained WeatherGenerator model
         mini_epoch: Mini_epoch of the checkpoint to load. -1 indicates last checkpoint available.
         model_path: Parent model directory containing run-id subdirectories.
-        config: Active configuration for resolving path_model when model_path is None.
+        config: Active configuration whose full_model_path specifies the exact checkpoint
+            directory (without appending run_id), used when model_path is None.
 
     Returns:
         Configuration object loaded from the specified run and mini_epoch.
@@ -721,14 +722,14 @@ def get_path_run(config: Config) -> Path:
 
 
 def get_path_model(config: Config | None = None, run_id: str | None = None) -> Path:
-    """Get the current runs model_path for storing model checkpoints."""
+    """Get full_model_path if set, otherwise the shared per-run checkpoint directory."""
     if config or run_id:
         run_id = run_id if run_id else get_run_id_from_config(config)
     else:
         msg = f"Missing run_id and cannot infer it from config: {config}"
         raise ValueError(msg)
     config = config if config is not None else _load_private_conf()
-    return _get_output_directory(config, "path_model", "models", run_id)
+    return _get_output_directory(config, "full_model_path", "models", run_id)
 
 
 def get_path_results(config: Config, mini_epoch: int) -> Path:
--- a/packages/common/src/weathergen/common/config.py
+++ b/packages/common/src/weathergen/common/config.py
@@ -208,7 +208,9 @@ def save(config: Config, mini_epoch: int | None):
         f.write(json_str)
 
 
-def load_run_config(run_id: str, mini_epoch: int | None, model_path: str | None) -> Config:
+def load_run_config(
+    run_id: str, mini_epoch: int | None, model_path: str | None, *, config: Config | None = None
+) -> Config:
     """
     Load a configuration file from a given run_id and mini_epoch.
     If run_id is a full path, loads it from the full path.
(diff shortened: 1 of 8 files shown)


clessig:
@grassesi : could you also have a look?

clessig:
@evenmn : could you lint please :)

packages/common/src/weathergen/common/config.py
```diff
@@ -208,15 +208,18 @@ def save(config: Config, mini_epoch: int | None):
...
     """
     Load a configuration file from a given run_id and mini_epoch.
     If run_id is a full path, loads it from the full path.
 
     Args:
         run_id: Run ID of the pretrained WeatherGenerator model
         mini_epoch: Mini_epoch of the checkpoint to load. -1 indicates last checkpoint available.
-        model_path: Path to the model directory. If None, uses the model_path from private config.
+        model_path: Parent model directory containing run-id subdirectories.
+        config: Active configuration for resolving path_model when model_path is None.
```
clessig: Can we avoid the very similar yet distinct `path_model` and `model_path`. Does this refer to the same thing or not? If the same, please use the same name, otherwise clearly distinguish.

packages/common/src/weathergen/common/config.py
```diff
@@ -255,9 +258,9 @@ def load_run_config(run_id: str, mini_epoch: int | None, model_path: str | None)
     with fname.open() as f:
         json_str = f.read()
 
-    config = OmegaConf.create(json.loads(json_str))
+    loaded_config = OmegaConf.create(json.loads(json_str))
```
clessig: Why do you rename here?

packages/common/src/weathergen/common/config.py
clessig: config is the private config here? Can we make this explizit in the name.

packages/common/src/weathergen/common/config.py
```diff
@@ -698,9 +701,23 @@ def load_streams(streams_directory: Path) -> Config:
     return OmegaConf.create(streams)
 
 
+def _get_output_directory(config: Config, key: str, folder: str, run_id: str) -> Path:
```
clessig: The more consistent name for the function is `_get_path_output`. Please change.

packages/common/src/weathergen/common/config.py
```diff
@@ -710,14 +727,16 @@ def get_path_model(config: Config | None = None, run_id: str | None = None) -> P
     else:
         msg = f"Missing run_id and cannot infer it from config: {config}"
         raise ValueError(msg)
-    return _get_shared_wg_path() / "models" / run_id
+    config = config if config is not None else _load_private_conf()
```
clessig: Please use `private_config` as var name

packages/common/src/weathergen/common/logger.py
```diff
@@ -123,7 +123,7 @@ def init_loggers(run_id=None, logging_config=None):
     # output_dir = f"./output/{timestamp}-{run_id}"
     output_dir = ""
     if run_id is not None:
-        output_dir = f"./logs/{run_id}"
+        output_dir = str(log_path) if log_path is not None else f"./logs/{run_id}"
```
clessig: Avoid the branch here: this should be handled in `config.get_path_logs(cf)`

src/weathergen/run_train.py
```diff
@@ -100,7 +100,7 @@ def run_inference(args):
     devices = Trainer.init_torch()
     cf = Trainer.init_ddp(cf)
 
-    init_loggers(cf.general.run_id)
+    init_loggers(cf.general.run_id, log_path=config.get_path_logs(cf))
```
clessig: See above.

src/weathergen/run_train.py
```diff
@@ -139,7 +139,7 @@ def run_continue(args):
     devices = Trainer.init_torch(multiprocessing_method=mp_method)
     cf = Trainer.init_ddp(cf)
 
-    init_loggers(cf.general.run_id)
+    init_loggers(cf.general.run_id, log_path=config.get_path_logs(cf))
```
clessig: See above.

src/weathergen/run_train.py
```diff
@@ -176,7 +176,7 @@ def run_train(args):
 
     # this line should probably come after the processes have been sorted out else we get lots
     # of duplication due to multiple process in the multiGPU case
-    init_loggers(cf.general.run_id)
+    init_loggers(cf.general.run_id, log_path=config.get_path_logs(cf))
```
clessig: See above.

clessig (commented):
Thanks, this is a useful feature. Some minor comments. We also should merge this into the preops branch.

packages/common/src/weathergen/common/config.py
evenmn: This was for some reason needed to resolve the listing issue. The linter complained even after running `./scripts/actions.sh lint`

packages/common/src/weathergen/common/config.py
evenmn: They are different: `model_path` is the parent directory without `<run-id>`, `path_model` is the full path after `<run-id>` is appended. The naming was indeed confusing, I renamed `path_model` to `full_model_path`.

evenmn: Yes! It's renamed now

packages/common/src/weathergen/common/config.py
evenmn: Done

packages/common/src/weathergen/common/config.py
evenmn: Changed

packages/common/src/weathergen/common/config.py
evenmn: Ah, I realized now that this was because this other parameter was named `config`. Since I now renamed it to `private_config`, I will revert this to just `config`

evenmn:
@clessig Thank you for the comments. Please have a look at my answers and latest commits

packages/common/src/weathergen/common/config.py
```diff
@@ -208,15 +208,23 @@ def save(config: Config, mini_epoch: int | None):
         f.write(json_str)
 
 
-def load_run_config(run_id: str, mini_epoch: int | None, model_path: str | None) -> Config:
+def load_run_config(
+    run_id: str,
+    mini_epoch: int | None,
+    model_path: str | None,
+    *,
```
clessig: What is `*` for?

packages/common/src/weathergen/common/config.py
```diff
@@ -698,26 +706,42 @@ def load_streams(streams_directory: Path) -> Config:
...
+
+def get_path_logs(config: Config) -> Path:
+    """Get the application log directory."""
+    return _get_path_output(config, "path_logs", "logs", get_run_id_from_config(config))
+
+
 def get_path_run(config: Config) -> Path:
     """Get the current runs results_path for storing run results and logs."""
-    return _get_shared_wg_path() / "results" / get_run_id_from_config(config)
+    return _get_path_output(config, "path_results", "results", get_run_id_from_config(config))
```
clessig: This function doesn't make sense--we don't *the* path for a run. Can you remove it. The function is used in trainer.py and in train_logger.py. It's calls should be replace by calls to `get_path_logs()`--but please check that you agree.

packages/common/src/weathergen/common/config.py
```diff
@@ -698,26 +706,42 @@ def load_streams(streams_directory: Path) -> Config:
...
 
 
 def get_path_model(config: Config | None = None, run_id: str | None = None) -> Path:
-    """Get the current runs model_path for storing model checkpoints."""
+    """Get full_model_path if set, otherwise the shared per-run checkpoint directory."""
```
clessig, line 723: config is here private config, right? Can we change the variable name.

packages/common/src/weathergen/common/config.py
```diff
@@ -698,26 +706,42 @@ def load_streams(streams_directory: Path) -> Config:
...
+    private_config = config if config is not None else _load_private_conf()
+    return _get_path_output(private_config, "full_model_path", "models", run_id)
 
 
 def get_path_results(config: Config, mini_epoch: int) -> Path:
     """Get the path to validation results for a specific mini_epoch and rank."""
     ext = StoreType(config.zarr_store).value  # validate extension
     base_path = get_path_run(config)
     fname = f"validation_chkpt{mini_epoch:05d}_rank{config.rank:04d}.{ext}"
+    fname = config.get("output_name") or fname
```
clessig: config is here private config, right? Can we change the variable name.

clessig (commented):
Thanks for the changes! Some minor comments left. For the functions we anyway touch, we should be clear if `config` is "model config" or "private config". It makes the code much more readable.

packages/common/src/weathergen/common/config.py
evenmn: It's not needed. I removed it

packages/common/src/weathergen/common/config.py
evenmn: config here can be the full run config, to get run-id. private config is the fallback

evenmn: I changed name to model_config

packages/common/src/weathergen/common/config.py
evenmn: This is also the full run config. Changed name to model_config

packages/common/src/weathergen/common/config.py
evenmn: I agree, `get_path_run` was replaced by `get_path_results`

evenmn (commented):
I have verified that the logs and models are written to the paths specified in `private_config` during training, and that the results are written to the expected file during inference

clessig (approved):
Thanks for the changes. Looks great now--approving.

Can you open the PR also against the preops branch.

PR #2815, opened 2026-09-14
Add support for non-contiguous training periods with date_ranges

### LINKED ISSUE #2814: Allow for non contiguous time ranges in training

Currently you can only set
```yaml
training_config:
  start_date: 1979-01-01T00:00
  end_date: 2022-12-31T00:00
```
to define the time window for training data. To have specific holdout periods within the time window date ranges are required. E.g. being able to set
```yaml
training_config:
  date_ranges:
    - start_date: 1979-01-01T00:00
      end_date:   1992-12-31T00:00
    - start_date: 2000-01-01T00:00
      end_date:   2010-12-31T00:00
```

_No response_

_No response_

### Area

- [ ] model
- [ ] science
- [ ] documentation
- [ ] performance

jpolz:
## Description

Allows to set
```
training_config:
  date_ranges:
    - start_date: 1979-01-01T00:00
      end_date:   1992-12-31T00:00
    - start_date: 2000-01-01T00:00
      end_date:   2010-12-31T00:00
```
but is backward compatible with the following format:
```
training_config:
  start_date: 1979-01-01T00:00
  end_date: 2022-12-31T00:00
```

## Issue Number

Closes https://github.com/ecmwf/WeatherGenerator/issues/2814

#### FastEvaluation


src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -123,9 +123,23 @@ def __init__(self, cf: Config, mode_cfg: dict, stage: Stage):
...
+        raw_ranges = mode_cfg.get("date_ranges")
+        if raw_ranges is not None:
+            self._valid_ranges: list[tuple] | None = [
+                (r.start_date, r.end_date) for r in raw_ranges
+            ]
+            t_start = min(r[0] for r in self._valid_ranges)
+            t_end = max(r[1] for r in self._valid_ranges)
+        else:
+            self._valid_ranges = None
+            t_start = self.mode_cfg.start_date
```
clessig, line 137: We should have and assert that makes sure t_start and t_end are not None (and time rnage was not specified)

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -213,9 +219,22 @@ def _calc_baseperms(self, fsm: int) -> np.typing.NDArray:
...
+
+        # include only indices whose mapped time falls within one of the valid periods
+        all_indices = np.arange(self.max_input_steps, perms_len)
+        t_start = self.time_window_handler.t_start
+        step = self.time_window_handler.t_window_step
+        times = t_start + all_indices * step
+        forecast_horizon = (fsm + self.output_offset) * self.time_step
+        mask = np.zeros(len(all_indices), dtype=bool)
+        for p_start, p_end in self._valid_ranges:
+            mask |= (times >= p_start) & (times + forecast_horizon <= p_end)
```
clessig: Why do we have here `<= p_end` and not `< p_end`?

clessig (commented):
This looks great, just a minor question.

src/weathergen/train/utils.py
```diff
@@ -170,6 +170,13 @@ def get_active_stage_config(
     final config that is used
     """
 
+    base_config = copy.deepcopy(base_config)
```
clessig, line 173: This surface a structural issue we are introducing with this PR: we have 2 ways to specify the same thing. Structurally, `date_ranges` would be sufficient. 

We will need `start_date` and `end_date` for backward compatibility for a while but do we want to keep it in the long run?

clessig:
@jpolz : could you please lint

src/weathergen/datasets/multi_stream_data_sampler.py
jpolz: changed to `< p_end` now which is consistent with the previous implementation

src/weathergen/train/utils.py
```diff
@@ -170,6 +170,17 @@ def get_active_stage_config(
...
+    # date_ranges must not leak into a stage that explicitly overrides start_date/end_date,
+    # whether the stale date_ranges comes from the base config (e.g. inherited from
+    # training_config) or from the merge config itself (e.g. a previously saved test_config).
+    if "start_date" in merge_config or "end_date" in merge_config:
+        if "date_ranges" in base_config:
+            del base_config["date_ranges"]
+        if "date_ranges" in merge_config:
```
clessig: How do we know here that date_ranges was not specified after start_date/end_date and is meant to take priority?

jpolz: Thanks for the comment. I am trying to disentangle that right now. As you mentioned the issue comes from the two keys that do the same thing. To avoid fixes as above it would be easier to revert to one key, but a list for start_date and end_date that specifies ranges is also confusing and harder to read than the ranges solution. Maybe we should have a proper discussion instead?

clessig: I don't think we should over-engineer. It's a decision if we want to retain start_data/end_date or not. I think it's a useful shortcut and the complications in the code through this are manageable. And if we decide to remove, we still would need to deal with it for backward compatibility.

I think I would add one extra test: `start_date==null` and `end_date==null`. So if one wants to switch from start_data/end_date to date_ranges one has to set start_data=0 and end_date=0. It's probably worth to think through this once more that we catch as many corner cases as possible. But it's a peculiar use case and I wouldn't try to find an optimal solution for all possible cases that can happen.

jpolz: Here is a new version that can be explained by the following diagram:

image

packages/common/src/weathergen/common/config.py
```diff
@@ -465,8 +504,13 @@ def load_merge_configs(
         # streams from an overwrite's streams_directory replace inherited streams
         if any(o.get("streams_directory") is not None for o in overwrite_configs):
             base_config.streams = None
-    # use OmegaConf.unsafe_merge if too slow
-    c = OmegaConf.merge(base_config, private_config, *overwrite_configs)
+    # merge one config at a time so date_ranges vs. start_date/end_date precedence (see
+    # reconcile_date_ranges) is resolved at each step, e.g. when continuing/fine-tuning
+    # training from a run whose training_config used date_ranges
+    c = base_config
+    for nxt in (private_config, *overwrite_configs):
```
clessig, line 511: Remove private_config here. It plays no role in `_reconcile_stage_date_ranges` and should hence not be passed

packages/common/src/weathergen/common/config.py
```diff
@@ -465,8 +504,13 @@ def load_merge_configs(
...
+        c, nxt = _reconcile_stage_date_ranges(c, nxt)
+        c = OmegaConf.merge(c, nxt)
```
clessig, line 513: Pass private_config here

packages/common/src/weathergen/common/config.py
```diff
@@ -411,6 +419,37 @@ def merge_configs(base_config: Config, update_config: Config):
...
+            override_stage = copy.deepcopy(override_stage)
+            del override_stage["date_ranges"]
+    return base_stage, override_stage
+
+
+def _reconcile_stage_date_ranges(acc: Config, nxt: Config) -> tuple[Config, Config]:
+    """Apply reconcile_date_ranges to every stage sub-config (training/validation/test)"""
+    acc = acc.copy()
+    nxt = nxt.copy()
+    for stage_key in ("training_config", "validation_config", "test_config"):
```
clessig, line 442: acc should always one of `"training_config", "validation_config", "test_config"` so I do not understand the logic here.

jpolz: Tried to clarify what is done. acc (now base_run_config) can have any of "training_config", "validation_config", "test_config" set. when I use an override config via --config in launch-slurm.py, e.g. containing

```
# CV fold 2 — test period: 1986–1992 (SSWs: Jan 1987, Dec 1987, Feb 1989)
# Training is non-contiguous: 1979–1985 + 1993–2010.
training_config:
  date_ranges:
    - start_date: 1979-01-01T00:00
      end_date:   1985-12-31T00:00
    - start_date: 1993-01-01T00:00
      end_date:   2010-12-31T00:00

validation_config:
  start_date: 2021-01-01T00:00
  end_date:   2022-12-31T00:00
```
Then this function handles it, hopefully in a correct way.

packages/common/src/weathergen/common/config.py
```diff
@@ -411,6 +419,43 @@ def merge_configs(base_config: Config, update_config: Config):
     return OmegaConf.merge(base_config, update_config)
 
 
+def reconcile_date_ranges(base_stage: Config, override_stage: Config) -> tuple[Config, Config]:
+    """
+    Resolve date_ranges vs. start_date/end_date precedence between two configs of the
+    same stage (e.g. two training_config blocks, or a training_config and a
+    validation_config) before they get merged with OmegaConf.merge.
+    """
+    if override_stage.get("start_date") is not None or override_stage.get("end_date") is not None:
```
shmh40: Does the overriding work properly both ways here? What happens if we want to override with date_ranges? We still keep the start_date and end_date right?

jpolz: Thanks for pointing that out. It actually doesn't work properly both ways. I am now thinking of a more generic way to treat such cases using a registry of mutually exclusive key pairs. I don't want to encourage introducing more such cases, but it makes sense to generalize nonetheless..

shmh40: Cool, thanks, otherwise lgtm!

clessig: @jpolz : can you please lint :)

packages/common/src/weathergen/common/config.py
```diff
@@ -411,6 +419,79 @@ def merge_configs(base_config: Config, update_config: Config):
...
+    _ALTERNATE_STAGE_REPRESENTATIONS, e.g. date_ranges vs. start_date/end_date) between two
+    configs of the same stage (e.g. two training_config blocks, or a training_config and a
+    validation_config) before they get merged with OmegaConf.merge.
+
+    Since alternate representations coexist as independent keys, a plain OmegaConf.merge
+    would leave a stale key from the base behind whenever the override switches
+    representation. So here, whichever representation the override specifies wins, and the
+    other representations are dropped from both stages before merging.
+    """
+    for representations in _ALTERNATE_STAGE_REPRESENTATIONS:
```
shmh40, line 455: If I have start and end in the default config, and then I add date_range in an override config, will this discard date_ranges from the config? Sorry if not, bit hard to work out exactly what is happening here

src/weathergen/datasets/multi_stream_data_sampler.py
```diff
@@ -213,9 +222,24 @@ def _calc_baseperms(self, fsm: int) -> np.typing.NDArray:
...
+        # include only indices whose full sample footprint (input steps through forecast
+        # target steps) falls within one of the valid periods
+        all_indices = np.arange(self.max_input_steps, perms_len)
+        t_start = self.time_window_handler.t_start
+        step = self.time_window_handler.t_window_step
+        times = t_start + all_indices * step
+        input_horizon = self.max_input_steps * step
+        forecast_horizon = (fsm + self.output_offset) * self.time_step
+        mask = np.zeros(len(all_indices), dtype=bool)
+        for p_start, p_end in self._valid_ranges:
```
shmh40, line 240: Could this ever go beyond the specified range? If time_window_len > time_window_step?

packages/common/src/weathergen/common/config.py
jpolz: the override config wins, so start_date and end_date will be discarded

src/weathergen/datasets/multi_stream_data_sampler.py
jpolz: It is designed not to. But I have not tested it yet and I am not aware of all possible edge cases

PR #2817, opened 2026-09-15
Register MLP norm only in self.layers, not also as self.lnorm

moritzhauschulz:
MLP.__init__ bound one norm module to two attribute paths for non-DiT MLPs:

```
    self.lnorm = norm(dim_in, eps=norm_eps)
    if not is_dit:
        self.layers.append(self.lnorm)
```

state_dict() therefore emitted the same tensor twice, as "lnorm.*" and as "layers.0.*". Consequences:

  - FSDP2 state_dict() reports duplicate parameter paths for the shared module, while named_modules() deduplicates and keeps only the canonical "layers.0" path, so the alias has no module to initialise.
  - Every checkpoint written this way carries two identical copies of each norm tensor. Loading such a checkpoint into de-aliased code reports the "lnorm.*" copies as unmatched -- harmless in itself, but it buries real mismatches in noise.

Register the norm only in self.layers. forward() already applies it through the self.layers loop and special-cases AdaLayerNorm to pass ada_ln_aux, so behaviour is unchanged. The DiT branch keeps self.lnorm: it is an AdaLNZero applied explicitly in forward() and never appended to self.layers, so it was never aliased.

Verified: a non-DiT MLP now exposes only layers.* in state_dict(), with and without dim_aux; forward passes unchanged; ruff clean.

## Issue Number

#### FastEvaluation


clessig:
@moritzhauschulz : this seems to be a general issue and should also be fixed on `develop-ssl-diffusion`?

moritzhauschulz:
> @moritzhauschulz : this seems to be a general issue and should also be fixed on `develop-ssl-diffusion`?

Yes. I was under the impression this branch would supersede `develop-ssl-diffusion`, but happy to discuss. I already made several PRs into it.

clessig:
> > @moritzhauschulz : this seems to be a general issue and should also be fixed on `develop-ssl-diffusion`?
> 
> Yes. I was under the impression this branch would supersede `develop-ssl-diffusion`, but happy to discuss. I already made several PRs into it.

@MatKbauer @moritzhauschulz : could we make sure we get the fix also on develop-ssl-diffusion-v1.

PR #2821, opened 2026-09-16
Add load_encoder_chkpt, and do not re-apply either overlay on continuation

moritzhauschulz:
Two changes to the checkpoint-overlay machinery in model_interface.py.

1. load_encoder_chkpt -- the mirror of load_decoder_chkpt. Filters a checkpoint to `encoder.*` (stream embedders, ae_* engines, learned queries) and overlays it after the primary load, leaving the forecast engine and decoder from the primary checkpoint untouched. This allows pairing one run's forecast engine with a different run's encoder. The two selections are disjoint, so both overlays may be configured at once.

   The shared body of both loaders is factored into _load_part_from_checkpoint rather than duplicated. This is deliberate: the decoder overlay was originally written by copying only the *unsharded* half of load_model's `module.` prefix handling, so every sharded lookup missed and the run silently trained a randomly initialised decoder. One body means the two parts cannot drift apart that way again.

2. Neither overlay is re-applied when continuing a run. The primary load is an if/elif: a continuation loads from `run_id_contd` -- the run's own checkpoint, which already contains whatever training that part has had -- and never consults `load_chkpt`. Re-overlaying afterwards reverted a decoder-finetune's decoder to the borrowed original on every chained-job restart, discarding the finetuning done since the previous restart. 

Known limitation: `run_id_contd` is not a train_continue signal -- inference passes `--from-run-id` through it -- so both overlays are now also skipped during inference. If the overlay path is needed at inference, gate on an explicit `overlay_decoder` / `overlay_encoder` parameter instead of on `run_id_contd`; note target_and_aux_utils.py is a third call site that also passes a non-None `run_id_contd`.

## Issue Number

-   [ ] **in progress** I have tried my changes with data and code:

#### FastEvaluation


MatKbauer (approved):
Verified successfully. Thanks Moritz 👍

